PR #12, opened 2018-10-10
Further losslessly optimize images

alrra:


mathiasbynens:
How did you do this? I had previously used ImageOptim.app on these images.

alrra:
> How did you do this? I had previously used ImageOptim.app on these images.

`ImageOptim` (maybe I have my setting set to use more time but optimize better?)

mathiasbynens:
Ah, yes. Mine was set to “extra” previously. Fixed it:



Thanks for the tip!

PR #30, opened 2018-10-17
Avoid weird paragraph spacing

tomayac:
Dependent on the width, the hidden bookmark would still consume screen real estate if it would break to a new line as the only word. This fixes the issue seen below:

* Still fitting at 377px width:
  screenshot 2018-10-17 at 10 09 10
* Breaking at 375px width:
  screenshot 2018-10-17 at 10 09 27

The downside is that this will cause a reflow upon hover, but this seems better in comparison to the spacing issue.

### DIFF
--- a/src/_css/main.css
+++ b/src/_css/main.css
@@ -212,12 +212,12 @@ main a:hover code, main a:focus code {
   background-color: inherit;
 }
 a.bookmark {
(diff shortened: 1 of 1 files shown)


mathiasbynens:
Thank you!

PR #45, opened 2018-10-24
Format Torque code examples

mathiasbynens:
This patch applies the Torque formatter [1] to the code examples in the documentation.

[1] https://v8.dev/docs/torque#torque-tooling

### DIFF
--- a/src/docs/torque.md
+++ b/src/docs/torque.md
@@ -202,7 +202,7 @@ type int31 extends int32 generates 'TNode<Int32T>' constexpr 'int31_t';
 Union types express that a value belongs to one of several possible types. We only allow union types for tagged values, because they can be distinguished at runtime using the map pointer. For example, JavaScript numbers are either Smi values or allocated HeapNumber objects.
 
 ```torque
-type Number = Smi|HeapNumber;
+type Number = Smi | HeapNumber;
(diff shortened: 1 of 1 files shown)


src/docs/torque.md
```diff
@@ -202,7 +202,7 @@ type int31 extends int32 generates 'TNode<Int32T>' constexpr 'int31_t';
 Union types express that a value belongs to one of several possible types. We only allow union types for tagged values, because they can be distinguished at runtime using the map pointer. For example, JavaScript numbers are either Smi values or allocated HeapNumber objects.
 
 ```torque
-type Number = Smi | HeapNumber;
+type Number = Smi|HeapNumber;
```
mathiasbynens: Is this intentional, or should we fix the formatter to add the spaces? They kind of looked nice. cc @tebbi

dannoc: No, this isn't intentional, but the JS formatter that we use somehow doesn't understand this syntax and we'd have to fix it/work around it explicitly.

mathiasbynens: Fixed by https://chromium-review.googlesource.com/c/v8/v8/+/1297949. Thanks, @dannoc!

PR #49, opened 2018-10-29
[docs] Add documentation for zero-cost async stack traces.

bmeurer:
Bug: v8:7522
Ref: nodejs/node#11865
Design-Document: http://bit.ly/v8-zero-cost-async-stack-traces


mathiasbynens:
Thanks!

PR #51, opened 2018-11-04
Further optimize `v8-turbofan.svg`

alrra:
@mathiasbynens This is not a lossless change. There is a minor difference, but it's not human noticeable.

---

Here is the diff:


alrra:
@mathiasbynens Since https://github.com/v8/v8.dev/commit/07a7820912b434c4e6b87344722bb53b07f6badc was not a lossless, but not human noticeable difference, I thought you would accept this PR too.

mathiasbynens:
Thanks!

PR #57, opened 2018-11-09
Document ICU version required for V8

gsathya:
See https://bugs.chromium.org/p/v8/issues/detail?id=8424 for background.


gsathya:
I fear that this is going to get stale quickly :disappointed:

mathiasbynens:
Thanks! Looks like these docs are in a pretty bad shape in general.

PR #58, opened 2018-11-09
Improve i18n docs

mathiasbynens:


mathiasbynens:
Thanks for the quick review. I’ve filed https://github.com/v8/v8.dev/issues/59 to track the remaining sections I’m suspicious about.

PR #68, opened 2018-11-13
support Windows builds (at least from my msys2 environment)

tebbi:


mathiasbynens:
Is using “git bash” or [the Linux subsystem](https://docs.microsoft.com/en-us/windows/wsl/install-win10) not an option for you?

tebbi:
I am using msys2, which should be more or less equivalent to git bash. The Linux subsystem is not available on corp machines. The mkdir from npm behaves really weirdly: it created a directory called "-p"...
Wrapping it in a call to sh solved this problem. My guess is that npm on Windows uses a very incomplete polyfill for mkdir.

tebbi:
I updated the pull request as discussed offline.

PR #89, opened 2018-12-05
fix: こにちは → こんにちは

myakura:
as we Japanese don't use "こにちは" https://en.wikipedia.org/wiki/Konnichi_wa


kosamari:
yes こんにちは is the right spelling :)

mathiasbynens:
Thank you @myakura and @kosamari!

PR #95, opened 2018-12-08
fix: cquery wiki url

tsangint:
[Previous url](https://github.com/cquery-project/cquery/wiki/Getting-started) will open a wiki edit page. `/Getting-started` should be removed.


tsangint:
I signed it!

PR #116, opened 2019-01-03
Fix typo in trash talk

hammacher:
There is a stray single quote at the end of the author line.


hammacher:
Too slow: https://github.com/v8/v8.dev/pull/115 :/

PR #122, opened 2019-01-04
document Torque's transient types

tebbi:


src/docs/torque.md
```diff
@@ -229,6 +229,33 @@ type CompareBuiltinFn = builtin(Context, Object, Object, Object) => Number;
...
+
+```torque
+extern transitioning macro Call(implicit context: Context)
+                               (Callable, Object): Object;
+```
+
+The way this is policed in the type system is that it is illegal to access a value of a transient type across a transitioning operation.
+
+```torque
+const fastArray : FastJSArry = Cast<FastJSArray>(array);
```
dannoc: This code actually doesn't compile, you need the otherwise... not sure how important that is, tho

mathiasbynens: typo: `FastJSArry` --> `FastJSArray`

src/docs/torque.md
```diff
@@ -229,6 +229,33 @@ type CompareBuiltinFn = builtin(Context, Object, Object, Object) => Number;
...
+
+```torque
+// A HeapObject with a JSArray map, and either fast packed elements, or fast
+// holey elements when the global NoElementsProtector is not invalidated.
+transient type FastJSArray extends JSArray
+    generates 'TNode<JSArray>';
+```
+
+For example, in the case of `FastJSArray`, the transient type is invalidated if the array changes to dictionary elements or if the global NoElementsProtector is invalidated.
+The way to express this in Torque is by annotating all callables that could potentially do that as `transitioning`. For example, calling a Javascript function can executy arbitrary Javascript, so it is `transitioning`.
```
peterwmwong: Typo: executy -> execute

tebbi:
Thanks for the reviews everyone!

PR #124, opened 2019-01-06
Change Google Chrome to Chromium and derivatives

ryzokuken:
/cc @bmeurer @mathiasbynens

### DIFF
--- a/src/index.njk
+++ b/src/index.njk
@@ -5,7 +5,7 @@ description: 'V8 is Google’s open source high-performance JavaScript and WebAs
 ---
 <h1>What is V8?</h1>
 
-<p>V8 is Google’s open source high-performance JavaScript and WebAssembly engine, written in C++. It is used in Google Chrome, the open source browser from Google, and in Node.js, among others. It implements <a href="https://tc39.github.io/ecma262/">ECMAScript</a> and <a href="https://webassembly.github.io/spec/core/">WebAssembly</a>, and runs on Windows 7 or later, macOS 10.12+, and Linux systems that use x64, IA-32, ARM, or MIPS processors. V8 can run standalone, or can be embedded into any C++ application.</p>
+<p>V8 is Google’s open source high-performance JavaScript and WebAssembly engine, written in C++. It is used in Chromium, the open source browser from Google and it's derivatives, and in Node.js, among others. It implements <a href="https://tc39.github.io/ecma262/">ECMAScript</a> and <a href="https://webassembly.github.io/spec/core/">WebAssembly</a>, and runs on Windows 7 or later, macOS 10.12+, and Linux systems that use x64, IA-32, ARM, or MIPS processors. V8 can run standalone, or can be embedded into any C++ application.</p>
 
 <h2>Latest blog posts</h2>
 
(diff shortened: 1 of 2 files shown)


src/index.njk
```diff
@@ -5,7 +5,7 @@ description: 'V8 is Google’s open source high-performance JavaScript and WebAs
 ---
 <h1>What is V8?</h1>
 
-<p>V8 is Google’s open source high-performance JavaScript and WebAssembly engine, written in C++. It is used in Google Chrome, the open source browser from Google, and in Node.js, among others. It implements <a href="https://tc39.github.io/ecma262/">ECMAScript</a> and <a href="https://webassembly.github.io/spec/core/">WebAssembly</a>, and runs on Windows 7 or later, macOS 10.12+, and Linux systems that use x64, IA-32, ARM, or MIPS processors. V8 can run standalone, or can be embedded into any C++ application.</p>
+<p>V8 is Google’s open source high-performance JavaScript and WebAssembly engine, written in C++. It is used in Chromium, the open source browser from Google and it's derivatives, and in Node.js, among others. It implements <a href="https://tc39.github.io/ecma262/">ECMAScript</a> and <a href="https://webassembly.github.io/spec/core/">WebAssembly</a>, and runs on Windows 7 or later, macOS 10.12+, and Linux systems that use x64, IA-32, ARM, or MIPS processors. V8 can run standalone, or can be embedded into any C++ application.</p>
```
mathiasbynens: it's --> its

mathiasbynens:
Thanks for your contribution, @ryzokuken!

I don't see the problem with using "Chrome" instead of "Chromium". What problem does this change fix?

The currently suggested change makes the description unnecessarily long IMHO. I don't think we should be mentioning "derivatives" as it's implicitly clear: if Chromium depends on V8, then Chromium consumers also use V8.

I'm inclined to close this PR. @natorion @danno

bmeurer:
@mathiasbynens I was reviewing slides from @ryzokuken where he was using this phrasing and suggested to replace *Chrome* with *Chromium*, since *Chrome* is **not** open source.

natorion:
I think the change to Chromium makes sense. The mentioning of the derivatives seems implicit if one forks Chromium.

mathiasbynens:
If we want to make this change, we should change https://v8.dev/docs as well.

ryzokuken:
> The currently suggested change makes the description unnecessarily long IMHO. I don't think we should be mentioning "derivatives" as it's implicitly clear: if Chromium depends on V8, then Chromium consumers also use V8.

> I think the change to Chromium makes sense. The mentioning of the derivatives seems implicit if one forks Chromium.

I realize that, and will remove the "derivatives" phrasing. 😅 

> If we want to make this change, we should change v8.dev/docs as well.

On it.

jakobkummerow:
Also, to be nit-picky, "Chromium, the open source browser from Google" is factually incorrect. "Google Chrome" is "the [open source] browser from Google". Chromium is a browser development project whose output is source code from which other entities can build a browser product.

In that sense, my (weak) preference would be the original wording. "Google Chrome" is a product that people know and recognize, so it makes for a good example; whereas who has ever heard of "Chromium"?

ryzokuken:
@mathiasbynens sounds fair.

@bmeurer I guess you'll be happy with this too?

mathiasbynens:
Thank you, @ryzokuken!

PR #132, opened 2019-01-16
Reverse list numbering of posts

mlippautz:


mlippautz:
Whoops, yes, the main reason was persistent numbering of posts.

Personally, I also think it highlights the team's long-term commitment in this area.

PR #154, opened 2019-03-18
docs: add cli options

aymen94:


aymen94:
I signed it!

jakobkummerow:
I'm not sure it is useful to "document" (with very little detail) *all* the command-line options. Many of them are already outdated, in the sense that they don't even exist any more in the latest version -- which implies that even just keeping this list updated will create ongoing churn and maintenance burden; if we do want a dump of `--flags` somewhere online then it should be auto-generated from every build. Also, many flags are very much for internal use, e.g. various tracing flags whose output is really only useful for people who know the code (and the respective flag) very well, so listing them all mostly creates noise and makes the list more confusing.

I think it would be more useful to select an *interesting* set of flags that are somewhat likely to be useful to folks, and document those with more background on what they are useful for, and/or how to use them and make sense of the output. We can then add a generic hint that calling `--help` will list the full set of implemented flags.

aymen94:
@jakobkummerow Good idea, Which think should remain?

mathiasbynens:
Thanks for the PR, @aymen94!

If we want to host the full list, we should automate it somehow (because no one is going to maintain it manually). However, as @jakobkummerow points out, that may not be super useful anyhow. I’m closing this PR.

PR #157, opened 2019-03-22
Add 3 links that hopefully save someone else a web search

tomayac:
CC: @mathiasbynens


mathiasbynens:
Thanks! (If we were to link these terms everywhere, we have a lot of work to do…)

tomayac:
Yeah, I guess the target audience is folks with more compiler optimization background than me.

PR #172, opened 2019-05-08
Add information on how to become an OWNER

natorion:


pavelfeldman:
Is there a reason why v8 policy needs to be different from them one for Chromium? My arrangement with Chromium is similar to the one with WebKit - as long as I am active and maintain reviewing capacity, I can stay a reviewer / OWNER.

https://chromium.googlesource.com/chromium/src/+/HEAD/docs/code_reviews.md

A hard requirement to be a Googler looks out of line with the rest of the letter and spirit of that policy.

PR #179, opened 2019-05-17
typo: Double word "all"

nschonni:


mathiasbynens:
Thanks! Did you use a tool to find all these? Sounds like we should add it to our lint checks :)

nschonni:
This was https://github.com/errata-ai/vale with just the "duplicate word" check running. There is still one for "really really" that I left which is why it can be hard to add it to CI sometimes

PR #183, opened 2019-05-19
chore: Use Travis Deploy Script provider

nschonni:


mathiasbynens:
Thanks!

mathiasbynens:
This fails because it can’t find the `firebase`: https://travis-ci.com/v8/v8.dev/builds/112367736#L707 I’ll revert. Let’s just stick to what we have, it works fine.

nschonni:
No problem. Might be good to ask if Travis is going to provide Firebase as a provider

mathiasbynens:
They do already support Travis as a provider, but I ran into similar issues when I tried that. The current approach is the first thing I tried that actually worked 🤷

PR #184, opened 2019-05-19
fix: Quote glob for Markdownlint

nschonni:
Markdownlint globs require the quotes to be expanded.
Address the flagged heading start issues by reflowing from H1->H2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "lint": "markdownlint src/**/*.md",
+    "lint": "markdownlint \"src/**/*.md\"",
     "eleventy": "eleventy",
     "html": "html-minifier --config-file=.html-minifier.json --input-dir=dist --output-dir=dist --file-ext=html",
     "css": "postcss src/_css/main.css --output dist/_css/main.css",
--- a/src/404.md
+++ b/src/404.md
@@ -4,6 +4,6 @@ permalink: /404.html
 excludeFromSitemap: true
 layout: layouts/base.njk
 ---
-# 404: document not found
+## 404: document not found
 
 Sorry, we couldn’t find that page.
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "lint": "markdownlint src/**/*.md",
+    "lint": "markdownlint \"src/**/*.md\"",
```
mathiasbynens: Without the quotes, this is just a Bash glob that should expand to the same list of files. That is,

```sh
echo src/**/*.md
```

…in the root of the repo prints the list of files, including e.g. `src/404.md`. Therefore, I wouldn’t have expected this change to make a difference.

Why does it?

src/404.md
```diff
@@ -4,6 +4,6 @@ permalink: /404.html
 excludeFromSitemap: true
 layout: layouts/base.njk
 ---
-# 404: document not found
+## 404: document not found
```
mathiasbynens: These should all remain `<h1>`s.

package.json
nschonni: I think without the quotes it is treated as a glob that only treats the "**" as a single folder level glob, with it it does expand it. Not saying it hasn't bit me a few times before, but I've seen this behaviour on other NodeJS CLI tools

src/404.md
nschonni: It can either be ignored inline with `<!-- markdownlint-disable ... -->`, or could be removed if you have a separate layout that doesn't already have an H1 like the base.nkj does

mathiasbynens: Either of those solutions sounds good to me.

RReverser:
Hi @nschonni and thanks for your PR!

I've created a new PR #219 that incorporated your package.json fix, but other changes became unnecessary as Markdownlint has evolved in the meanwhile and now allows to correctly handle `title:` vs headings clashes, so I've just changed the config instead.

Thanks again!

PR #206, opened 2019-07-09
Update elements-kind.md

jakobkummerow:
Moved "avoid creating holes" to the end of the "performance tips" section, because it usually isn't important.
Added another example to "avoid OOB reads" section.


mathiasbynens:
Thanks!

PR #219, opened 2019-08-13
Fix and upgrade Markdownlint config

RReverser:
As #184 has shown, the lint command was ignored on Linux due to specifics of glob expansion. Now that the glob is quoted, it works correctly across all platforms and actually reports errors.

Also, Markdownlint had issues with MD002 rule, which was deprecated in the new version in favour of MD041, so I upgraded Markdownlint to pick up these changes.

Also, MD025 was reporting violations for clash between YAML `title:` and the actual heading, which is now disabled in the config.

This supersedes #184.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "lint": "markdownlint src/**/*.md",
+    "lint": "markdownlint \"src/**/*.md\"",
     "eleventy": "eleventy",
     "html": "html-minifier --config-file=.html-minifier.json --input-dir=dist --output-dir=dist --file-ext=html",
     "css": "postcss src/_css/main.css --output dist/_css/main.css && postcss src/_css/feature-support.css --output dist/_css/feature-support.css",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,6 +49,7 @@
     "markdown-it-attrs": "^3.0.0",
     "markdown-it-container": "^2.0.0",
     "markdown-it-footnote": "^3.0.2",
+    "markdownlint": "^0.16.0",
```
mathiasbynens: Is this necessary?

RReverser: Good catch, probably not, I upgraded `markdownlint` from command line, when it should have been `markdownlint-cli`.

RReverser: Actually... looks like it's already latest. Weird, as the rule should be disabled then.

**UPD**: Ah, the reason was that there was explicit `"MD002": ...` line in the config which re-enabled the rule, but now that it's removed, everything works as intended.

RReverser: Fixed.

PR #220, opened 2019-08-13
Use shelljs/shx for Windows build support

RReverser:
Make sure that the Website can be built on Windows natively.


package.json
```diff
@@ -5,13 +5,13 @@
...
-    "js": "mkdir -p dist/_js && minify --mangle.topLevel src/_js/main.mjs --outFile dist/_js/main.mjs && minify --mangle.topLevel src/_js/legacy.js --outFile dist/_js/legacy.js",
+    "js": "shx mkdir -p dist/_js && minify --mangle.topLevel src/_js/main.mjs --outFile dist/_js/main.mjs && minify --mangle.topLevel src/_js/legacy.js --outFile dist/_js/legacy.js",
     "json": "json-minify src/.webmanifest > dist/.webmanifest",
     "xml": "minify-xml --in-place dist/sitemap.xml && minify-xml --in-place dist/blog.atom && minify-xml --in-place dist/features.atom",
     "sw": "workbox injectManifest workbox-config.js && minify dist/sw.js --outFile dist/sw.js",
-    "copy-img": "for browser in chrome firefox; do cp ./node_modules/@browser-logos/${browser}/${browser}.svg src/_css/img/${browser}.svg; done; cp ./node_modules/@browser-logos/safari/safari_256x256.png src/_css/img/safari.png",
+    "copy-img": "shx cp ./node_modules/@browser-logos/{chrome,firefox}/*.svg src/_css/img/ && shx cp ./node_modules/@browser-logos/safari/safari_256x256.png src/_css/img/safari.png",
     "img": "node add-intrinsicsize.js",
-    "clean": "rm -rf -- dist",
+    "clean": "shx rm -rf dist",
```
mathiasbynens: Does the `--` need to be removed?

RReverser: I don't think so / not sure, but it's not necessarily there either.

mathiasbynens: Call me paranoid, but I'd prefer keeping it, in case we ever need to add `*` to the arguments.

RReverser:
> although I haven't been able to test this due to no Windows machine

@mathiasbynens I've actually made an attempt with Travis changes to enable macOS + Windows CI testing, but given how slow `npm install` is on Windows version, figured you might not want to wait that long, so removed it back. Let me know if you think it's still worth it.

mathiasbynens:
> @mathiasbynens I've actually made an attempt with Travis changes to enable macOS + Windows CI testing, but given how slow `npm install` is on Windows version, figured you might not want to wait that long, so removed it back. Let me know if you think it's still worth it. https://travis-ci.com/v8/v8.dev/builds/123004951

How slow are we talking exactly?

It sounds useful. For emergency PRs, V8 org admins like you and I can always bypass CI and merge.

RReverser:
> How slow are we talking exactly?

That's why I included a link to the job, so that you could check the times on all systems :)

package.json
RReverser: I mean... `*` would work too (why wouldn't it?). But ok, I can add it back.

mathiasbynens:
So builds will take a factor of ~7x longer. That seems fine to me. Let's do it! We can always revisit if it turns out to bother us.

RReverser:
@mathiasbynens Ok, added back. I suspect it might be better than that sometimes (in that particular case looks like `npm install` failed midway and restarted as another attempt), but we'll see.

package.json
RReverser: Fixed.

RReverser:
@mathiasbynens Re-requested review as I'd prefer to have your look at the new changes.

.travis.yml
```diff
@@ -1,7 +1,12 @@
 language: node_js
+node_js: node # needs to be explicitly specified for nvs on Windows
```
mathiasbynens: Will this still use the version specified in `.nvmrc`?

RReverser: Oh. Probably no.

RReverser: Hmm, I wonder what's the best way forward then... `nvs` (which Travis uses as a replacement for `nvm` on Windows) doesn't read `.nvmrc`... Should I duplicate the version in the Travis config then?

RReverser: Looks like `nvs` looks for file `.node-version` as per https://github.com/jasongin/nvs/issues/113, and I could try to just symlink one into another and add that to the repo... But feels hacky.

mathiasbynens: Hacky indeed but I’d prefer that over duplicating the version number in two places.

mathiasbynens: Would we also have to update the README to add a Windows alternative to `nvm use`?

RReverser: Ugh they're not joking when they say Windows is in beta. Looks like that won't work either, because they use `nvs add` and `.node-version` is read only with `nvs auto`, and customising this command requires basically removing `language:` and writing completely custom install steps in `.travis.yml`. I'm not sure that ends up easier than duplicating version number.

mathiasbynens: > Ugh they're not joking when they say Windows is in beta.

In that case, let's not bother with Windows support on CI for now.

RReverser: Done, I left it commented out with an explainer to make tracking of upstream issue easier. Also left macOS CI enabled just in case.

PR #223, opened 2019-08-13
Enable MD033

nschonni:
Enable MD033 with explicit allowed elements. This caught one unescaped non-HTML element

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -20,7 +20,6 @@
       "img",
       "kbd",
       "pre",
-      "reason",
       "sub",
       "sup"
     ]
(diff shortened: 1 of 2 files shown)


.markdownlint.json
```diff
@@ -6,6 +6,23 @@
...
+      "div",
+      "em",
+      "figcaption",
+      "figure",
+      "i",
+      "img",
+      "kbd",
+      "pre",
+      "sub",
+      "sup"
```
mathiasbynens: I wonder how this passes even though `iframe` is missing from the list (and we use it for YouTube embeds in a few places). Any ideas?

nschonni: Could be a bug or part of the regex. I did notice the iframe would get flagged if I removed "div" since that was the element nesting them AFAIK

mathiasbynens: Interesting. I guess it only covers "outer" elements.

nschonni: Some of the inner ones seem to be flagged like "em", but maybe it's just the direct decendants of an outer element. Or just a regex bug :)

PR #228, opened 2019-08-23
Clarify recommendations regarding .mjs

GeoffreyBooth:
Resolves https://github.com/google/WebFundamentals/issues/7623.

The `.mjs` extension is a perfectly valid recommendation for development, but users should not be encouraged to publish JavaScript files with an `.mjs` extension unless they’re sure that such files will be served with the proper MIME type.

### DIFF
--- a/src/features/dynamic-import.md
+++ b/src/features/dynamic-import.md
@@ -42,7 +42,7 @@ Here’s how to statically import and use the `./utils.mjs` module:
 ```
 
 :::note
-**Note:** The previous example uses the `.mjs` extension to signal that it’s a module rather than a regular script. On the web, file extensions don’t really matter, as long as the files are served with the correct MIME type (e.g. `text/javascript` for JavaScript files) in the `Content-Type` HTTP header. The `.mjs` extension is [especially useful](https://github.com/nodejs/node-eps/blob/master/002-es-modules.md#32-determining-if-source-is-an-es-module) on other platforms such as Node.js, where there’s no concept of MIME types or other mandatory hooks such as `type="module"` to determine whether something is a module or a regular script. We’re using the same extension here for consistency across platforms and to clearly make the distinction between modules and regular scripts.
+**Note:** The previous example uses the `.mjs` extension to signal that it’s a module rather than a regular script. On the web, file extensions don’t really matter, as long as the files are served with the correct MIME type (e.g. `text/javascript` for JavaScript files) in the `Content-Type` HTTP header.
 :::
 
 This syntactic form for importing modules is a *static* declaration: it only accepts a string literal as the module specifier, and introduces bindings into the local scope via a pre-runtime “linking” process. The static `import` syntax can only be used at the top-level of the file.
--- a/src/features/dynamic-import.md
+++ b/src/features/dynamic-import.md
@@ -42,7 +42,7 @@ Here’s how to statically import and use the `./utils.mjs` module:
 ```
 
 :::note
-**Note:** The previous example uses the `.mjs` extension to signal that it’s a module rather than a regular script. On the web, file extensions don’t really matter, as long as the files are served with the correct MIME type (e.g. `text/javascript` for JavaScript files) in the `Content-Type` HTTP header.
+**Note:** The previous example uses the `.mjs` extension to signal that it’s a module rather than a regular script. On the web, file extensions don’t really matter, as long as the files are served with the correct MIME type (e.g. `text/javascript` for JavaScript files) in the `Content-Type` HTTP header. The `.mjs` extension is especially useful on other platforms such as [Node.js](https://nodejs.org/api/esm.html#esm_enabling) and [`d8`](/docs/d8), where it’s a convenient way to indicate to the runtime that your code is a module. We’re using the same extension here for consistency across platforms and to clearly make the distinction between modules and regular scripts.
 :::
 
 This syntactic form for importing modules is a *static* declaration: it only accepts a string literal as the module specifier, and introduces bindings into the local scope via a pre-runtime “linking” process. The static `import` syntax can only be used at the top-level of the file.
--- a/src/features/modules.md
+++ b/src/features/modules.md
@@ -126,9 +126,17 @@ Another difference relates to the `async` attribute, which causes the script to 
 
 You may have noticed we’re using the `.mjs` file extension for modules. On the Web, the file extension doesn’t really matter, as long as the file is served with [the JavaScript MIME type `text/javascript`](https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages:javascript-mime-type). The browser knows it’s a module because of the `type` attribute on the script element.
 
-During development, the `.mjs` file extension makes it crystal clear that a file is a module as opposed to a regular script. (It’s not always possible to tell just by looking at the code.) As mentioned before, modules are treated differently than regular scripts, so the difference is hugely important! Additionally, you may want to configure your editor to treat `.mjs` files as `.js` files to get syntax highlighting. Most modern editors already do this by default.
+In production, you should name your files with an `.mjs` extension only if you are sure that they will be served with the required `Content-Type: text/javascript` header. If your host lacks such support, or if you are publishing open source JavaScript for public use by others, it’s safer to use the `.js` extension.
 
-In production, you should save your files with an `.mjs` extension only if you are sure that they will be served with the required `Content-Type: text/javascript` header. If your host lacks such support, or if you are publishing open source JavaScript for public use by others, it’s safer to use the `.js` extension.
+During development, we recommend using the `.mjs` extension for modules, for two reasons:
+
+1. The `.mjs` extension makes it crystal clear to you and anyone else looking at your project that the file is a module as opposed to a regular script. (It’s not always possible to tell just by looking at the code.) As mentioned before, modules are treated differently than regular scripts, so the difference is hugely important!
+
+1. It ensures that your file is parsed as a module by runtimes such as [Node.js](https://nodejs.org/api/esm.html#esm_enabling) and [`d8`](/docs/d8), and build tools such as [Babel](https://babeljs.io/docs/en/options#sourcetype). While these environments and tools each have proprietary ways via configuration to interpret `.js` files as modules, the `.mjs` extension is the cross-compatible way to ensure that files are treated as modules.
+
+:::note
+**Note:** You may want to configure your editor to treat `.mjs` files as `.js` files to get syntax highlighting. Most modern editors already do this by default.
+:::
 
 ### Module specifiers { #specifiers }
 
--- a/src/features/modules.md
+++ b/src/features/modules.md
@@ -126,14 +126,9 @@ Another difference relates to the `async` attribute, which causes the script to 
 
 You may have noticed we’re using the `.mjs` file extension for modules. On the Web, the file extension doesn’t really matter, as long as the file is served with [the JavaScript MIME type `text/javascript`](https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages:javascript-mime-type). The browser knows it’s a module because of the `type` attribute on the script element.
 
-Still, we recommend using the `.mjs` extension for modules, for two reasons:
+During development, the `.mjs` file extension makes it crystal clear that a file is a module as opposed to a regular script. (It’s not always possible to tell just by looking at the code.) As mentioned before, modules are treated differently than regular scripts, so the difference is hugely important! Additionally, you may want to configure your editor to treat `.mjs` files as `.js` files to get syntax highlighting. Most modern editors already do this by default.
 
-1. During development, it makes it crystal clear that the file is a module as opposed to a regular script. (It’s not always possible to tell just by looking at the code.) As mentioned before, modules are treated differently than regular scripts, so the difference is hugely important!
-1. It’s consistent with Node.js, where [the experimental modules implementation](https://nodejs.org/api/esm.html) only supports files with the `.mjs` extension by default.
-
-:::note
-**Note:** To deploy `.mjs` on the web, your web server needs to be configured to serve files with this extension using the appropriate `Content-Type: text/javascript` header, as mentioned above. Additionally, you may want to configure your editor to treat `.mjs` files as `.js` files to get syntax highlighting. Most modern editors already do this by default.
-:::
+In production, you should save your files with an `.mjs` extension only if you are sure that they will be served with the required `Content-Type: text/javascript` header. If your host lacks such support, or if you are publishing open source JavaScript for public use by others, it’s safer to use the `.js` extension.
 
 ### Module specifiers { #specifiers }
 


src/features/modules.md
```diff
@@ -126,14 +126,9 @@ Another difference relates to the `async` attribute, which causes the script to
...
 
-1. During development, it makes it crystal clear that the file is a module as opposed to a regular script. (It’s not always possible to tell just by looking at the code.) As mentioned before, modules are treated differently than regular scripts, so the difference is hugely important!
-1. It’s consistent with Node.js, where [the experimental modules implementation](https://nodejs.org/api/esm.html) only supports files with the `.mjs` extension by default.
-
-:::note
-**Note:** To deploy `.mjs` on the web, your web server needs to be configured to serve files with this extension using the appropriate `Content-Type: text/javascript` header, as mentioned above. Additionally, you may want to configure your editor to treat `.mjs` files as `.js` files to get syntax highlighting. Most modern editors already do this by default.
```
mathiasbynens: Why remove this?

src/features/dynamic-import.md
```diff
@@ -42,7 +42,7 @@ Here’s how to statically import and use the `./utils.mjs` module:
 ```
 
 :::note
-**Note:** The previous example uses the `.mjs` extension to signal that it’s a module rather than a regular script. On the web, file extensions don’t really matter, as long as the files are served with the correct MIME type (e.g. `text/javascript` for JavaScript files) in the `Content-Type` HTTP header. The `.mjs` extension is [especially useful](https://github.com/nodejs/node-eps/blob/master/002-es-modules.md#32-determining-if-source-is-an-es-module) on other platforms such as Node.js, where there’s no concept of MIME types or other mandatory hooks such as `type="module"` to determine whether something is a module or a regular script. We’re using the same extension here for consistency across platforms and to clearly make the distinction between modules and regular scripts.
+**Note:** The previous example uses the `.mjs` extension to signal that it’s a module rather than a regular script. On the web, file extensions don’t really matter, as long as the files are served with the correct MIME type (e.g. `text/javascript` for JavaScript files) in the `Content-Type` HTTP header.
```
mathiasbynens: Why remove the last part of the paragraph? It’s all still relevant, AFAIK.

src/features/modules.md
mathiasbynens: > The `.mjs` extension is a perfectly valid recommendation for development, but users should not be encouraged to publish JavaScript files with an `.mjs` extension unless they’re sure that such files will be served with the proper MIME type.

The note which this patch removes already says this.

src/features/modules.md
```diff
@@ -126,14 +126,9 @@ Another difference relates to the `async` attribute, which causes the script to
...
 
-Still, we recommend using the `.mjs` extension for modules, for two reasons:
+During development, the `.mjs` file extension makes it crystal clear that a file is a module as opposed to a regular script. (It’s not always possible to tell just by looking at the code.) As mentioned before, modules are treated differently than regular scripts, so the difference is hugely important! Additionally, you may want to configure your editor to treat `.mjs` files as `.js` files to get syntax highlighting. Most modern editors already do this by default.
 
-1. During development, it makes it crystal clear that the file is a module as opposed to a regular script. (It’s not always possible to tell just by looking at the code.) As mentioned before, modules are treated differently than regular scripts, so the difference is hugely important!
-1. It’s consistent with Node.js, where [the experimental modules implementation](https://nodejs.org/api/esm.html) only supports files with the `.mjs` extension by default.
```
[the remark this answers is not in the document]
WebReflection: about the first point, in commonjs era nobody ever had an issue in writing modules as `.js` script, so that sentence is overly opinionated/forviant for no reason

the last point is not consistent, as NodeJS is currently already capable of loading `.js` files as ESM

src/features/dynamic-import.md
WebReflection: https://nodejs.org/api/esm.html#esm_enabling shows that `.mjs` doesn't solve cases such as input code / pipe and there is a flag to simply enable that as ESM.

the whole CommonJS ecosystem grew and survived with everyone writing modules as `.js` without ever being confused.

Underlying that `.mjs` solves modules from regular scripts ambiguity, which was never the case in CommonJS, which also never existed on the Web, is misleading, as many developers never had that issue since you need to know what you are importing when importing something, you don't import files by accident.

src/features/modules.md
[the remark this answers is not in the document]
mathiasbynens: > about the first point, in commonjs era nobody ever had an issue in writing modules as .js script, so that sentence is overly opinionated/forviant for no reason

Please stay on topic. CommonJS "modules" are not ECMAScript modules, as in JavaScript programs that are parsed using the Module Goal. The article describes how modules are treated differently from classic scripts, and this section refers to that. 

> the last point is not consistent, as NodeJS is currently already capable of loading .js files as ESM https://nodejs.org/api/esm.html#esm_enabling

It is with additional configuration, but not by default, as our current article states.

src/features/dynamic-import.md
mathiasbynens: I'm not interested in rehashing the whole `.mjs` debate here. Again, let's keep this discussion on topic, and focus on the merits of this patch.

src/features/modules.md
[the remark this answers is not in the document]
WebReflection: additional configuration does not need `.mjs` so that sentence should at least be modified:

> It’s consistent with Node.js, where the experimental modules implementation only supports, _unless configured differently_, files with the `.mjs` extension by default

the current sentence bases an argument over an _experimental_ flag, which should never be encouraged in references, and sounds like the only option, easily misleading readers that `.mjs` is consistent with NodeJS, which won't likely be the common case once the flag disappears.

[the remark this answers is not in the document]
mathiasbynens: That's what "by default" means. You have a file, and you want to run it as a module in Node.js. If it's `.mjs`, it just works. If not, you need additional configuration.

src/features/dynamic-import.md
WebReflection: It is in topic: the `.mjs` extension is **not** especially useful because it doesn't separate, as example, applications/programs by module.

The entry point of a web page is not a module that exports anything, it's an application/program that might use the native module system or not.

```html
<!doctype html>
<script type="module" src="app.js"></script>
```

If `app.js` is named `app.mjs` instead, reading the discussed argument, it means that `app.js` is a module that would export something.

That's 99% of the time not the case. `app.js` or any `main` / `index` / `entry` point on the Web is not a module per-se, so saying that `.mjs` solves ambiguity, when it has more ambiguity than a normal script, is misleading.

Entry points are scripts that might, or might not need, modules. The `.mjs` extension there becomes less semantic than a file that is a real module, exporting actually things.

mathiasbynens: @WebReflection Despite [repeated](https://github.com/v8/v8.dev/pull/228#discussion_r317018300) [requests](https://github.com/v8/v8.dev/pull/228#discussion_r317019777) to the contrary, you continue to derail this conversation with strawman arguments and off-topic discussions. This behavior is inappropriate. Please note that [the Chromium Code of Conduct](https://v8.dev/terms#code-of-conduct) applies to this repository.

I'm trying to discuss the changes in @GeoffreyBooth's patch, and your actions are distracting from that. Let's try to work together and stay on topic.

src/features/modules.md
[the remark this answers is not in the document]
WebReflection: if you have `{"type": "module"}` in `/package.json` the default is another, same if you received your laptop configured by somebody responsible for laptops in the company.

I understand that sentence and all I am saying is that, as is, it's easily misleading and underlying that `.mjs` might be not necessary in some case would improve it.

If not, it feels like this whole document is too focused on telling people to use `.mjs`, which should be a developer choice, not something sold as subjective best/practice.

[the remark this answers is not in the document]
mathiasbynens: No one is claiming that it is necessary in all cases, though -- in fact, the article already includes a counter-example (it explicitly states that on the web, extensions don't matter). Once again, you're putting up a strawman argument that I'm not interested in getting into. Please stop.

src/features/dynamic-import.md
WebReflection: I am trying to explain why that sentence is misleading. Last attempt:

> The `.mjs` extension is especially useful on other platforms such as Node.js, where there’s no concept of MIME types or other mandatory hooks such as `type="module"`

Others platforms have the module flag, only NodeJS doesn't. SpiderMonkey and JSC can both use the `-m` flag, and the Web has `type="module"`.

If that point wants to be relevant, shouldn't it include which platform misses the flag to opt in for modules?

> We’re using the same extension here for consistency across platforms and to clearly make the distinction between modules and regular scripts.

All platforms have a flag to switch parsing goal, which is not solved by `.mjs` because Web needs mime type, SpiderMonkey still wants `-m` flag, and so does JSC.

So, what are these other platforms that would need `.mjs`, beside NodeJS?

If there was a list of other platforms it'd be a great point to keep, but as it is, it looks like a supporting argument for the sake of it.

src/features/dynamic-import.md
```diff
@@ -42,7 +42,7 @@ Here’s how to statically import and use the `./utils.mjs` module:
 ```
 
 :::note
-**Note:** The previous example uses the `.mjs` extension to signal that it’s a module rather than a regular script. On the web, file extensions don’t really matter, as long as the files are served with the correct MIME type (e.g. `text/javascript` for JavaScript files) in the `Content-Type` HTTP header. The `.mjs` extension is [especially useful](https://github.com/nodejs/node-eps/blob/master/002-es-modules.md#32-determining-if-source-is-an-es-module) on other platforms such as Node.js, where there’s no concept of MIME types or other mandatory hooks such as `type="module"` to determine whether something is a module or a regular script. We’re using the same extension here for consistency across platforms and to clearly make the distinction between modules and regular scripts.
```
mathiasbynens: Seems like we should just keep this, and update the link to <https://nodejs.org/api/esm.html#esm_enabling>. We could also point out that `d8` recognizes `.mjs`, in case we want to add more examples than just Node.js.

src/features/modules.md
mathiasbynens: Same here. We can just keep this and update the link to <https://nodejs.org/api/esm.html#esm_enabling>.

src/features/dynamic-import.md
mathiasbynens: > Others platforms have the module flag, only NodeJS doesn't. SpiderMonkey and JSC can both use the `-m` flag, and the Web has `type="module"`.

JSC also recognizes `.mjs` in which case you don't need the `-m` flag. Same for `d8` where with `.mjs` you don't need the `--module` flag. Same for `deno`. Are you suggesting we include more non-Node.js examples?

> If that point wants to be relevant, shouldn't it include which platform misses the flag to opt in for modules?

I'm open to adding more non-Node.js examples, although I don't feel it's super valuable in this context. We're talking about a **brief side note** in a large article on JS modules; I don't think anyone reasonably expects it to be an exhaustive resource of all non-Web environments that run JavaScript.

> If there was a list of other platforms it'd be a great point to keep, but as it is, it looks like a supporting argument for the sake of it (aka: NodeJS is the only inconsistent here, not the other way around)

The article specifically talks about "platforms where there's no concept of MIME types or other mandatory hooks". This includes all JS engine binaries. I don't think it's relevant which of those special-case `.mjs`, since the point is that `.mjs` can help in these general cases (as it already does in `d8`/`jsc`/`deno`), without any downsides.

WebReflection: > Are you suggesting we include more non-Node.js examples?

To keep it relevant, I think it should mention other non-NodeJS-related platforms that understands `.mjs` as is, even if these have also flags to opt in.

That would be IMO more informative than mentioning only NodeJS.

mathiasbynens: Ok, let's update this section to use the link you suggested + include `d8` as an additional example.

```suggestion
:::note
**Note:** The previous example uses the `.mjs` extension to signal that it’s a module rather than a regular script. On the web, file extensions don’t really matter, as long as the files are served with the correct MIME type (e.g. `text/javascript` for JavaScript files) in the `Content-Type` HTTP header. The `.mjs` extension is [especially useful](https://nodejs.org/api/esm.html#esm_enabling) on other platforms such as Node.js and [`d8`](/docs/d8), where there’s no concept of MIME types or other mandatory hooks such as `type="module"` to determine whether something is a module or a regular script. We’re using the same extension here for consistency across platforms and to clearly make the distinction between modules and regular scripts.
```

mathiasbynens:
> In general, I removed the notes about Node.js because they were written years ago when .mjs was going to be required to use ESM syntax in Node and that’s no longer the case. 

It was never the case. `.mjs` was never going to be required in Node.js — the plan has always been to allow for configuration to override the default.

But then again, the article doesn’t claim `.mjs` is required, so I’m not really sure what this patch is trying to fix.

> The diff makes things look like the changes were more dramatic than they were.

I still think it’s more than we need. All the original sentences that are affected by this patch are still accurate as-is and convey the same information. It’s just that some links are outdated (we should point to https://nodejs.org/api/esm.html#esm_enabling nowadays, as @WebReflection pointed out). And we could [mention `d8`](https://github.com/v8/v8.dev/pull/228#discussion_r317049841) as per @WebReflection’s request above. 

Do you want to amend this patch [accordingly](https://github.com/v8/v8.dev/pull/228#discussion_r317049841), or would you prefer I suggest these as changes?

GeoffreyBooth:
> the article doesn’t claim `.mjs` is required, so I’m not really sure what this patch is trying to fix.

I wrote up at the [top](https://github.com/v8/v8.dev/pull/228#issue-310205910) what I’m trying to fix: The `.mjs` extension is a perfectly valid recommendation for development, but users should not be encouraged to publish JavaScript files with an `.mjs` extension unless they’re sure that such files will be served with the proper MIME type. The recommendation is the issue I’m hoping to address.

The article is very direct: “we recommend using the `.mjs` extension for modules.” There are no qualifications; all users should use `.mjs`, period. There is no acknowledgement of use cases where `.mjs` would be inappropriate or not recommended; the closest the article comes is this sentence:

> To deploy `.mjs` on the web, your web server needs to be configured to serve files with this extension using the appropriate `Content-Type: text/javascript` header, as mentioned above.

In other words, the recommendation to use `.mjs` is so strong that the user should reconfigure their server in order to reap the benefits of `.mjs`; or presumably find a new host if they lack the ability to configure the current one. Using `.js` instead isn’t even mentioned as an option.

As I wrote in https://github.com/google/WebFundamentals/issues/7623, there are significant groups of users who lack control over their server configuration: they’re using shared hosting, or delivering files for a client who will host them, or they’re publishing an open source library (or code within something like a WordPress theme) where the eventual hosting environment is unknown. For such users, `.js` is the safest choice or possibly even the only option. By not acknowledging these users, in my opinion, the article implies that they’re doing something wrong (or worse, that they’re not worthy of acknowledgement: _pity those poor plebeians who can’t configure their own servers!)._

Rather than amending the recommendation with caveats, I think it’s simpler and more equitable to just not make a recommendation at all. Say what `.mjs` is and why it’s useful, and when it can and can’t be used, and leave it at that. The users who are in a position to be able to use it, and who find its advantages appealing, will use it; everyone else should feel fine using `.js` for ESM.

mathiasbynens:
The article already covers this:

> To deploy `.mjs` on the web, your web server needs to be configured to serve files with this extension using the appropriate `Content-Type: text/javascript` header, as mentioned above.

…and:

> On the web, file extensions don’t really matter […]

Again, I don’t think this needs any changes, and I am not in favor of this patch in its current form, which removes the general recommendation.

I would accept a patch making [the updates explained above](https://github.com/v8/v8.dev/pull/228#issuecomment-524363647).

GeoffreyBooth:
> I am not in favor of this patch in its current form, which removes the general recommendation.

I think it’s important to limit the general recommendation to development, and to make clear that production is a different story. Simply saying that extensions don’t matter to browsers and that servers need configuration is not the same as saying that some users should continue to use `.js` in certain scenarios. A naïve user might read the article as it’s currently written and think, “well I use shared hosting where `.mjs` isn’t supported. I guess I can’t use ESM syntax.” There’s nothing in the article suggesting a solution that would work for them, and that’s why I find the current language misleading.

There is also an equity issue here. You and I might not know any developers who lack the luxury of full control of their servers, or professional systems engineering support, but there are millions of JavaScript developers out there scraping by without such privileges. I think part of the passion that the `.mjs` debate entails is indignation on the part of this population that they’re being advised to use this new thing, and therefore they’re backward if they’re not using it; but their economic situation is such that they don’t have the resources to use it. Especially coming from the ivory tower of Google, I think it’s important that Google make recommendations that apply to everyone and not just the privileged.

src/features/dynamic-import.md
GeoffreyBooth: I’ve updated my revision per your note.

The “no concept of MIME types or other mandatory hooks such as `type="module"`” language isn’t quite correct. As you know, Node has `"type": "module"` now, which in fairness didn’t exist back when this sentence was written; but before you argue that the current text is still correct because of the “mandatory” qualifier, keep in mind that HTML’s `type="module"` isn’t mandatory either: a regular script can load a module via `import()`, with no `type="module"` in sight.

But that’s just one clause, and the paragraph is stronger without it. `.mjs`’s utility is that it indicates a file’s module-ness for _all_ non-browser environments and tools, rather than Node.js’ `"type": "module"` or Spidermonkey’s `-m` or the various proprietary configuration options in Babel and Webpack and so on. I think that point comes across more clearly without the digression into `.mjs`’s competitors.

src/features/modules.md
```diff
@@ -126,13 +126,16 @@ Another difference relates to the `async` attribute, which causes the script to
...
+
+1. It ensures that your file is parsed as a module by runtimes such as [Node.js](https://nodejs.org/api/esm.html#esm_enabling) and [`d8`](/docs/d8), and build tools such as [Babel](https://babeljs.io/docs/en/options#sourcetype). While these environments and tools each have proprietary ways via configuration to interpret `.js` files as modules, the `.mjs` extension is the cross-compatible way to ensure that files are treated as modules.
 
 :::note
-**Note:** To deploy `.mjs` on the web, your web server needs to be configured to serve files with this extension using the appropriate `Content-Type: text/javascript` header, as mentioned above. Additionally, you may want to configure your editor to treat `.mjs` files as `.js` files to get syntax highlighting. Most modern editors already do this by default.
+**Note:** You may want to configure your editor to treat `.mjs` files as `.js` files to get syntax highlighting. Most modern editors already do this by default.
```
GeoffreyBooth: I’ve revised this to be much closer to your original text. I still think it weakens the article to include a discussion of non-Web platforms, but if you want that included, I think this language is a fairer and stronger explanation of `.mjs`’s benefits than the previous text. Now that all ESM-supporting runtimes (both browsers and Node.js and d8 and Deno and Spidermonkey and JSC) support ESM in either `.js` or `.mjs` files, I think it’s more accurate to say that the benefit of `.mjs` is that it’s a _consistent_ way of declaring your code to be a module, in contrast to each runtime’s proprietary configuration.

The previous previous text (before you added “by default” in response to `"type": "module"`) seems like it was written back when it seemed like `.mjs` was the only way that ESM was going to be supported in Node.js, and therefore it was really important to persuade the world to adopt `.mjs` in order to help Node join the ESM bandwagon (and vice versa, to help prevent ESM adoption from being slowed by incompatibility with Node). I know you say that something like `"type": "module"` was always the plan, and I’ve heard that from others in the Node modules group, but I’ve scoured the Internet for any record of such history and I can’t find any. I would be very curious to read any if you have any links you could share.

src/features/dynamic-import.md
mathiasbynens: > The “no concept of MIME types or other mandatory hooks such as `type="module"`” language isn’t quite correct. As you know, Node has `"type": "module"` now, which in fairness didn’t exist back when this sentence was written; but before you argue that the current text is still correct because of the “mandatory” qualifier, keep in mind that HTML’s `type="module"` isn’t mandatory either: a regular script can load a module via `import()`, with no `type="module"` in sight.

Yeah, but the script doing the dynamic `import()` would be a classic script, not a module. The current phrasing is precise and correct.

mathiasbynens:
> I think it’s important to limit the general recommendation to development, and to make clear that production is a different story. Simply saying that extensions don’t matter to browsers and that servers need configuration is not the same as saying that some users should continue to use `.js` in certain scenarios. A naïve user might read the article as it’s currently written and think, “well I use shared hosting where `.mjs` isn’t supported. I guess I can’t use ESM syntax.” There’s nothing in the article suggesting a solution that would work for them, and that’s why I find the current language misleading.

There is though: “On the web, file extensions don’t really matter […]”. (I serve [my super old personal site](https://mathiasbynens.be/)’s scripts without any extensions, hah!) It feels like we keep going back and forth on this discussion rehashing the same arguments.

> There is also an equity issue here. You and I might not know any developers who lack the luxury of full control of their servers, or professional systems engineering support, but there are millions of JavaScript developers out there scraping by without such privileges. I think part of the passion that the `.mjs` debate entails is indignation on the part of this population that they’re being advised to use this new thing, and therefore they’re backward if they’re not using it; but their economic situation is such that they don’t have the resources to use it. Especially coming from the ivory tower of Google, I think it’s important that Google make recommendations that apply to everyone and not just the privileged.

This whole discussion is about a **side note** on a feature explainer on the V8 website. We recommend modern JavaScript techniques all the time. It doesn’t mean we don’t care about users that need to support legacy browsers or enviroments. When we say “use `Map` instead of plain objects [in situation `$x`]”, that’s not the same as recommending folks to drop support for environnments that lack `Map` support, and I don’t appreciate you implying that it is. Developers commonly use tooling such as transpilers and bundlers for their projects. The article we’re talking about explicitly mentions bundlers (and actively recommends the continued use of bundlers).

src/features/dynamic-import.md
GeoffreyBooth: Allow me to be blunt, if you don’t mind. The original sentence was this:

> The `.mjs` extension is especially useful on other platforms such as Node.js, where there’s no concept of MIME types or other hooks such as `type="module"` to determine whether something is a module or a regular script.

Then `"type": "module"` came to Node, and so that sentence was no longer true. So you [added](https://github.com/v8/v8.dev/commit/e3ba0f22395d4141a6f61ae8aa17f234c29da497#diff-356674ffcf237bcfbc2efdc7a09c7ea1R45) “mandatory” before “hooks.”

As currently written, “mandatory” is a weasel word. It makes the sentence true again (arguably) but it is misleading to the reader in that it implies that `.mjs` is the only way to use ESM in Node. No reasonable reader would read this sentence and assume that there was any way to use ESM in Node other than via `.mjs`. No one would read this and think, “oh, but does Node have any _non_-mandatory hooks?”

If you want to recommend `.mjs`, that’s fine; my revised text makes that case for you, without deceptive language.

GeoffreyBooth: P.S. to my last comment: I didn’t mean to imply that you were being deceptive in your edit. I only meant that it comes across that way to me, and I’m hardly unbiased. Apologies if I came across otherwise.

We’re debating the wrong thing. The question isn’t whether the current text is wrong or incorrect; the question should be whether the new text is _better._ So let’s compare:

> **Old:** The `.mjs` extension is especially useful on other platforms such as Node.js, where there’s no concept of MIME types or other mandatory hooks such as `type="module"` to determine whether something is a module or a regular script.

> **New:** The `.mjs` extension is especially useful on other platforms such as Node.js and `d8`, where it’s a convenient way to indicate to the runtime that your code is a module.

Both sentences offer a reason why `.mjs` is especially useful. The old sentence’s reason is why `.mjs` is useful _for platforms:_ `.mjs` is useful for such platforms to have a way to be able to recognize modules. The new sentence’s reason is why `.mjs` is useful _for users:_ this is a convenient way to signal to platforms that a file is a module.

As this article is written for users, I think the latter sentence is more appropriate.

mathiasbynens:
We still believe server software should feel encouraged to add first-class support for `.mjs`, and weakening our recommendations in this way would go against that.

In order to resolve this, I'll tweak your patch based on my feedback throughout this discussion and merge it in. Thanks for your work on this!

PR #229, opened 2019-08-23
Add feature explainer for optional chaining

MayaLekova:
Hey Mathias, can you PTAL?

### DIFF
--- a/src/features/optional-chaining.md
+++ b/src/features/optional-chaining.md
@@ -115,10 +115,10 @@ _Stacking_ means that more than one optional chaining operator can be applied on
 
 ```js
 // An optional chain may be followed by another optional chain.
-const firstNameLength = db?.users[42].names?.first.length;
+const firstNameLength = db.users?.[42].names?.first.length;
 ```
 
-Still, be considerate about using more than one optional chaining operator in a single chain. If a value is guaranteed to not be nullish, then using `?.` to access properties on it should be discouraged. For instance in the example above, `users` and `users[42]` are always considered to be defined.
+Still, be considerate about using more than one optional chaining operator in a single chain. If a value is guaranteed to not be nullish, then using `?.` to access properties on it should be discouraged. In the example above, `db` is considered to always be defined, but `db.users` and `db.users[42]` may not be. If the user has a `names` property, then `names.first.length` is assumed to always be defined.
 
 _Optional deletion_ means that the `delete` operator can be combined with an optional chain:
 
--- a/src/features/optional-chaining.md
+++ b/src/features/optional-chaining.md
@@ -115,10 +115,10 @@ _Stacking_ means that more than one optional chaining operator can be applied on
 
 ```js
 // An optional chain may be followed by another optional chain.
-const firstNameLength = db.users?.[42].names?.first.length;
+const firstNameLength = db.users?.[42]?.names.first.length;
 ```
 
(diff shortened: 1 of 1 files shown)


mathiasbynens:
Thanks for getting this ready! Let's publish this some time next week. I'll take it from here.

src/features/optional-chaining.md
```diff
@@ -0,0 +1,136 @@
...
+// with a dynamically generated property name.
+const optionName = 'optional setting';
+const optionLength = db?.user?.preferences?.[optionName].length;
+```
+
+This last form is also available for optionally indexing arrays, e.g.:
+
+```js
+// If the array is missing the user with the given index,
+// then `userName` gracefully evaluates to `undefined`.
```
jridgewell: This is incorrect. We're checking if `usersArray` is nullish, not the element at that index.

MayaLekova: Thanks, that was an important omission.

MayaLekova: Also, please check in the 2nd commit, where I added such a section.

src/features/optional-chaining.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+_Stacking_ means that more than one optional chaining operator can be applied on a sequence of property accesses:
+
+```js
+// An optional chain may be followed by another optional chain.
+const firstNameLength = db?.users[42].names?.first.length;
+```
+
+Still, be considerate about using more than one optional chaining operator in a single chain. If a value is guaranteed to not be nullish, then using `?.` to access properties on it should be discouraged. For instance in the example above, `users` and `users[42]` are always considered to be defined.
```
mathiasbynens: Thanks for adding the new section! I think we can tweak the example a little bit more:

```js
const firstNameLength = db?.users[42].names?.first.length;
```

In practice, it's probably more common for the root object itself to always be defined, but then any properties on it to not be. How about, instead of the above, something like:

```js
const firstNameLength = db.users?.[42].names?.first.length;
```

Then you could say:

> In the example above, `db` is considered to always be defined, but `db.users` and `db.users[42]` may not be. If the user has a `names` property, then `names.first.length` is assumed to always be defined.

WDYT?

MayaLekova: Thanks, PTAL at the new version.

src/features/optional-chaining.md
```diff
@@ -115,10 +115,10 @@ _Stacking_ means that more than one optional chaining operator can be applied on
 
 ```js
 // An optional chain may be followed by another optional chain.
-const firstNameLength = db?.users[42].names?.first.length;
+const firstNameLength = db.users?.[42].names?.first.length;
 ```
 
-Still, be considerate about using more than one optional chaining operator in a single chain. If a value is guaranteed to not be nullish, then using `?.` to access properties on it should be discouraged. For instance in the example above, `users` and `users[42]` are always considered to be defined.
+Still, be considerate about using more than one optional chaining operator in a single chain. If a value is guaranteed to not be nullish, then using `?.` to access properties on it should be discouraged. In the example above, `db` is considered to always be defined, but `db.users` and `db.users[42]` may not be. If the user has a `names` property, then `names.first.length` is assumed to always be defined.
```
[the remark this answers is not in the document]
MayaLekova: That was my omission, it seems Mathias meant the same. PTAL again, thanks a lot!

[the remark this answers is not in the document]
mathiasbynens: Yep, that's actually what I meant to propose the other day, but I got it wrong 😳 Thanks for spotting this, Justin!

PR #238, opened 2019-09-03
Update Firefox support for public class fields

Arkni:
Firefox added support for public class fields as of the release 69

See https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/69#JavaScript

### DIFF
--- a/src/features/class-fields.md
+++ b/src/features/class-fields.md
@@ -190,7 +190,7 @@ class Cat extends Animal {
 ### Support for public class fields { #support-public-fields }
 
 <feature-support chrome="72 /blog/v8-release-72#public-class-fields"
-                 firefox="no"
+                 firefox="69 https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/69#JavaScript"
                  safari="no"
                  nodejs="12 https://twitter.com/mathias/status/1120700101637353473"
                  babel="no"></feature-support>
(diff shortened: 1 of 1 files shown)


chicoxyzzy:
Unfortunately, there is no separation between instance and static public fields (latter is not implemented in Firefox 69) in feature support table, so this change is not quite correct. @mathiasbynens should "Support for public class fields" be splitted into "Support for instance public class fields" and "Support for static class fields"?

Arkni:
Hi @chicoxyzzy,

I read your reply in the Twitter thread and I saw the other PR.

Splitting the section into two implies that the other section about private class fields should be split too.

If splitting the section is not an option, I think as a solution, the component that renders the support listings can be adapted to show partial implementations as yellow/orange with possibility to add a note bellow or a tooltip.

chicoxyzzy:
> If splitting the section is not an option, I think as a solution, the component that renders the support listings can be adapted to show partial implementations as yellow/orange with possibility to add a note bellow or a tooltip.

Yes, I think that `partial` value for feature support is a good option. There is no such value, but I think it could be added easily https://github.com/v8/v8.dev/blob/master/feature-support.js

adalinesimonian:
Completely missed this PR while making one of my own. Thought it was a tiny change no one could have possibly noticed. Oh well, consider this a bump.

mathiasbynens:
I've added support for "partial feature support" in 725d6776f7f8656fbd16237075071773cdfa88a7, and have updated this patch to use it. This now looks good to go. Thanks, everyone!

Arkni:
Thanks a lot @mathiasbynens for working on this. And sorry I wasn't able to work on it due to some struggles the last few months.

PR #242, opened 2019-09-11
Add V8 Lite blog post

rmcilroy:


rmcilroy:
PTAL, thanks.

RReverser:
Minor nit for now: can you compress avatars with Squoosh (just to the point where they still look reasonable)? These seem quite a bit larger than most in the folder.

rmcilroy:
Good point, done.

PR #243, opened 2019-09-12
Replace v8-lite.svg with v8-lite.png

RReverser:
SVG mostly consisted from base64 PNG with some modifications, so this reduces size of the image from 108KB to 9KB.


rmcilroy:
I have a svg of the underlying chart, however I can't import it into the Google docs drawing (only copy and paste it) which seems to turn it into a png. Any ideas how to do this?

mathiasbynens:
I manually fiddled around with this in Sketch and fixed it here: https://github.com/v8/v8.dev/commit/d9c77bf788b09d3d447b3da34d1c93f914159fd3

rmcilroy:
Fantastic, thanks Mathias!

PR #250, opened 2019-09-25
Improve template link

bmeurer:
Change the link to the design document template to make it usable as a template directly.


mathiasbynens:
cc @natorion

PR #254, opened 2019-10-03
Add dark mode, fixes #252

### LINKED ISSUE #252: Add dark mode
We should be able to do better than `html, img, h1 a, #footer { filter: invert(1); }` 👻

Ugly screenshot of the above:

tomayac:
**This PR is an attempt at adding dark mode to v8.dev 🌒:**

- Adds a **toggle** to the footer that initially adheres to the user's `prefers-color-scheme` preference, but allows them to override it:
  Screen Shot 2019-10-03 at 15 01 27
  Screen Shot 2019-10-03 at 15 02 55
- Tries its best to **darken images** that are *not* photographic and *not* logos with the option to opt-out individual images by adding a class `no-darkening`:
  - Automatically darkened:
     Screen Shot 2019-10-03 at 15 04 11
  - Left in peace logo:
     Screen Shot 2019-10-03 at 15 10 58
  - Opted-out of darkening via ``:    
     Screen Shot 2019-10-03 at 15 08 22
- Dynamically pulls in the light or the dark **Twitter widget** (but doesn't automatically switch it yet):
   Screen Shot 2019-10-03 at 15 13 12
   Screen Shot 2019-10-03 at 15 12 56
- Slightly **re-colorizes photographic images** (avatars for now):
   Screen Shot 2019-10-03 at 15 53 43
   Screen Shot 2019-10-03 at 15 53 53


RReverser:
Haven't looked at the code yet, but this is extremely cool!

tomayac:
Full page screenshot, compare to the one from https://github.com/v8/v8.dev/issues/252:

localhost_8080_blog_v8-release-78_ (1)

src/_css/dark.css
```diff
@@ -0,0 +1,25 @@
...
+  --icon-no-support-filter: brightness(0.5) grayscale(100%);
+  --note-color: var(--darkestgray);
+  --note-background-color: var(--yellow);
+  --tag-color: var(--nearwhite);
+  --tag-background-color: var(--darkgray);
+  --logo-hover-background-color: var(--darkgray);
+}
+
+:root {
+  -webkit-font-smoothing: antialiased;
```
mathiasbynens: Do we need this?

mathiasbynens: Is this necessary?

src/_js/main.mjs
```diff
@@ -11,6 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import * as DarkModeToggle from '/_js/dark-mode-toggle.min.mjs';
```
mathiasbynens: Can this be just `import '/_js/dark-mode-toggle.min.mjs';`?

src/_includes/layouts/base.njk
```diff
@@ -4,6 +4,28 @@
...
+    <!--
+      Conditionally either load the light or the dark stylesheet. The matching file
+      will be downloaded with `highest`, the non-matching file with `lowest`
+      priority. If the browser doesn't support `prefers-color-scheme`, the media
+      query is unknown and the files are downloaded with `lowest` priority (but
+      above I already force `highest` priority for my default light experience).
+    -->
+    <link rel="stylesheet" href="/_css/dark.css" media="(prefers-color-scheme: dark)">
+    <link rel="stylesheet" href="/_css/light.css" media="(prefers-color-scheme: no-preference), (prefers-color-scheme: light)">
+    <!-- The main stylesheet -->
```
mathiasbynens: I think in this case we can avoid the complexity by inlining the contents of `light.css` and `dark.css` into `main.css`.

mathiasbynens: The only change to `base.jk` is then the addition of `<meta name="color-scheme" content="dark light">`. WDYT?

src/_css/dark.css
tomayac: Yes, I would argue it improves legibility significantly.

With antialiasing:

Screen Shot 2019-10-03 at 18 05 15

Without antialiasing:

Screen Shot 2019-10-03 at 18 05 28

src/_js/main.mjs
tomayac: Ooops, sure. https://github.com/v8/v8.dev/pull/254/commits/2c89589b24279015498b7706163800c40bd7098b

src/_includes/layouts/base.njk
tomayac: This makes sure we really only load what we actually need, and with HTTP/2 small files aren't a problem in this case. It also allows the toggle to just disable or enable the override without the need for an extra "dark" class or something.

mathiasbynens: > This makes sure we really only load what we actually need, and with HTTP/2 small files aren't a problem in this case.

I’m more worried about the additional content-blocking `<script>` on every page than about loading a few CSS bytes that we might not need.

> It also allows the toggle to just disable or enable the override without the need for an extra "dark" class or something.

I don’t understand this part. With what I had in mind, there is no `dark` class. We’d just add the contents of `light.css` to the top of `main.css`, and then add:

```css
@media (prefers-color-scheme: dark) {
  /* contents of dark.css go here */
}
```

Am I misunderstanding?

mathiasbynens: Another benefit is that this way, everything would Just Work™ with our service worker, without us needing to update our Workbox configuration.

tomayac:
> Is re-coloring the images generally expected of a dark mode? 

It's the ideal case, yes. In an optimal world, you'd have two versions of each image, but that's not feasible probably. The [heuristic](https://github.com/v8/v8.dev/blob/754f5df29b74f856bd0698b78e04e7ed538eb85a/src/_css/dark.css#L22-L24) tries to make sure it only captures diagrams, and the broad majority of the images I sampled are with a white background that through inversion and hue-rotating can be made look pretty great in dark mode. For the occasional image that looks odd, the opt-out via the `no-darkening` class is a quick fix (we could crowdsource this).

> I worry that some of our charts would no longer have decent foreground/background contrast (important for accessibility)

I checked the contrast and  it still passes WCAG AA for both [#527E40](https://webaim.org/resources/contrastchecker/?fcolor=527E40&bcolor=262625) and [#262625](https://webaim.org/resources/contrastchecker/?fcolor=5585D3&bcolor=262625).

> Re: the Twitter widget, I’d personally wouldn’t bother with changing it dynamically

Agreed, removed the `todo`: https://github.com/v8/v8.dev/pull/254/commits/754f5df29b74f856bd0698b78e04e7ed538eb85a

> But trying to force it into the correct setting from our end (even on initial load!) feels like overkill to me. WDYT?

I politely disagree, their light theme looks completely out of place in dark mode. We should really make sure the correct version loads at least initially.

> I’m on the fence about the toggle. The minified version is 9.2 kB

Gzipped it's [2.5kb](https://www.pika.dev/packages/dark-mode-toggle).

Screen Shot 2019-10-03 at 18 18 54

> That’s a huge increase for a small feature. OTOH, the toggle looks amazing and works incredibly well. 

My [research](https://medium.com/dev-channel/let-there-be-darkness-maybe-9facd9c3023d#9084) has shown that a significant number of people like to have their operating system UI dark, but at the same time they *don't* want the web to be dark. The toggle allows for respecting both choices, giving preference to the operating system setting.

mathiasbynens:
> the broad majority of the images I sampled are with a white background that through inversion and hue-rotating can be made look pretty great in dark mode

For the record, for diagrams we _should_ always have a white (and not transparent) background — even before this dark mode patch. So now we have even more reasons to do so! I just noticed that the diagrams in the most recent blog post still have a transparent background (fyi @RReverser) but we can definitely fix that.

> I politely disagree, their light theme looks completely out of place in dark mode. We should really make sure the correct version loads at least initially.

Okay. In that case ~~we do need to add the new `.js` file to our service worker config though~~. Scratch that, it would be simpler to just add this to `main.mjs`, conditional on `document.querySelector('.twitter-timeline')`. That way we don’t have to change the build system or the service worker config, either. WDYT? (I’m still hoping that longer-term Twitter can fix this on their end through `theme=os-default` or something similar, at which point we’d be able to remove the additional custom JS).

> Gzipped it's 2.5kb.

The minified `main.mjs` is 504 bytes gzip’d, so adding 2.5 kB would make it a 6× increase. Not quite as bad as 10× but still significant, IMHO.

src/_includes/layouts/base.njk
tomayac: It's explained in the [article](https://web.dev/prefers-color-scheme#loading-strategy), but the *tl;dr* is: [not all browsers](https://caniuse.com/#feat=prefers-color-scheme) understand the `prefers-color-scheme` media query, which leads to such browsers downloading stylesheets with unknown media query with [lowest priority](https://blog.tomayac.com/2018/11/08/why-browsers-download-stylesheets-with-non-matching-media-queries-180513/). This is bad, because either `light.css` or `dark.css` are needed, apart from `main.css`. The small [inline script](https://github.com/v8/v8.dev/blob/754f5df29b74f856bd0698b78e04e7ed538eb85a/src/_includes/layouts/base.njk#L8-L18) makes sure this doesn't happen and `light.css`(as an arbitrarily chosen default) is loaded with highest priority.

The way the toggle makes sure even browsers that don't support the media query get the theme switching experience is by changing the media to something all browsers know how to deal with:

Screen Shot 2019-10-03 at 18 49 21

You can see this in action on [this old Chrome 74 browser session](https://www.browserling.com/browse/win/7/chrome/74/https%3A%2F%2Fdark-mode-baseline.glitch.me%2F), only Chrome 76 started supporting `prefers-color-scheme`.

src/_css/img/moon.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" class="dark-mode-toggle-icon" viewBox="0 0 24 24"><path d="M20 8.69V4h-4.69L12 .69 8.69 4H4v4.69L.69 12 4 15.31V20h4.69L12 23.31 15.31 20H20v-4.69L23.31 12 20 8.69zM12 18c-.89 0-1.74-.2-2.5-.55C11.56 16.5 13 14.42 13 12s-1.44-4.5-3.5-5.45C10.26 6.2 11.11 6 12 6c3.31 0 6 2.69 6 6s-2.69 6-6 6z"/><path d="M0 0h24v24H0V0z" fill="none"/></svg>
```
mathiasbynens: Can these files be minified further without breaking anything? E.g. the `class="dark-mode-toggle-icon"` _looks_ like it could be removed (but I’m probably missing something).

RReverser:
> For the record, for diagrams we should always have a white (and not transparent) background — even before this dark mode patch. So now we have even more reasons to do so! I just noticed that the diagrams in the most recent blog post still have a transparent background (fyi @RReverser) but we can definitely fix that.

I'm confused / missing the context - why is white background better? Recoloring white background to a dark one feels more like a hack, whereas having background originally transparent allows to style it to any theme more easily.

RReverser:
As for the toggle - looking at https://unpkg.com/dark-mode-toggle@0.4.5/dist/dark-mode-toggle.min.mjs, it appears that 1) large part of it is just an embedded CSS code and 2) all that CSS is not minified (it contains whitespaces, newlines etc.).

Arguably CSS minification might not help much with gzip sizes, but perhaps still worth applying when bundling `.min.js`?

mathiasbynens:
> I'm confused / missing the context - why is white background better? 

An image with a transparent (instead of white) background and black text currently looks ok on v8.dev, but only _accidentally_ — because our main background happens to be white.

There are a few exceptions where tranparency makes sense (e.g. on `/logos`) and that where we’ll now be using `no-darkening`. But generally, for 99% of diagrams/charts/any blog post figures, the background should be white.

src/_includes/layouts/base.njk
mathiasbynens: So `<dark-mode-toggle>` *requires* separate stylesheets?

src/_css/img/moon.svg
tomayac: Good catch, https://github.com/v8/v8.dev/pull/254/commits/4ae061636329bd0184ddd64fc6b878d783dbfe57 fixes this. I have run them through [svgomg](https://jakearchibald.github.io/svgomg/), but somehow the classes remained. Seems like https://github.com/svg/svgo/issues/1068 is dealing with this.

tomayac:
> Isn't that the job of dark mode's recoloring to adjust text colors in SVG either way?

In theory [`currentColor` could be leveraged](https://web.dev/prefers-color-scheme#use-currentcolor-for-inline-svgs) for this purpose, but it only works for inlined SVGs. White background and inverting plus hue-rotating works OK enough in 90% of the cases I'd say.

src/_includes/layouts/base.njk
tomayac: It absolutely [doesn't](https://github.com/GoogleChromeLabs/dark-mode-toggle#-using-a-css-class-that-you-toggle), but I'd still argue it's the nicer architecture, especially once you start working with bigger apps. For example, *"[i]mplementing Dark Mode [in WebKit's Web Inspector] took over **1,000 lines of CSS**"* ([source](https://webkit.org/blog/8892/dark-mode-in-web-inspector/)).

tomayac:
> The minified `main.mjs` is 504 bytes gzip’d, so adding 2.5 kB would make it a 6× increase.

Props for the proper use of '×' 👍.  I would still like to call out that we're talking about `(504 bytes) + (2.5 kilobytes) = 3.00400 kilobytes` here.

src/_includes/layouts/base.njk
mathiasbynens: Great to hear! I completely agree that for more complex apps having separate stylesheets is the better approach.

However, for the extremely minimal/simple case of v8.dev, I **strongly** prefer the class-based approach as it is a much less invasive change. We don’t have to change our HTML, we don’t have to change our service worker config, and we don’t have to change our build pipeline.

I look at it this way: the current patch adds 667 bytes to each and every HTML page on v8.dev to try and prevent loading ~700 bytes of CSS (either `light.css` or `dark.css`) unnecessarily. You’ve done such an amazing job keeping both `light.css` and `dark.css` super small — let’s leverage it!

tomayac:
> Arguably CSS minification might not help much with gzip sizes, but perhaps still worth applying when bundling .min.js?

I have opened https://github.com/GoogleChromeLabs/dark-mode-toggle/issues/10 for this.

src/_includes/layouts/base.njk
tomayac: @mathiasbynens Commit https://github.com/v8/v8.dev/pull/254/commits/e82bb820e649c847670cdac171ebf4c56e84806b changes the logic to now use a class `dark` with one CSS file instead of conditional loading several CSS files. PTAL.

src/_css/main.css
```diff
@@ -27,6 +37,45 @@
...
+  --avatar-img-filter: none;
+  --icon-no-support-filter: grayscale(100%);
+  --note-color: var(--main-color);
+  --note-background-color: var(--lightyellow);
+  --tag-color: var(--black);
+  --tag-background-color: var(--lightestgray);
+  --logo-hover-background-color: var(--lightestgray);
+}
+
+:root.dark {
```
mathiasbynens: Is there a nice way to make this work for `@media (prefers-color-scheme: dark) {…}` as well? So that, if the JS fails to load for some reason, we still get the light/dark mode depending on OS preference?

tomayac: > Is there a nice way to make this work for `@media (prefers-color-scheme: dark) {…}` as well? So that, if the JS fails to load for some reason, we still get the light/dark mode depending on OS preference?

Unfortunately not, we'd have to roll back to https://github.com/v8/v8.dev/pull/254/commits/4ae061636329bd0184ddd64fc6b878d783dbfe57 with the multiple CSS files to make this completely independent from JS. With the class approach (one CSS file), the custom element listens for `prefers-color-scheme` changes and then [acts](https://github.com/v8/v8.dev/blob/e82bb820e649c847670cdac171ebf4c56e84806b/src/_js/main.mjs#L25-L29) upon them.

mathiasbynens: I guess there _could_ be a solution in theory, where `<dark-mode-toggle>` would fetch the CSS file, inspect the contents to figure out what's wrapped in `@media (prefers-color-scheme: dark) { … }`, and then do its magic that way.

But I agree, the current approach seems fine.

tomayac: > I guess there could be a solution in theory, where `<dark-mode-toggle>` would fetch the CSS file

Wait, I thought the constraint was that its *"JS fails to load for some reason"*… :-)

mathiasbynens: > > I guess there could be a solution in theory, where `<dark-mode-toggle>` would fetch the CSS file
> 
> Wait, I thought the constraint was that its *"JS fails to load for some reason"*… :-) 

Yes, that is the constraint. I'm probably not explaining my thoughts very well, apologies :P

With a pure CSS approach using `@media (prefers-color-scheme: dark) { … }`, **everything works even without JS**. This is the standard approach.

I'm saying that `<dark-mode-toggle>` could (theoretically) build on top of this, without the need for the earlier approach that included HTML + additional `<script>` changes, and without special hardcoded CSS classes (the current approach) that still rely on JS.

To be clear, I don't want to block this patch on this. I'm fine with the current approach. I'm just thinking out loud here :)

tomayac: The way to make this work with *and* without JavaScript would be as follows:

- **With JavaScript** (i.e., relying on `prefers-color-scheme` with the toggle to override it):

   Just do what we do in https://github.com/v8/v8.dev/pull/254/commits/e82bb820e649c847670cdac171ebf4c56e84806b. The reason why we use a hard-coded `dark` class (or in the previous multiple CSS files approach, override the `media` attribute values) is that we explicitly *fight* the browser; we override the system preference it communicates down to us.

- **Without JavaScript** (i.e., relying on `prefers-color-scheme` alone):

   Wrap the [CSS rules relevant for dark mode](https://github.com/v8/v8.dev/blob/e82bb820e649c847670cdac171ebf4c56e84806b/src/_css/main.css#L57-L79) as follows (we would want to automate this probably):

   ```html
   <noscript>
     <style>
       @media (prefers-color-scheme: dark) {
         /* Inline https://github.com/v8/v8.dev/blob/e82bb820e649c847670cdac171ebf4c56e84806b/src/_css/main.css#L57-L79 */
         /* This could of course also be outsourced to an external file */
       }
     </style>
   </noscript>
   ```

src/index.njk
```diff
@@ -16,6 +16,4 @@ description: 'V8 is Google’s open source high-performance JavaScript and WebAs
 <p>More articles can be found in <a href="/blog">the blog archive</a>.</p>
 
 <h2>Latest <a href="https://twitter.com/v8js" rel="me nofollow">@v8js</a> tweets</h2>
-
-<a href="https://twitter.com/v8js" rel="me nofollow" class="twitter-timeline" data-dnt="true" data-height="1000" data-chrome="noheader nofooter" data-theme="light">View tweets by @v8js</a>
-<script async src="https://platform.twitter.com/widgets.js"></script>
```
mathiasbynens: Let's keep this as-is and move the Twitter embed logic to `main.mjs`.

src/_js/main.mjs
```diff
@@ -57,3 +74,16 @@ const firstScript = document.scripts[0];
...
+  if (twitterWidgetOnPage) {
+    document.querySelector('main').insertAdjacentHTML('beforeend', prefix +
+        (toggle ? toggle.mode : 'light') + suffix);
+  }
+})(
+  document.querySelector('script[src*="platform.twitter.com"]'),
+  document.querySelector('dark-mode-toggle'),
+  '<a href="https://twitter.com/v8js" rel="me nofollow" class="twitter-timeline" data-dnt="true" data-height="1000" data-chrome="noheader nofooter" data-theme="',
+  '">View tweets by @v8js</a>'
+);
```
mathiasbynens: Can we simplify this to something like:

```js
const twitterTimeline = document.querySelector('.twitter-timeline');
if (twitterTimeline) {
  const main = document.querySelector('main');
  const toggle = document.querySelector('dark-mode-toggle');
  const prefix = '<a href="https://twitter.com/v8js" rel="me nofollow" class="twitter-timeline" data-dnt="true" data-height="1000" data-chrome="noheader nofooter" data-theme="';
  const suffix = '">View tweets by @v8js</a>';
  main.insertAdjacentHTML('beforeend', prefix + (toggle ? toggle.mode : 'light') + suffix);
}
```

It’s easier to read, and it also avoids querying the DOM for `dark-mode-toggle` unnecessarily. 

Alternatively, I’d be happy to do it as a follow-up to this patch, no worries.

tomayac: Done, see notes in commit message of https://github.com/v8/v8.dev/pull/254/commits/f8ba3e615c15fed363c0a06b614d103b01595528

mathiasbynens: Awesome! Thanks

src/_includes/layouts/base.njk
```diff
@@ -4,7 +4,9 @@
     <meta charset="utf-8">
     <title>{% if '/' === page.url %}V8 JavaScript engine{% else %}{{ (renderData.title or title) | markdown | striptags | decodeHtmlEntities }} · V8{% endif %}</title>
     <meta name="viewport" content="width=device-width,initial-scale=1">
+    <meta name="color-scheme" content="dark light">
     <link rel="stylesheet" href="/_css/main.css">
+    <noscript><style>:root{display:block}</style></noscript>
```
mathiasbynens: `<html>` gets `class=js` when JS is enabled (see `<script>document.documentElement.className+=' js'</script>` a couple of lines down). Can we leverage that within `main.css` instead of adding `<noscript><style>` here?

src/_css/main.css
```diff
@@ -13,12 +13,22 @@
 */
 
 :root {
+  color-scheme: dark light;
+  display: none;
```
mathiasbynens: If we move this declaration to `:root.js { … }`, we can get rid of the `<noscript><style>`.

tomayac: Addressed in https://github.com/v8/v8.dev/pull/254/commits/3b32524d665aa734e362db26b9040280bd3881d0.

src/_includes/layouts/base.njk
tomayac: Addressed in https://github.com/v8/v8.dev/pull/254/commits/3b32524d665aa734e362db26b9040280bd3881d0.

src/_js/main.mjs
```diff
@@ -59,6 +59,14 @@ if (location.search.includes('utm_source')) {
   history.replaceState({}, '', location.pathname);
 }
 
+// Helper function to dynamically insert scripts.
+const insertScript = (src) => {
+  const firstScript = document.scripts[0];
```
mathiasbynens: Let’s hoist this one out of the function.

tomayac: Done https://github.com/v8/v8.dev/pull/254/commits/b3dadcae605cdf26e322cbd929b20c7f428e31be

src/_js/main.mjs
```diff
@@ -53,7 +78,11 @@ ga.q = [];
...
-const scriptElement = document.createElement('script');
-scriptElement.src = 'https://www.google-analytics.com/analytics.js';
-firstScript.parentNode.insertBefore(scriptElement, firstScript);
+insertScript('https://www.google-analytics.com/analytics.js');
+
+// Dynamically either insert the dark or the light themed Twitter widget.
+const twitterTimeline = document.querySelector('.twitter-timeline');
+if (twitterTimeline) {
+  twitterTimeline.dataset.theme = darkModeToggle.mode;
+  insertScript('https://platform.twitter.com/widgets.js');
```
mathiasbynens: We’ll want to add `insertScript` + `insertScript('https://platform.twitter.com/widgets.js');` to `legacy.js`  as well.

src/_js/main.mjs
```diff
@@ -42,6 +59,14 @@ if (location.search.includes('utm_source')) {
   history.replaceState({}, '', location.pathname);
 }
 
+// Helper function to dynamically insert scripts.
+const firstScript = document.scripts[0];
+const insertScript = (src) => {
+  const scriptElement = document.createElement('script');
+  scriptElement.src = src;
+  firstScript.parentNode.insertBefore(scriptElement, firstScript);
```
RReverser: Why do we always want to insert it as the first script instead of `document.head.appendChild(...)`?

mathiasbynens: The difference doesn’t matter in practice anymore these days. I’d be fine with changing it!

mathiasbynens:
The white flash on every page load was very noticeable. I’ve removed it here: a04cdb13fd95f2210f32526b30fcc90d82c0d697 This way there’s a) a much milder flash (content appears instead of remaining hidden — it’s just in the wrong theme for a split second), and b) it occurs only when the dark mode is enabled.

tomayac:
Note that [paint holding](https://developers.google.com/web/updates/2019/05/paint-holding) avoids the flash of white as much as possible, and, once [implemented](https://bugs.chromium.org/p/chromium/issues/detail?id=939811), the `color-scheme` CSS property makes sure that *white* is at least *black* in dark mode.

(I’m offline for good now 👋)

RReverser:
Overall this looks good to me, but I'd love to figure out how to make this work just like demo page of dark-mode-toggle does. Somehow, it works well with DevTools override in that triggering override from DevTools changes state of the toggle, whereas implementation in this branch doesn't.

src/_includes/layouts/base.njk
```diff
@@ -43,7 +43,7 @@
           <a href="https://twitter.com/v8js" rel="me nofollow">Twitter</a> ·
           <a href="https://github.com/v8/v8.dev/tree/master/{{ page.inputPath }}" rel="nofollow">Edit this page on GitHub</a>
         </nav>
-        <dark-mode-toggle permanent dark="Light Theme" light="Dark Theme"></dark-mode-toggle>
+        <dark-mode-toggle dark="Light Theme" light="Dark Theme"></dark-mode-toggle>
```
RReverser: @mathiasbynens This completely removes storage of the chosen option... I don't think that's what we want, I think we just want external color scheme event to modify the stored option just like the toggle does itself.

RReverser: I think what we want is to actually change this bit in dark-mode-toggle:

https://github.com/GoogleChromeLabs/dark-mode-toggle/blob/master/src/dark-mode-toggle.mjs#L141-L144

Currently, it assumes that if user toggled the switch, then it should always persist its state, but what we want is for external event to change mode too.

RReverser: TL;DR - try to remove `if (...)` there.

RReverser: I worked around this without changing dark-mode-toggle by manually listening to external events for now, so either the toggle or external event can change the theme. @mathiasbynens let me know what you think about the behaviour.

mathiasbynens: I like it! Thanks Ingvar :+1:

RReverser:
Just tried opening v8.dev on Windows 10 and love how well it works when switching system theme :)

image
image

PR #257, opened 2019-10-07
Promise.allSettled is enabled by default in Firefox 71

chicoxyzzy:
In Firefox 68 it's only available under flag


mathiasbynens:
Thanks!

PR #264, opened 2019-10-18
Add instructions to back-merge via gerrit

backes:
Instead of using the command-line script, it is often easier to use gerrit (if the patch still applies cleanly).
This adds a section explaining how to do that.


natorion:
This method brakes tools like mergeinfo.py or Chromiumdash if the original commit hash is not included in the commit message of the merge commit.

Could you add this please?

backes:
Ah, good to know! Is it this line that is required?
(cherry picked from commit 1d6542bb65dc502b58fcde8e0d1b8108f95eb849)

backes:
Michael, can you please take another look?

PR #266, opened 2019-10-22
Dynamically re-render Twitter widget, fixes #260

### LINKED ISSUE #260: Re-render twitter widget on color scheme change
Currently, when changing color scheme via either system settings or the toggle, Twitter widget remains in its initial scheme and looks out of place.

tomayac:
Fixes https://github.com/v8/v8.dev/issues/260. Changes in this PR:
- Dynamically re-render the Twitter widget upon color scheme changes.
- Update `<dark-mode-toggle>` to v0.4.10 that now reacts on color scheme changes even with the `permanent` attribute set.


.gitignore
```diff
@@ -2,7 +2,7 @@
 src/_css/img/safari.png
 src/_css/img/chrome.svg
 src/_css/img/firefox.svg
-src/_js/dark-mode-toggle.mjs
+src/_js/dark-mode-toggle.min.mjs
```
mathiasbynens: ?

tomayac: This is actually no longer required. Removing and resolving the conversation…

src/_js/main.mjs
```diff
@@ -13,25 +13,55 @@
...
+// Dynamically either insert the dark- or the light-themed Twitter widget.
+let twitterTimelineContainer = document.querySelector('.twitter-widget');
+const updateTwitterTimeline = async () => {
+  if (twitterTimelineContainer) {
+    await twitterLoaded;
+    const newContainer = twitterTimelineContainer.cloneNode();
+    twitterTimelineContainer.insertAdjacentElement('afterend', newContainer);
+    await twttr.widgets.createTimeline({
+      screenName: 'v8js',
+      sourceType: "profile"
```
mathiasbynens:
```suggestion
      sourceType: 'profile',
```

src/_js/main.mjs
```diff
@@ -13,25 +13,55 @@
...
+    },
+    newContainer,
+    {
+      dnt: true,
+      height: 1000,
+      chrome: 'noheader nofooter',
+      theme: darkModeToggle.mode
```
mathiasbynens:
```suggestion
      theme: darkModeToggle.mode,
```

src/_js/main.mjs
```diff
@@ -13,25 +13,55 @@
...
 const updateThemeClass = () => {
   root.classList.toggle('dark', darkModeToggle.mode === 'dark');
 };
 
+// Only load the Twitter script when we need it.
+const twitterLink = document.querySelector('.twitter-link');
+let twitterLoaded = null;
+if (twitterLink) {
+  twitterLoaded = import('https://platform.twitter.com/widgets.js')
+  .then(() => twitterLink.remove());
```
mathiasbynens:
```suggestion
    .then(() => twitterLink.remove());
```

package.json
```diff
@@ -32,13 +32,16 @@
     "@11ty/eleventy": "^0.8.3",
     "@11ty/eleventy-plugin-rss": "^1.0.6",
     "@11ty/eleventy-plugin-syntaxhighlight": "^2.0.3",
+    "@babel/cli": "^7.6.4",
+    "@babel/plugin-syntax-dynamic-import": "^7.2.0",
     "@browser-logos/chrome": "^1.0.12",
     "@browser-logos/firefox": "^2.0.5",
     "@browser-logos/safari": "^1.0.4",
     "autoprefixer": "^9.6.1",
     "babel-minify": "^0.5.1",
```
RReverser: We probably don't need this as direct dependency anymore, as we don't use their CLI.

src/_js/main.mjs
```diff
@@ -13,25 +13,55 @@
...
 // Listen for toggle changes (which includes `prefers-color-scheme` changes)
 // and toggle the `dark` class accordingly.
-darkModeToggle.addEventListener('colorschemechange', updateThemeClass);
-
-// Force listening for external events (by default "permanent" prevents this).
-matchMedia('(prefers-color-scheme:dark)').addListener(({matches}) => {
-  darkModeToggle.mode = matches ? 'dark' : 'light';
+darkModeToggle.addEventListener('colorschemechange', () => {
   updateThemeClass();
+  updateTwitterTimeline();
```
RReverser: These are always bundled, let's just call `updateTwitterTimeline` from `updateThemeClass`.

src/_js/main.mjs
```diff
@@ -91,4 +106,6 @@ ga.q = [];
 ga('create', UA_ID, 'auto');
 ga('set', 'referrer', document.referrer.split('?')[0]);
 ga('send', 'pageview');
-insertScript('https://www.google-analytics.com/analytics.js');
+const script = document.createElement('script');
```
RReverser: Why was this changed back / it seems to work fine when trying to `import`?

src/_js/main.mjs
```diff
@@ -13,25 +13,55 @@
...
+  twitterLoaded = import('https://platform.twitter.com/widgets.js')
+    .then(() => twitterLink.remove());
+}
+
+// Dynamically either insert the dark- or the light-themed Twitter widget.
+let twitterTimelineContainer = document.querySelector('.twitter-widget');
+const updateTwitterTimeline = async () => {
+  if (twitterTimelineContainer) {
+    await twitterLoaded;
+    const newContainer = twitterTimelineContainer.cloneNode();
```
RReverser: This container should probably be hidden initially, otherwise I guess UI would jump around?

RReverser: Oh, right, it still doesn't resolve if it's hidden because iframe onload is not firing...

package.json
tomayac: Removing…

src/_js/main.mjs
tomayac: Done…

src/_js/main.mjs
tomayac: Oh, right, they don't send CORS headers for localhost, but they do for anything else. Changing back to dynamic import then.

src/_js/main.mjs
tomayac: Good point, fixed.

tomayac: It works when the container is not showing, so we're good I guess.

package.json
RReverser: Marking this back as unresolved - it's specifically `babel-minify` line that I commented on that shouldn't be necessary anymore :)

mathiasbynens:
Generally, I defer to @RReverser on this one.

What's the file size delta for (minified) `/_js/main.mjs`? (Side note: we should probably get some automated tooling to list file size deltas on PRs that touch the JS or CSS assets.)

tomayac:
Unfortunately Twitter makes this clumsier than it should be, but here's hoping that the widget eventually—like their PWA already—becomes `prefers-color-scheme` aware…

mathiasbynens:
_enables Protector of the Realm mode_

I don't want to block this patch, but I do feel like we're taking the increase in JS a little bit too lightly, and justifying it too easily. This is some non-critical functionality for a single page on v8.dev (the homepage). It doesn't affect any other page — other than the 1.85× JS bundle size and the costs that come with it. This is a large increase, and I would personally not have bothered given that the change in functionality is relatively minimal. In other words, I don't think this _pays for itself_.

I like @RReverser's suggestion to move the Twitter-specific JS to a separate script that's only used on the home page. The Google Analytics code could remain in the main bundle, since it's used on all pages. I'll leave it up to y'all to decide whether this is follow-up work or part of this PR.

Thanks for your work on this!

RReverser:
@mathiasbynens If it's any consolation, both before and after the change JS still fits into just 1 TCP packet, so at least it shouldn't require an extra network round-trip :P

tomayac:
Why don’t we open a meta issue and set up a [size budget](https://web.dev/performance-budgets-101#add-performance-budgets-to-your-build-process) of the size of a TCP package and split the file as soon as we fail this budget?

RReverser:
@mathiasbynens If we're speaking of file sizes, do you know if we can legally shorten the license preamble? For example, I'm not sure if we need to include a paragraph that is already contained in the license file we reference.

It's not particularly compressible, and currently (post this PR) accounts for 26% of the gzipped `main.mjs` - that seems like a lot and would only be worse by duplicating it in a separate file.

tomayac:
The [short form](http://www.apache.org/foundation/license-faq.html#Apply-My-Software) is OK for the minified script.

RReverser:
If it's indeed OK, that would be great to change!

mathiasbynens:
Prior to this patch, we didn’t have any license header in the minified output, only in the source code. File-size-wise that seems optimal.

RReverser:
@mathiasbynens Oh, we didn't? I thought it's kinda required to preserve license comments in output?

RReverser:
Without comments, the new version is just 735 bytes gzipped, so I guess comparison I showed above is incorrect and it's only 700 bytes -> 735 bytes.

src/_js/main.mjs
```diff
@@ -91,4 +104,4 @@ ga.q = [];
 ga('create', UA_ID, 'auto');
 ga('set', 'referrer', document.referrer.split('?')[0]);
 ga('send', 'pageview');
-insertScript('https://www.google-analytics.com/analytics.js');
+import('https://www.google-analytics.com/analytics.js');
```
RReverser: @tomayac @mathiasbynens Looks like no-cors error still appears on v8.dev - apparently because `import` is using CORS requests whereas regular scripts don't (TIL: https://v8.dev/features/dynamic-import).

It's a shame as `import` is kind of nice for situations like this, but I guess I'll have to change it back for Google Analytics...

PR #270, opened 2019-10-26
Re-remove comments from output

RReverser:
They were accidentally enabled in #266, so disabling them again.

This cuts `main.mjs.gz` size down from 1294 bytes to 780 bytes (before the initial PR: 700 bytes).

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,5 +4,6 @@
       "topLevel": true
     }
   }]],
-  "plugins": ["@babel/plugin-syntax-dynamic-import"]
+  "plugins": ["@babel/plugin-syntax-dynamic-import"],
+  "comments": false
 }


jridgewell:
https://github.com/v8/v8.dev/pull/270/files#diff-e56633f72ecc521128b3db6586074d2cR2

`preset-minify` is really experimental, you'd probably have better results with `loose` mode and terser.

RReverser:
It works well enough for us, and pretty similarly to just `babel-minify` CLI we used before, but I do have another branch with terser we might switch to later.

RReverser:
@jridgewell Pushed my Terser branch in https://github.com/v8/v8.dev/pull/271. It does seem to work quite a bit better!

PR #272, opened 2019-10-26
Fix Google Analytics

RReverser:
`import` works only with CORS headers, which are not sent by GA script.

Ref: https://github.com/v8/v8.dev/pull/266/files#r339317674


RReverser:
I'll merge this fix without waiting.

PR #277, opened 2019-11-06
Explain that string.matchAll(nonGlobalRegExp) throws

mathiasbynens:
Per https://github.com/tc39/ecma262/pull/1716.

@joshualitt @schuay PTAL


joshualitt:
LGTM

Issue #274, opened 2019-10-27
Low or no contrast text colours


declanbyrd:
While in dark mode, on the dynamic import documentation page some of the links are either difficult to read because the contrast is low, or in this case there are words with no contrast to the background.

![image](https://user-images.githubusercontent.com/24387679/67635346-d0f90980-f8bd-11e9-9318-f5ec4f3b5d0a.png)

Light theme for comparison:
![image](https://user-images.githubusercontent.com/24387679/67635403-99d72800-f8be-11e9-9f66-1643d6502998.png)

In main.css the colour of the links and note background are the same for dark mode:

```CSS
:root.dark {
  --main-and-footer-link-color: var(--yellow);
  --note-background: var(--yellow);
}
```

I'd be happy to create a PR to fix this with a bit of guidance to align with the style guide.

mathiasbynens:
Thanks for spotting and reporting this, @declanbyrd! A PR would be amazing.

Re: guidance, the general rule is to follow the style of the surrounding code (which in the case of CSS isn’t that much). Please let us know if there’s any specific questions you have. cc @tomayac @RReverser

PR #278, opened 2019-11-06
Fix contrast for links in a note in dark mode

tomayac:
Fixes https://github.com/v8/v8.dev/issues/274

Screen Shot 2019-11-06 at 09 21 34

### DIFF
--- a/src/_css/main.css
+++ b/src/_css/main.css
@@ -297,7 +297,7 @@ main a:hover code, main a:focus code {
   color: var(--darkblue);
   border-bottom: 1px solid var(--darkblue);
 }
-.note a:hover {
+.note a:hover, .note a:focus {
   color: var(--white);
   background-color: var(--darkblue);
 }
--- a/src/_css/main.css
+++ b/src/_css/main.css
@@ -293,6 +293,14 @@ main a:hover, main a:focus, footer a:hover, footer a:focus {
 main a:hover code, main a:focus code {
   background-color: inherit;
 }
+.note a {
+  color: var(--darkblue);
+  border-bottom: 1px solid var(--darkblue);
+}
+.note a:hover {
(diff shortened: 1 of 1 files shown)


src/_css/main.css
```diff
@@ -293,6 +293,14 @@ main a:hover, main a:focus, footer a:hover, footer a:focus {
 main a:hover code, main a:focus code {
   background-color: inherit;
 }
+.note a {
+  color: var(--darkblue);
+  border-bottom: 1px solid var(--darkblue);
```
tomayac: @mathiasbynens By the way, any reason you use `border` instead of [`text-decoration`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration) and [`text-decoration-skip-ink`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip-ink)?

tomayac:
CC: @declanbyrd, the initial reporter. Thanks!

src/_css/main.css
```diff
@@ -293,6 +293,14 @@ main a:hover, main a:focus, footer a:hover, footer a:focus {
...
+}
+.note a:hover {
```
mathiasbynens:
```suggestion
.note a:hover, .note a:focus {
```

src/_css/main.css
mathiasbynens: No

mathiasbynens: Either way, let’s discuss that potential change as a separate patch, as it involves rules all over the stylesheet.

tomayac: @RReverser Switching bottom borders for proper text decoration might be a good follow-up PR.

tomayac: https://github.com/v8/v8.dev/issues/279

PR #287, opened 2019-11-14
Nullish Coalescing was added to Firefox 72

chicoxyzzy:


mathiasbynens:
Will this be in Firefox 72 stable, or is it 72 Nightly-only for now? +@codehag

codehag:
It is in 72 nightly, which will be the release branch on January 7th, 2020

mathiasbynens:
@codehag What I meant is, is this fully shipped (i.e. it will become available once Firefox 72 hits stable), or shipped in Nightly only?

For example, `String.prototype.replaceAll` is in Firefox 72, but is (for now) [only exposed in Nightly](https://bugzilla.mozilla.org/show_bug.cgi?id=1540021). That is, it won’t automatically become available in Firefox 72 once it hits stable.

Is nullish coalescing similarly gated on an additional flag, or will it simply be shipped ship once 72 hits stable?

codehag:
It’s not flagged, so it will become available when firefox 72 becomes
stable.

mathiasbynens:
Thanks, everyone!

PR #294, opened 2019-11-28
Bump installed version of `<dark-mode-toggle>`

tomayac:
The new release fixes a bug with non-hover devices that led to the toggle leaving too much bottom margin.

### DIFF
--- a/src/_css/main.css
+++ b/src/_css/main.css
@@ -425,8 +425,8 @@ img {
 }
 
 dark-mode-toggle {
-  --dark-mode-toggle-light-icon: url(/_css/img/sun.svg);
-  --dark-mode-toggle-dark-icon: url(/_css/img/moon.svg);
+  --dark-mode-toggle-light-icon: url(/_css/img/moon.svg);
+  --dark-mode-toggle-dark-icon: url(/_css/img/sun.svg);
   --dark-mode-toggle-color: var(--main-and-footer-link-color);
   --dark-mode-toggle-icon-filter: invert(80%);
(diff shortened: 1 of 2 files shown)


src/_css/main.css
```diff
@@ -425,8 +425,8 @@ img {
 }
 
 dark-mode-toggle {
-  --dark-mode-toggle-light-icon: url(/_css/img/sun.svg);
-  --dark-mode-toggle-dark-icon: url(/_css/img/moon.svg);
+  --dark-mode-toggle-light-icon: url(/_css/img/moon.svg);
+  --dark-mode-toggle-dark-icon: url(/_css/img/sun.svg);
```
mathiasbynens: lolwhoops

tomayac: Right? 

But in the end, it's this old discussion if a button should show the current state, or the state after pressing the button.

chicoxyzzy: Previous version (before this change) looks more intuitive to me 🙊

tomayac: If you look at the [demo](https://googlechromelabs.github.io/dark-mode-toggle/demo/index.html), the naming makes sense for the `switch`, but not so much for the `toggle` appearance. Would renaming to `--dark-mode-toggle-on-icon` and `--dark-mode-toggle-off-icon` make it any better?

chicoxyzzy: Oh, it seems that I was looking at old version. What I see now at [v8.dev](https://v8.dev) LGTM 🙂.

PR #305, opened 2019-12-18
Add a SIMD Feature explainer

dtig:


ripsawridge:
Wow, this is fantastic, will read today!

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+The WebAssembly SIMD proposal defines a portable, performant subset of SIMD operations that are available across most modern architectures. This proposal derived many elements from the [SIMD.js proposal](https://github.com/tc39/ecmascript_simd), which in turn was originally derived from the [Dart SIMD](https://www.researchgate.net/publication/261959129_A_SIMD_programming_model_for_dart_javascriptand_other_dynamically_typed_scripting_languages) specification. The SIMD.js proposal was an API proposed at TC39 with new types and functions for performing SIMD computations, but this was archived in favor of supporting SIMD operations more transparently in WebAssembly. The [WebAssembly SIMD proposal](https://github.com/WebAssembly/simd) was introduced as away for browsers to take advantage of the data level parallelism using the underlying hardware.
```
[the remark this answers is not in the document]
backes: note: "away" -> "a way"

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+## WebAssembly SIMD Proposal
+
+The high level goal of the WebAssembly SIMD proposal is to introduce vector operations to the WebAssembly Specification, in a way that guarantees portable performance. 
+
+The set of SIMD instructions is large, and varied across architectures. The set of operations included in the WebAssembly SIMD proposal consist of operations that are well supported on a wide variety of platforms, and are proven to be performant. To this end, the current proposal is limited to standardizing Fixed-Width 128-bit SIMD operations.
```
backes:
```suggestion
The set of SIMD instructions is large, and varied across architectures. The set of operations included in the WebAssembly SIMD proposal consists of operations that are well supported on a wide variety of platforms, and are proven to be performant. To this end, the current proposal is limited to standardizing Fixed-Width 128-bit SIMD operations. 
```

backes: "consist" -> "consists"

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+The high level goal of the WebAssembly SIMD proposal is to introduce vector operations to the WebAssembly Specification, in a way that guarantees portable performance. 
+
+The set of SIMD instructions is large, and varied across architectures. The set of operations included in the WebAssembly SIMD proposal consist of operations that are well supported on a wide variety of platforms, and are proven to be performant. To this end, the current proposal is limited to standardizing Fixed-Width 128-bit SIMD operations. 
+
+The current proposal introduces a new v128 value type, and a number of new operations that operate on this type. The criteria used to determine these operations are -
+
+* The operations should be well supported across multiple modern architectures.
+
+* Performance wins should be positive across multiple relevant arches within an instruction group.
```
backes:
```suggestion
* Performance wins should be positive across multiple relevant architectures within an instruction group.
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+* The chosen set of operations should minimize performance cliffs if any.
+
+ 
+
+The proposal is in active development, both V8, and the toolchain have working prototype implementations for experimentation. As these are prototype implementations, they are subject to change as new operations are added to the proposal.
```
backes:
```suggestion
The proposal is in active development, both V8 and the toolchain have working prototype implementations for experimentation. As these are prototype implementations, they are subject to change as new operations are added to the proposal.
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Using WebAssembly SIMD
+
+### Enabling experimental SIMD support in Chrome
+
+WebAssembly SIMD support is prototyped behind a flag in chrome, to try out the SIMD support on the browser, pass --enable-features=WebAssemblySimd, or toggle the WebAssemblySimd flag in chrome://flags. This work is bleeding edge, and continuously being worked on. To minimize the chances of breakage, please use the latest version of the toolchain as detailed below, and a recent Chrome Canary. If something doesn’t look right, please file a bug at crbug.com/v8.
```
backes: "crbug.com/v8" should be a link (or does that happen automatically?)

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Building C / C++ to target SIMD
+
+WebAssembly’s SIMD support depends on using a recent build of clang with the WebAssembly LLVM backend enabled. Emscripten has support for the WebAssembly SIMD proposal as well. Install, and activate the latest-upstream distribution of emscripten using [emsdk](https://emscripten.org/docs/getting_started/downloads.html) to use the bleeding edge SIMD features.
```
backes:
```suggestion
WebAssembly’s SIMD support depends on using a recent build of clang with the WebAssembly LLVM backend enabled. Emscripten has support for the WebAssembly SIMD proposal as well. Install and activate the latest-upstream distribution of emscripten using [emsdk](https://emscripten.org/docs/getting_started/downloads.html) to use the bleeding edge SIMD features.
```

backes: The whole text seems to use ’ instead of ' (some UTF-8 apostrophe instead of ascii). Not sure if this is intended.

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+        v128_t prod = wasm_i32x4_mul(a, b);
+
+        wasm_v128_store(&out[i], prod);
+
+    }
+
+}
+```
+
+This manually rewritten code assumes that the input and output arrays are aligned and do not alias and that size is a multiple of four. The autovectorizer cannot make these assumptions and has to generate extra code to handle the cases where they are not true, so hand-written SIMD code often ends up being smaller than autovectorized SIMD code.
```
backes: Out of curiosity: What happens if you use "__builtin_assume" or "__builtin_assume_aligned"?
If this works, would it be worth mentioning?

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Compelling Use Cases
+
+The WebAssembly SIMD proposal seeks to accelerate high compute applications like audio/video codecs, image processing applications, cryptographic applications, etc. Currently WebAssembly SIMD is experimentally supported in widely used open source projects like [Halide]([https://github.com/halide/Halide/blob/master/README_webassembly.md](https://github.com/halide/Halide/blob/master/README_webassembly.md)), [OpenCV.js]([https://docs.opencv.org/3.4/d5/d10/tutorial_js_root.html](https://docs.opencv.org/3.4/d5/d10/tutorial_js_root.html)), and [XNNPACK]([https://github.com/google/XNNPACK](https://github.com/google/XNNPACK)). [TODO: add links]
```
backes: These links don't look right: [name] ([url](url)).
Also, there is still the "TODO: add links".

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+One of the most visually appealing demos where it’s easy to observe the difference in performance SIMD makes, is a following hand-tracking system. Without SIMD, you can get only around 3 frames per second on a modern laptop, while with SIMD enabled you get a much smoother experience at 15-16 frames per second.
+
+![image alt text](/src/_img/simd/image_2.gif) 
+
+[https://pursuit.page.link/MediaPipeHandTrackingSimd](https://pursuit.page.link/MediaPipeHandTrackingSimd)
+
+Visit the link above in Chrome Canary with SIMD enabled to try it!
+
+Another interesting set of demos that makes use of SIMD for smooth experience, come from OpenCV - a popular computer vision library that can also be compiled to WebAssembly. They’re available by link bit.ly/opencv-camera-demos, or you can check out the pre-recorded versions below:
```
backes: The link should be linkified.

RReverser:
Looks like the GIFs are not animated - is that intentional?

If we do want animations, please re-add them and convert to MP4 - it tends to be much smaller and more efficient (either using any online tool or FFMpeg, whichever is more convenient, or leave it up to me).

If we don't, then please convert them to JPEG e.g. using https://squoosh.app/ - that should make them 4-5 times smaller.

src/features/simd.md
RReverser: Probably not without an explicit `https://`, need to check.

src/features/simd.md
RReverser: Yeah AFAIK it works. If we do mention it, it might be worth mentioning `#pragma clang loop vectorize(enable)` and friends as a way to assert that certain loop is autovectorised. I wouldn't go into too many details, just mentioning it and linking to the corresponding LLVM page (e.g. https://llvm.org/docs/Vectorizers.html) should be enough.

src/features/simd.md
dtig: I think consist may still be the right word here because it's referring to the set of operations and not the proposal.

dtig:
> Looks like the GIFs are not animated - is that intentional?

Nope, not intentional - it looked like they were working when I uploaded to _img/ will try again. 
> 
> If we do want animations, please re-add them and convert to MP4 - it tends to be much smaller and more efficient (either using any online tool or FFMpeg, whichever is more convenient, or leave it up to me).
> 
> If we don't, then please convert them to JPEG e.g. using https://squoosh.app/ - that should make them 4-5 times smaller.

src/features/simd.md
backes: Ack, but the set *consists* of stuff, because it's a single set, no?

src/features/simd.md
mathiasbynens: In general, we don't do auto-linking on v8.dev. Do one of these:

- To turn `https://example.com/` into a link with custom link text, use `[foo bar baz](https://example.com/)`
- To turn `https://example.com/` into a link, wrap it in `<>`: `<https://example.com/>`

In 99% of cases you want the former.

src/features/simd.md
mathiasbynens: That's awesome, and what we do for all of v8.dev (at least everything that went through me) :)

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+        wasm_v128_store(&out[i], prod);
+
+    }
+
+}
+```
+
+This manually rewritten code assumes that the input and output arrays are aligned and do not alias and that size is a multiple of four. The autovectorizer cannot make these assumptions and has to generate extra code to handle the cases where they are not true, so hand-written SIMD code often ends up being smaller than autovectorized SIMD code.
+
+## Compelling Use Cases
```
mathiasbynens: Nit: use sentence case

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+Applications that have already been ported to use WebAssembly may benefit from SIMD with no source modifications thanks to LLVM’s autovectorization optimizations.
+
+These optimizations can automatically transform loops that perform arithmetic operations on each iteration into equivalent loops that perform the same arithmetic operations on multiple inputs at a time using SIMD instructions. LLVM’s autovectorizers are enabled by default at optimization levels -O2 and -O3 when the -msimd128 flag is supplied.
+
+For example, consider the following function that multiplies the elements of two input arrays together and stores the results in an output array.
+
+```
+void multiply_arrays(int* out, int* in_a, int* in_b, int size) {
```
mathiasbynens: these double newlines in all code examples seem accidental

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
+---
+title: 'Enable High Performance Workloads using WebAssembly SIMD, and V8'
```
mathiasbynens:
```suggestion
title: 'Enable high-performance workloads using WebAssembly SIMD and V8'
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+author: 'Deepti Gandluri ([@dptig](https://twitter.com/dptig)), Thomas Lively ([@tlively52](https://twitter.com/tlively52))'
+date: 2019-17-12
```
mathiasbynens: `yyyy-dd-mm`

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+![image alt text](/src/_img/simd/image_3.gif)
+
+Card Reading
+
+![image alt text](/src/_img/simd/image_4.gif)
+
+Invisibility cloak
+
+![image alt text](/src/_img/simd/image_5.gif)
```
mathiasbynens: Nit: use `-` instead of underscores in file names

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Enabling experimental SIMD support in Chrome
+
+WebAssembly SIMD support is prototyped behind a flag in chrome, to try out the SIMD support on the browser, pass --enable-features=WebAssemblySimd, or toggle the WebAssemblySimd flag in chrome://flags. This work is bleeding edge, and continuously being worked on. To minimize the chances of breakage, please use the latest version of the toolchain as detailed below, and a recent Chrome Canary. If something doesn’t look right, please file a bug at crbug.com/v8.
+
+### Building C / C++ to target SIMD
+
+WebAssembly’s SIMD support depends on using a recent build of clang with the WebAssembly LLVM backend enabled. Emscripten has support for the WebAssembly SIMD proposal as well. Install and activate the latest-upstream distribution of emscripten using [emsdk](https://emscripten.org/docs/getting_started/downloads.html) to use the bleeding edge SIMD features.
+
+```
```
chicoxyzzy:
```suggestion
```sh
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+./emsdk install latest-upstream
+
+./emsdk activate latest-upstream
+```
+
+There are a couple of different ways to enable generating SIMD code when porting your application to use SIMD. Once the latest upstream emscripten version has been installed, compile using emscripten, and pass the -msimd128 flag to enable SIMD.
+
+```
```
chicoxyzzy:
```suggestion
```sh
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+emcc -msimd128 -O3 foo.c -o foo.js
+```
+
+Applications that have already been ported to use WebAssembly may benefit from SIMD with no source modifications thanks to LLVM’s autovectorization optimizations.
+
+These optimizations can automatically transform loops that perform arithmetic operations on each iteration into equivalent loops that perform the same arithmetic operations on multiple inputs at a time using SIMD instructions. LLVM’s autovectorizers are enabled by default at optimization levels -O2 and -O3 when the -msimd128 flag is supplied.
+
+For example, consider the following function that multiplies the elements of two input arrays together and stores the results in an output array.
+
+```
```
chicoxyzzy:
```suggestion
```cpp
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+        out[i] = in_a[i] * in_b[i];
+
+    }
+
+}
+```
+
+Without passing the -msimd128 flag, the compiler emits this WebAssembly loop:
+
+```
```
chicoxyzzy:
```suggestion
```wasm
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+            (i32.load ... get address in `in_b` ...)
+
+    ...
+
+)
+```
+
+But when the -msimd128 flag is used, the autovectorizer will turn this into code that includes this loop:
+
+```
```
chicoxyzzy:
```suggestion
```wasm
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+)
+```
+
+As you can see, the loop body has the same structure but SIMD instructions are being used to load, multiply, and store four elements at a time inside the loop body.
+
+For finer grained control over the SIMD instructions generated by the compiler, include [this header file](https://github.com/emscripten-core/emscripten/blob/incoming/system/include/wasm_simd128.h), which defines a set of intrinsics. Intrinsics are special functions that, when called, will be turned by the compiler into the corresponding WebAssembly SIMD instructions, unless it can make further optimizations.
+
+As an example, here is the same function from before manually rewritten to use the SIMD intrinsics.
+
+```
```
chicoxyzzy:
```suggestion
```cpp
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+## WebAssembly SIMD Proposal
+
+The high level goal of the WebAssembly SIMD proposal is to introduce vector operations to the WebAssembly Specification, in a way that guarantees portable performance. 
+
+The set of SIMD instructions is large, and varied across architectures. The set of operations included in the WebAssembly SIMD proposal consist of operations that are well supported on a wide variety of platforms, and are proven to be performant. To this end, the current proposal is limited to standardizing Fixed-Width 128-bit SIMD operations. 
+
+The current proposal introduces a new v128 value type, and a number of new operations that operate on this type. The criteria used to determine these operations are -
```
chicoxyzzy:
```suggestion
The current proposal introduces a new v128 value type, and a number of new operations that operate on this type. The criteria used to determine these operations are:
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+ 
+
+The proposal is in active development, both V8 and the toolchain have working prototype implementations for experimentation. As these are prototype implementations, they are subject to change as new operations are added to the proposal.
+
+## Using WebAssembly SIMD
+
+### Enabling experimental SIMD support in Chrome
+
+WebAssembly SIMD support is prototyped behind a flag in chrome, to try out the SIMD support on the browser, pass --enable-features=WebAssemblySimd, or toggle the WebAssemblySimd flag in chrome://flags. This work is bleeding edge, and continuously being worked on. To minimize the chances of breakage, please use the latest version of the toolchain as detailed below, and a recent Chrome Canary. If something doesn’t look right, please file a bug at crbug.com/v8.
```
chicoxyzzy:
```suggestion
WebAssembly SIMD support is prototyped behind a flag in chrome, to try out the SIMD support on the browser, pass `--enable-features=WebAssemblySimd`, or toggle the WebAssemblySimd flag in `chrome://flags`. This work is bleeding edge, and continuously being worked on. To minimize the chances of breakage, please use the latest version of the toolchain as detailed below, and a recent Chrome Canary. If something doesn’t look right, please [file a bug](https://crbug.com/v8).
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+WebAssembly’s SIMD support depends on using a recent build of clang with the WebAssembly LLVM backend enabled. Emscripten has support for the WebAssembly SIMD proposal as well. Install and activate the latest-upstream distribution of emscripten using [emsdk](https://emscripten.org/docs/getting_started/downloads.html) to use the bleeding edge SIMD features.
+
+```
+./emsdk install latest-upstream
+
+./emsdk activate latest-upstream
+```
+
+There are a couple of different ways to enable generating SIMD code when porting your application to use SIMD. Once the latest upstream emscripten version has been installed, compile using emscripten, and pass the -msimd128 flag to enable SIMD.
```
chicoxyzzy:
```suggestion
There are a couple of different ways to enable generating SIMD code when porting your application to use SIMD. Once the latest upstream emscripten version has been installed, compile using emscripten, and pass the `-msimd128` flag to enable SIMD.
```

src/features/simd.md
```diff
@@ -0,0 +1,179 @@
...
+
+```
+emcc -msimd128 -O3 foo.c -o foo.js
+```
+
+Applications that have already been ported to use WebAssembly may benefit from SIMD with no source modifications thanks to LLVM’s autovectorization optimizations.
+
+These optimizations can automatically transform loops that perform arithmetic operations on each iteration into equivalent loops that perform the same arithmetic operations on multiple inputs at a time using SIMD instructions. LLVM’s autovectorizers are enabled by default at optimization levels -O2 and -O3 when the -msimd128 flag is supplied.
```
chicoxyzzy:
```suggestion
These optimizations can automatically transform loops that perform arithmetic operations on each iteration into equivalent loops that perform the same arithmetic operations on multiple inputs at a time using SIMD instructions. LLVM’s autovectorizers are enabled by default at optimization levels -O2 and -O3 when the `-msimd128` flag is supplied.
```

src/features/simd.md
RReverser: +1, consists seems correct here.

src/features/simd.md
mathiasbynens: Can we get these images in higher quality? GIF is probably not the best format

RReverser: Please check above where I commented on this. @dtig will share originals with me and I'll convert to MP4.

src/features/simd.md
dtig: This should be fixed now.

src/features/simd.md
dtig: I've added both .gif, and .mp4 formats to this PR, I've chosen to animate the gifs for now because getting the videos to render exactly how I wanted them with loop play was challenging. Suggestions on how to do this correctly are appreciated. If the gifs turn out to be sufficient, I'll delete the .mp4 versions of these files.

RReverser: `<video autoplay loop muted playsinline>` should work - that's how e.g. Giphy, Imgur, Twiter and others render "GIFs" (also MP4s in fact).

src/features/simd.md
RReverser: These actually don't work:
```
Language does not exist: sh
Problem writing Eleventy templates: (more in DEBUG output)
> Having trouble rendering liquid (and markdown) template ./src/features/simd.md (TemplateContentRenderError)
> Cannot read property 'rest' of undefined (TypeError):
    TypeError: Cannot read property 'rest' of undefined
```

It should be `bash` instead.

RReverser: Fixed.

chicoxyzzy: Thanks Ingvar! It's funny because I made it `bash` locally. I have no idea why I proposed `sh` here  ¯\_(ツ)_/¯ 😅

src/features/simd.md
RReverser: @dtig Sounds about right. Let's accept the suggestion then?

src/features/simd.md
```diff
@@ -0,0 +1,151 @@
+---
+title: ' WebAssembly SIMD: Enable Fast, Parallel applications'
```
RReverser: Nit: we mostly seem to use normal sentence case for titles in features section. Also there is an extra leading space in title.

How about this?

```suggestion
title: 'Fast, parallel applications with WebAssembly SIMD'
```

PR #306, opened 2019-12-18
Handle pre in .note width changes

tomayac:
This PR properly changes the width of code blocks in notes sections.

### DIFF
--- a/src/_css/main.css
+++ b/src/_css/main.css
@@ -301,6 +301,9 @@ main a:hover code, main a:focus code {
   color: var(--white);
   background-color: var(--darkblue);
 }
+.note pre[class^="language"] {
+  width: calc(100% + 1.5 * var(--offset));
+}
 a.bookmark {
   display: none;
 }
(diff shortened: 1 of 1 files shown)


src/_css/main.css
```diff
@@ -301,6 +301,9 @@ main a:hover code, main a:focus code {
   color: var(--white);
   background-color: var(--darkblue);
 }
+.note pre[class^="language"] {
```
mathiasbynens: Do we need the attribute selector? Wouldn't `.note pre` work?

tomayac: It's just a safety net in case someone includes a non-syntax-highlighted `pre` block. It could probably be removed.

mathiasbynens: That's what I meant: we'd want those non-syntax highlighted `pre` blocks to also render correctly, right?

tomayac: Let me fix it.

PR #308, opened 2019-12-18
Make page zoomable

### LINKED ISSUE #307: Page can't be zoomed 😕

tomayac:
By setting absolute font sizes the page can't be zoomed [cmd] + [+]/[-], which is really bad for #a11y. This PR fixes this issue.

Fixes #307

### DIFF
--- a/src/_css/main.css
+++ b/src/_css/main.css
@@ -85,7 +85,7 @@ html, body, fieldset, nav ul, nav ul li, h1, h2, h3 {
   padding: 0;
 }
 html {
-  font: 16px/1.6 sans-serif;
+  font: 100%/1.6 sans-serif;
   color: var(--html-color);
   background: var(--html-background);
 }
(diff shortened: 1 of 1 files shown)


mathiasbynens:
Zooming using `Cmd` + `+`/`-` works fine for me on v8.dev. I use it all the time! Can you clarify what this is fixing exactly?

tomayac:
Sorry, this description was misleading. This is about users who have their default browser level font size set differently so they don't need to zoom in all the time. By setting an absolute unit, you override their choice. By setting a relative unit (100%), you respect it. And by making everything a function of the base size things look universally great. 🎉

Screen Shot 2019-12-18 at 11 37 24

mathiasbynens:
Awesome, thanks for educating me!

src/_css/main.css
```diff
@@ -113,7 +113,7 @@ body {
 }
 #header h1 a {
   padding: 1em;
-  font-size: 16px;
+  font-size: 1em;
```
mathiasbynens: Turns out this should've been `rem`. Whoops! https://github.com/v8/v8.dev/commit/a25843d09a6b0f32a74171d5edd8dc0199a91bf1

PR #310, opened 2019-12-18
Conditionally include PWACompat

### LINKED ISSUE #275: Why I can't add PWA to my phone in Safari 13
I can add the V8 to my home page in the Mac, but I can't get a PWA page in my iPhone.

You can see the screenshot in the below, the icon is not **the v8 logo**:
IMG_1520

I am not sure if it is caused by compatibility issues.

tomayac:
On Apple mobile devices add the proprietary app icon and splashscreen markup. No one should have to do this manually, and eventually this annoyance will go away once https://bugs.webkit.org/show_bug.cgi?id=183937 is fixed.

Fixes #275

### DIFF
--- a/src/_js/main.mjs
+++ b/src/_js/main.mjs
@@ -111,6 +111,7 @@ document.head.appendChild(gaScript);
 // On Apple mobile devices add the proprietary app icon and splashscreen markup.
 // No one should have to do this manually, and eventually this annoyance will
 // go away once https://bugs.webkit.org/show_bug.cgi?id=183937 is fixed.
-if (/\b(iPad|iPhone|iPod)\b/.test(navigator.userAgent)) {
+if ((/\b(iPad|iPhone|iPod)\b/.test(navigator.userAgent)) &&
+    (!matchMedia('(display-mode: standalone)').matches)) {
   import('https://unpkg.com/pwacompat');
 }
--- a/src/_js/main.mjs
+++ b/src/_js/main.mjs
@@ -111,7 +111,6 @@ document.head.appendChild(gaScript);
 // On Apple mobile devices add the proprietary app icon and splashscreen markup.
 // No one should have to do this manually, and eventually this annoyance will
 // go away once https://bugs.webkit.org/show_bug.cgi?id=183937 is fixed.
-if ((/\b(iPad|iPhone|iPod)\b/.test(navigator.userAgent)) &&
-    (!matchMedia('(display-mode: standalone)').matches)) {
+if (/\b(iPad|iPhone|iPod)\b/.test(navigator.userAgent)) {
   import('https://unpkg.com/pwacompat');
 }
--- a/src/_js/main.mjs
+++ b/src/_js/main.mjs
@@ -107,3 +107,10 @@ ga('send', 'pageview');
 const gaScript = document.createElement('script');
 gaScript.src = 'https://www.google-analytics.com/analytics.js';
 document.head.appendChild(gaScript);
+
+// On Apple mobile devices add the proprietary app icon and splashscreen markup.
+// No one should have to do this manually, and eventually this annoyance will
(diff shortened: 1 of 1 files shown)


mathiasbynens:
I'd rather not land workarounds for non-critical engine-specific bugs.

This is an annoying lack of functionality in WebKit that should be fixed upstream, IMHO. Until then v8.dev works fine without.

tomayac:
I respectfully disagree. It looks *super ugly* on an iOS/iPadOS home screen and someone went the extra mile to open an issue about it. You (the V8 team) should weigh in on the WebKit bug and ask for it to be fixed. We all agree it's a *temporary* solution and not our problem, but in this case users should not have to suffer from it.

tomayac:
CC: @samthor.

samthor:
I don't actually have a strong opinion here. PWACompat does its job but it might not be right here.

tomayac:
Users first.

RReverser:
@tomayac Is this still an issue on Apple? If it is, I'd agree that UX is most important and given that this adds overhead only for a specific User-Agent, it should be fine to include.

tomayac:
> @tomayac Is this still an issue on Apple devices? If it is, I'd agree that UX is most important and given that this adds overhead only for a specific User-Agent, it should be fine to include.

This is still an issue. I have just made the conditional loading even lazier, so that it doesn't load on installed experiences.

RReverser:
Looking at what it does, do I understand right that we're interested in just `<link rel="icon" ...>` addition with a different size? If so, maybe it's easier to add the required size to the static markup?

tomayac:
The library dynamically (but lazily) creates splash screens(!) and icons for all devices.

RReverser:
Right, but my question is whether we need all of that to address the specific issue?

tomayac:
> Right, but my question is whether we need all of that to address the specific issue?

We give people the same experience they get on Android when they install a PWA. Needed? Probably no. Nice to have? Probably yes.

RReverser:
Right, let me rephrase: are these other features also not working in iOS natively without PWACompat?

tomayac:
> Right, let me rephrase: are these other features also not working in iOS natively without PWACompat?

This is the [list of things](https://github.com/GoogleChromeLabs/pwacompat/#details) the library does. The splash screens are a nice touch. Right now `minimal-ui` isn't supported on iOS, so it runs in a regular browser window (which doesn't trigger splash screens to show). 

image

RReverser:
Sounds reasonable; one more question: it seems like the meta tag for the splash screen should *always* be present in the app to work, no?

If so, we probably should revert the last change so that PWACompat is included even for installed experiences.

tomayac:
> Sounds reasonable; one more question: it seems like the meta tag for the splash screen should _always_ be present in the app to work, no?

The splash screens will be dynamically created based on the Web App Manifest values. No meta tag needed.

> If so, we probably should revert the last change so that PWACompat is included even for installed experiences.

When `navigator.standalone` evaluates to `true`, the library does nothing. Since right now this is `false` since we run in a browser tab, the library will be included. If we ever decide to run `standalone` (which iOS supports), we don't have to think of changing the code.

RReverser:
So you're saying that this latest added condition is not actually ever triggered? I'd suggest to just remove it for now then. If / when we decide to run in standalone, we can revisit the code (and see if Apple fixed their side by then).

tomayac:
> So you're saying that this latest added condition is not actually ever triggered? I'd suggest to just remove it for now then. If / when we decide to run in standalone, we can revisit the code (and see if Apple fixed their side by then).

[Done](https://github.com/v8/v8.dev/pull/310/files#diff-ea2880996ad513ea45772bdff95dcb24).

mathiasbynens:
> @tomayac […] given that this adds overhead only for a specific User-Agent, it should be fine to include.

Unfortunately, it doesn’t just add overhead for Safari. Check the DevTools → Network tab in Chrome, for example.

Issue #313, opened 2019-12-19
“release” tag not clickable due to tweets overlay


mathiasbynens:
https://v8.dev/blog/v8-release-80

![](https://user-images.githubusercontent.com/81942/71155815-dda33980-223e-11ea-9eb4-cd509c07cdd9.png)

Issue #312, opened 2019-12-19
Tweets SVG positioning is still off in Firefox


mathiasbynens:
https://v8.dev/blog/v8-release-80 on Firefox 71:

![](https://user-images.githubusercontent.com/81942/71154925-f579be00-223c-11ea-9f92-d8546f1c6354.png)

Perhaps CSS `@supports` is the answer here?

mathiasbynens:
cc @LeszekSwirski @RReverser

PR #314, opened 2019-12-19
Really actually fix those tweets on firefox

LeszekSwirski:
It wouldn't be CSS layout work without deeply nested `<div>` tags


mathiasbynens:
Does this fix both #312 and #313?

LeszekSwirski:
You know it!

PR #315, opened 2019-12-19
Torque enums

nico-hartmann:
Adds a small chapter about the new enum language feature to the Torque manual

### DIFF
--- a/src/docs/torque.md
+++ b/src/docs/torque.md
@@ -360,12 +360,14 @@ syntactical structure:
 </code></pre>
 
 A basic example looks like this:
+
 ```torque
 extern enum LanguageMode extends Smi {
   kStrict,
   kSloppy
 }
 ```
+
 This declaration defines a new type `LanguageMode`, where the `extends` clause specifies the underlying
 type, that is the runtime type used to represent a value of the enum. In this example, this is `TNode<Smi>`,
 since this is what the type `Smi` `generates`. A `constexpr LanguageMode` converts to `LanguageMode`
@@ -393,10 +395,12 @@ FromConstexpr<LanguageMode, LanguageMode::constexpr kStrict>(o: LanguageMode::co
     return %RawDownCast<LanguageMode>(%FromConstexpr<Smi>(o));
 }
 ```
+
 Note that `constexpr` is part of the type name and thus goes after the namespace scope.
 
 Torque's enumerations work very well together with the `typeswitch` construct, because the
 values are defined using distinct types:
+
 ```torque
 typeswitch(language_mode) {
   case (LangugageMode::kStrict): {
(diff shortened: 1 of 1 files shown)


src/docs/torque.md
```diff
@@ -348,6 +349,81 @@ Call(f, Undefined);
...
+```
+
+Note that `constexpr` is part of the type name and thus goes after the namespace scope.
+
+Torque's enumerations work very well together with the `typeswitch` construct, because the
+values are defined using distinct types:
+
+```torque
+typeswitch(language_mode) {
+  case (LangugageMode::kStrict): {
```
ripsawridge: s/Langugage/Language

src/docs/torque.md
```diff
@@ -348,6 +349,81 @@ Call(f, Undefined);
...
+    // ...
+  }
+  case (LanguageMode::kSloppy): {
```
ripsawridge: Here too

src/docs/torque.md
```diff
@@ -348,6 +349,81 @@ Call(f, Undefined);
...
+type, that is the runtime type used to represent a value of the enum. In this example, this is `TNode<Smi>`,
+since this is what the type `Smi` `generates`. A `constexpr LanguageMode` converts to `LanguageMode`
+in the generated CSA files since no `constexpr` clause is specified on the enum to replace the default name.
+If the `extends` clause is omitted, Torque will generate only the `constexpr` version of the type. The `extern` keyword tells Torque that there is a C++ definition of this enum. Currently, only `extern` enums are supported.
+
+Torque generates a distinct type and constant for each of the enum's entries. Those are defined
+inside a namespace that matches the enum's name. Necessary specializations of `FromConstexpr<>` are
+generated to convert from the entry's `constexpr` types to the enum type. The value generated for an entry in the C++ files is `<enum-constexpr>::<entry-name>` where `<enum-constexpr>` is the `constexpr` name generated for the enum. In the above example, those are `LanguageMode::kStrict` and `LanguageMode::kSloppy`.
+
+To summarize, the above example is desugared into something that looks very similar to this:
```
ripsawridge: I'd consider leaving out the desugared code example. What about just mentioned where the output for enums is written to in gen/torque-generated.

src/docs/torque.md
nico-hartmann: Done.

src/docs/torque.md
nico-hartmann: Done.

src/docs/torque.md
nico-hartmann: There is no distinct file where stuff is written for enums. They appear all over the place where they are used, mostly.

PR #316, opened 2020-01-03
Mention the status of `await` inside of modules

jkrems:
Tried to also cram in a mention of "more differences when compared to CommonJS" but couldn't find a good way to fit that into the flow.

/cc @mathiasbynens

### DIFF
--- a/src/features/modules.md
+++ b/src/features/modules.md
@@ -82,6 +82,8 @@ Modules are a little different from classic scripts:
 
 - The new static `import` and `export` syntax is only available within modules — it doesn’t work in classic scripts.
 
+- `await` cannot be used as a variable name anywhere in a module - variables in classic scripts can be named `await` outside of async functions.
+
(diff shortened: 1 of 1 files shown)


src/features/modules.md
```diff
@@ -82,6 +82,8 @@ Modules are a little different from classic scripts:
 
 - The new static `import` and `export` syntax is only available within modules — it doesn’t work in classic scripts.
 
+- `await` cannot be used as a variable name anywhere in a module - variables in classic scripts can be named `await` outside of async functions.
```
mathiasbynens:
```suggestion
- Top-level `await` is available in modules, but not in classic scripts. Relatedly, `await` cannot be used as a variable name anywhere in a module, although variables in classic scripts _can_ be named `await` outside of async functions.
```

mathiasbynens: How about this?

jkrems: Good to have that context in there!

mathiasbynens:
Thanks!

PR #318, opened 2020-01-07
Fix link and usage of V8_DEPRECATED

backes:
The codesearch links were not working any more, thus they are updated. The code example is wrong since https://crrev.com/c/1847361.


mathiasbynens:
Note that anyone in @v8/v8-core can LGTM, not just rreverser@. Just a tip :)

backes:
> Note that anyone in @v8/v8-core can LGTM, not just rreverser@. Just a tip :)

Ah, thanks, I was not sure about that. Thanks for merging!

PR #320, opened 2020-01-16
Add a polyfill example for top-level await

surma:
cc @mathiasbynens

### DIFF
--- a/src/features/top-level-await.md
+++ b/src/features/top-level-await.md
@@ -71,6 +71,18 @@ try {
 }
 ```
 
+### Polyfills
+
+You can effectively hold the execution of a static import by `await`-ing a conditional, dynamic import. This is very useful when using polyfills that you only want to load when the browser needs them:
+
+```js
+if (browserNeedsPolyfill()) {
+  await import("./polyfill.mjs");
+}
+
+import "./main.js";
+```
+
 ## Module execution order
 
 One of the biggest changes to JavaScript with top-level `await` is the order of execution of modules in your graph. The JavaScript engine executes modules in [post-order traversal](https://en.wikibooks.org/wiki/A-level_Computing/AQA/Paper_1/Fundamentals_of_algorithms/Tree_traversal#Post-order): starting from the left-most subtree of your module graph, modules are evaluated, their bindings are exported, and their siblings are executed, followed by their parents. This algorithm runs recursively until it executes the root of your module graph.


surma:
@MylesBorins PTAL! You barely ever see import-after-code in examples, so I thought it was worth adding.

@mathiasbynens said it works in v8, but both of us are actually not 100% sure that it’s spec compliant :D

src/features/top-level-await.md
```diff
@@ -71,6 +71,18 @@ try {
...
+### Polyfills
+
+You can effectively hold the execution of a static import by `await`-ing a conditional, dynamic import. This is very useful when using polyfills that you only want to load when the browser needs them:
+
+```js
+if (browserNeedsPolyfill()) {
+  await import("./polyfill.mjs");
+}
+
+import "./main.js";
```
mathiasbynens: Maybe include an example of multiple polyfills as well (to show that you don't want to import them sequentially but rather with `await Promise.all([...])`).

mathiasbynens: Oh! You could use `Array#reduce` to build the array of needed polyfills! `/me runs`

jridgewell: Does this actually work? My thought was that sibling modules could execute before the async import?

MylesBorins: Siblings can indeed execute before if you have an await. Also if this is in the "Root Node" of an application it won't even execute until the entire graph is done.

Further... it would appear that according to the [HTML integration](https://github.com/whatwg/html/pull/4352) that one script tag would not block a second script tag (follow similar sibling behavior)... so honestly I'm unsure that `await` is a reliable / good mechanism for polyfills

/cc @littledan and @domenic who might have other opinions here.

surma: That definitely makes this pattern unviable then, which I was unsure about and was the whole reason I looped you in in the first place :D Ignore this PR then.

jridgewell: One more:

Let's call this module `entrypoint.js`, and it has this TLA and a later `import "main.js"` (or a specifier import/default import, etc). Can `main.js` be executed before the TLA resolves?

Definitely `entrypoint.js`'s siblings could execute while the TLA is resolving.

littledan: Huh, I'm pretty confused by this discussion. I think it's good to show people how they can conditionally load polyfills, so that the practice of unconditionally including all the polyfills can be reduced. I always thought that this was one of the main motivating use cases for top-level await.

Don't you want to encourage people to include the polyfills at the application integration point, rather than deeper in a module graph? If so, I think this pattern works.

I'm confused the discussion of multiple script tags. This example doesn't use multiple script tags, but `<script type=module>` is already not a load-blocking script, and yet continues to happen in order per HTML semantics. This is just like if your script tag had a fetch and .then to eval it, to load a polyfill--no better or worse.

MylesBorins:
```suggestion
import('./main.mjs')
```

MylesBorins: This example works fine if you change the static import here to be a dynamic one. This would avoid the import being hoisted.

My original assessment was incorrect based on looking at this wrong. If we have no children and this is the entry point for your application this pattern works great.

We could also show if there are multiple polyfills of creating an array of polyfill promises and `await Promise.all` if we are not worried about them being in flight at the same time

jridgewell: > Don't you want to encourage people to include the polyfills at the application integration point, rather than deeper in a module graph? If so, I think this pattern works.

As a practice, yes. But I don't think TLA actually works this way. What this code wants is for `main.js` to load _after_ `polyfill.js`. But I don't think we prevent `main.js` from executing while we wait for `polyfill.js` to load.

But maybe I'm wrong, so that's why I asked. Maybe we can depend on `polyfill.js` executing before `main.js`.

> I'm confused the discussion of multiple script tags.

I think that came up because it's hard to distinguish between which sibling we're talking about. Definitely siblings of the module that includes a TLA can execute first. But can modules requested (eg, `main.js`) by the module that includes a TLA (this code example, or `entrypoint.js`) execute before the TLA (`polyfill.js`)?

MylesBorins:
In the original example I believe the static import would be hoisted.
Making it a dynamic import would have the desired behavior (assuming there
are no static imports in the file at all)

jridgewell:
> In the original example I believe the static import would be hoisted.
Making it a dynamic import would have the desired behavior (assuming there
are no static imports in the file at all)

That's my thought as well. I kinda wish this pattern would work, but I don't know how we would spec it.

MylesBorins:
I am not sure we should. Arguably the import of main is dynamic since it
depends on the result of polyfill

src/features/top-level-await.md
littledan: > But I don't think we prevent main.js from executing while we wait for polyfill.js to load.

Oh, that's right, it would have to be a dynamic import of "main.js", not a static import. Static imports are all hoisted. This doesn't relate to sibling execution.

littledan:
Agree with @MylesBorins change. We decided a long time ago that static imports are hoisted and dynamic imports are not. This is just core to JS semantics, and not a decision that was made as part of the TLA proposal.

surma:
If both imports are dynamic, I don’t think the sample is showing off anything new. The exact code would have worked pre-TLA, except for the requirement of a AIIFE.

As a side note: The reason I was hoping the static import would work is because it would have interesting implications for bundlers: It would let you avoid another network round-trip when the polyfill is _not_ needed. But you can still achieve that by wrapping the top-level code in `main.mjs` in an exported function.

```js
import {init} from "./main.mjs";
if (browserNeedsPolyfill()) {
  await import("./polyfill.mjs");
}
init();
```

All in all, after learning how TLA actually behaves, I feel like my sample doesn‘t provide actual value. Up to y’all, tho.

littledan:
Hmm, I don't really understand how this changes the implications for bundlers. Couldn't they still use that technique you're talking about?

The fundamental thing going on here (which isn't related to TLA) is that the top-level evaluation of modules is "hoisted", rather than happening inline with the import statement (like in Python).

surma:
Bundlers currently use the approach that static imports will be “inlined”, while dynamic imports will be left in a separate file.

So if we had the following `index.mjs` with 2 dynamic imports, we will end up with 3 files in total:

```js
if (browserNeedsPolyfill()) {
  await import("./polyfill.mjs");
}
import("./main.mjs");
```

The browser would:

- get `index.mjs`
- evaluate whether it needs to get `polyfill.mjs`
- get `main.mjs`

In total up to 3 round trips. 

My _hope_ was (as I was unaware of the hoisting) that my version with one static import would work. In that scenario the browser would:

- get `index.mjs`
- evaluate whether it needs to get ` polyfill.mjs`
- execute the contents of `main.js`, which have been bundled

So up to 2 round trips, benefitting environments that don’t need the polyfill.

But as I said, the version with the `init` function gets us to that scenario.

jridgewell:
So this might be an important one to merge to instruct devs that **this does not work**.

> Bundlers currently use the approach that static imports will be “inlined”, while dynamic imports will be left in a separate file.

Maybe something a custom module attribute might solve?

littledan:
If we want to teach devs that static import statements are hoisted and executed before any code in the module runs, that would be valuable, in case they come from languages where this isn't the case. I don't think this is unique to top-level await, and so it doesn't need to be documented together with it.

jridgewell:
> I don't think this is unique to top-level await, and so it doesn't need to be documented together with it

I think it'll be common for devs to reason that imports are run before module code. But now we have an await that delays certain modules, and it tripped up 3 TC39 delegates. So something about TLA and the order in which we write imports (before or after the TLA) causes confusion. I think that explicit documentation with TLA, and we can additionally add documentation that static imports are always hoisted in the generic ESM docs.

>  It's hard for me to see how a module attribute would change this basic behavior.

Eg, if a bundler sees `import('main.mjs, { include: 'always'})`, it knows it should include `main.mjs` in this module bundle and not split it out. Now we can do `await import('main.mjs', { include: 'always ' })`, it gets transformed by the bundler into whatever promise resolution it uses, and we avoid the [roundtrips](https://github.com/v8/v8.dev/pull/320#issuecomment-576228292).

RReverser:
@surma So... should we close this then?

surma:
Lol didn’t realize that my own PR was still open. Yes, I think it’s fair to close this one.

Issue #131, opened 2019-01-16
Consider adding ToC for larger documentation entries


szuend:
I was just browsing the Torque user manual and thought a table of contents or something similar would be nice for larger articles. Especially for articles like the torque user manual, where readers might only be interested in one specific section.

mathiasbynens:
I started looking into this, and then hit https://github.com/Oktavilla/markdown-it-table-of-contents/issues/38. Luckily, we can work around it, but it’s feels dirty.

WIP branch: https://github.com/v8/v8.dev/compare/toc?expand=1 (We still need to tweak the output HTML and add some styling.)

camillobruni:
Declaring bug bankruptcy after 4 years of inactivity.

PR #321, opened 2020-01-24
Add a blog post: Understanding the ECMAScript spec, part 1

marjakh:


src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+> 2. **Assert**: IsPropertyKey(P) is true.
+> 3. Let desc be **?** O.**\[\[GetOwnProperty\]\]**(P).
+> 4. If desc is undefined, return false.
+> 5. Return true.
+
+[Spec: HasOwnProperty](https://tc39.es/ecma262#sec-hasownproperty)
+
+But what's an "abstract operation"? What are the things inside \[\[ \]\]? Why is there a "?" in front of a function? What do the asserts mean?
+
+Let's find out!
```
gahaas: You could consider inserting a table of content here, what aspects of the spec you are going to describe here.

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+  - ECMAScript
+description: 'Tutorial on reading the ECMAScript specification'
+tweet: ''
+---
+
+... where we take a simple function in the spec and try to understand the notation.
+
+## Preface
+
+Even if you know JavaScript, reading its language specification, the ECMAScript spec, can be pretty daunting. At least that's how I felt when I started reading it for the first time. Reading the spec from the start doesn't help either, as the spec is not designed to be read in that order, and notations are not explained before they occur.
```
mathiasbynens: > notations are not explained before they occur

They usually are, AFAICT? However, the spec is so large that you often start reading a particular section without having read the earlier chapters explaining [the notational conventions](https://tc39.es/ecma262/#sec-notational-conventions).

src/blog/understanding-ecma-part-1.md
mathiasbynens: We don't support ToCs yet (see #131) but we could of course manually insert a list.

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+and
+
+> HasOwnProperty(O, P)
+>
+>The **abstract operation** HasOwnProperty is used to determine whether an object has an own property with the specified property key. A Boolean value is returned. The operation is called with arguments O and P where O is the object and P is the property key. This abstract operation performs the following steps:
+>
+> 1. **Assert**: Type(O) is Object.
+> 2. **Assert**: IsPropertyKey(P) is true.
+> 3. Let desc be **?** O.**\[\[GetOwnProperty\]\]**(P).
```
mathiasbynens: Let's wrap these things in code spans, which both looks nicer and lets us avoid the weird escaping syntax:

```
> 1. **Assert**: `Type(O)` is `Object`.
> 2. **Assert**: `IsPropertyKey(P)` is `true`.
> 3. Let desc be `? O.[[GetOwnProperty]](P)`.
```

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+>
+> Return ! OrdinaryGetOwnProperty(O, P).
+
+[Spec: \[\[GetOwnProperty\]\]](https://tc39.es/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots-getownproperty-p)
+
+(We'll find out what the exclamation mark means in the next chapter.)
+
+_OrdinaryGetOwnProperty_ is not an internal method, since it's not associated with any object; instead, the object it operates on is passed as a parameter.
+
+_OrdinaryGetOwnProperty_ is called "ordinary" since it operates on ordinary objects. EcmaScript objects can be either **ordinary** or **exotic**. Ordinary objects must have the default behavior for a set of methods called **essential internal methods**. If an object deviates from the default behavior, it's exotic.
```
mathiasbynens:
```suggestion
`OrdinaryGetOwnProperty` is called "ordinary" since it operates on ordinary objects. ECMAScript objects can be either _ordinary_ or _exotic_. Ordinary objects must have the default behavior for a set of methods called _essential internal methods_. If an object deviates from the default behavior, it's exotic.
```

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+What about the question marks and exclamation marks? To understand them, we need to look into **Completion Records**!
+
+A Completion Record is a "record" (or a "struct", if you're more familiar with that term) defined by the spec. A record is a data type which has a fixed set of named fields. C-like languages have structs - JavaScript doesn't, since you can always add properties into an object.
+
+Again, the Completion Record is defined only for spec purposes. A JavaScript engine doesn't have to have a corresponding internal data type.
+
+A Completion Record has three fields:
+
+| Name | Description |
```
mathiasbynens: Let's use `:::table-wrapper` here, similar to other tables on v8.dev.

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+... where we take a simple function in the spec and try to understand the notation.
+
+## Preface
+
+Even if you know JavaScript, reading its language specification, the ECMAScript spec, can be pretty daunting. At least that's how I felt when I started reading it for the first time. Reading the spec from the start doesn't help either, as the spec is not designed to be read in that order, and notations are not explained before they occur.
+
+Let's start with a concrete example and walk through the spec to understand it. The following code demonstrates usage of Object.prototype.hasOwnProperty:
+
+```javascript
+let o = {'foo': 1};
```
mathiasbynens:
```suggestion
const o = {'foo': 1};
```

(per Google style guide)

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+Which is the same as "Return Foo()"; it behaves the same way for both abrupt and normal completions.
+
+## Asserts
+
+Asserts in the spec assert invariant conditions of the algorithms. They are added for clarity, but don't add any requirements to the implementation - the implementation doesn't need to check them.
+
+## Moving on
+
+Now we have built the understanding needed for reading the spec for simple methods and abstract operations like Object.prototype.hasOwnProperty and HasOwnProperty. They still delegate to other abstract operations, but based on this blog post you should be able to figure out what they do. You'll need to find out about Property Descriptors, which is just another ECMAScript specification type.
```
mathiasbynens:
```suggestion
Now we have built the understanding needed for reading the spec for simple methods like `Object.prototype.hasOwnProperty` and abstract operations like `HasOwnProperty`. They still delegate to other abstract operations, but based on this blog post you should be able to figure out what they do. You'll need to find out about Property Descriptors, which is just another ECMAScript specification type.
```

mathiasbynens:
This is looking great so far! Just did a pass and left some minor comments.

src/blog/understanding-ecma-part-1.md
marjakh: Removed this sentence because it doesn't add much value anyway :)

src/blog/understanding-ecma-part-1.md
marjakh: Undecided about this one...

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+Internal slots and methods are used in the algorithms described by the spec. Internal slots are used for storing the state of the object, and internal methods are functions associated with the object.
+
+[Spec: Object internal methods and internal slots](https://tc39.es/ecma262/#sec-object-internal-methods-and-internal-slots)
+
+For example, every JavaScript object has an internal slot \[\[Prototype\]\] and an internal method \[\[GetOwnProperty\]\].
+
+Internal slots and methods are not accessible from JavaScript, for example, you cannot access _o.\[\[Prototype\]\]_ or call _o.\[\[GetOwnProperty\]\]()_. A JavaScript engine can implement them for their own internal use, but doesn't have to.
+
+Each internal method delegates to a similarly-named abstract operation:
```
syg: This is sometimes true for ordinary objects, but many internal methods are specified with algorithmic steps like abstract operations. Perhaps "Sometimes internal methods delegate to similarly-named abstract operations, such as in the case of ordinary objects' [[GetOwnProperty]]:"?

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+The most well-known exotic object is the _Array_, since its length property behaves in a non-default way: setting the length property can remove elements from the _Array_.
+
+Essential internal methods are the methods listed [here](https://tc39.es/ecma262/#table-5).
+
+## Completion Records
+
+What about the question marks and exclamation marks? To understand them, we need to look into **Completion Records**!
+
+A Completion Record is a "record" (or a "struct", if you're more familiar with that term) defined by the spec. A record is a data type which has a fixed set of named fields. C-like languages have structs - JavaScript doesn't, since you can always add properties into an object.
```
syg: I think the "struct" comparison is more confusing than not. The spec certainly doesn't care about the memory layout of a Record's fields, for instance (or even has a notion of memory layout). I think it would be fine to simply define what ecma262 considers Records to be in this paragraph.

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+| Name | Description |
+--- | ---
+| \[\[Type\]\] | One of: normal, break, continue, return, or throw. All other types except "normal" are referred to as "abrupt completions".|
+| \[\[Value\]\] | the value that was produced when the completion occurred, e.g., the return value of a function or the exception (if one is thrown)|
+| \[\[Target\]\] | used for directed control transfers (ignoring it for this blog post)|
+
+[Spec: Completion Record](https://tc39.es/ecma262/#sec-completion-record-specification-type)
+
+Every abstract operation implicitly returns a Completion Record. Even if it looks like an abstract operation would return a simple type such as Boolean, it's implicitly wrapped into a Completion Record with the type "normal" (see [Implicit Completion Values](https://www.ecma-international.org/ecma-262/index.html#sec-implicit-completion-values)).
```
syg: We're trying to change this! The implicitness doesn't work in all contexts and is generally confusing. Maybe a note that the editors are currently exploring making all completion record handling more explicit.

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+> let obj be Foo()
+> // obj is a Completion Record
+> ReturnIfAbrupt(obj)
+> // If we're still here, obj is now the value extracted from the Completion Record
+> Bar(obj)
+
+And now the question mark comes into play: ? Foo() is equivalent to _ReturnIfAbrupt(Foo())_.
+
+Similarly, Let val be ! Foo() is equivalent to:
```
syg: Here and elsewhere: the specese snippets like "Let val be ! Foo()" be styled differently.

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+Internal slots are (assumed) data members (of a JavaScript object or a spec-internal data type) and internal methods are (assumed) member functions (of a JavaScript object or a spec-internal data type).
+
+Internal slots and methods are used in the algorithms described by the spec. Internal slots are used for storing the state of the object, and internal methods are functions associated with the object.
+
+[Spec: Object internal methods and internal slots](https://tc39.es/ecma262/#sec-object-internal-methods-and-internal-slots)
+
+For example, every JavaScript object has an internal slot \[\[Prototype\]\] and an internal method \[\[GetOwnProperty\]\].
+
+Internal slots and methods are not accessible from JavaScript, for example, you cannot access _o.\[\[Prototype\]\]_ or call _o.\[\[GetOwnProperty\]\]()_. A JavaScript engine can implement them for their own internal use, but doesn't have to.
```
syg:
```suggestion
Internal slots and methods are not accessible from JavaScript. For example, you cannot access _o.\[\[Prototype\]\]_ or call _o.\[\[GetOwnProperty\]\]()_. A JavaScript engine can implement them for their own internal use, but doesn't have to.
```

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+## ECMAScript Language Types and ECMAScript Specification Types
+
+Let's start with something that looks familiar. The spec uses values such as undefined, true and false, which we already know from JavaScript. They all are **ECMAScript language values**, values of **ECMAScript language types** which the spec also defines.
+
+The spec also uses ECMAScript language values internally, for example, an internal data type might contain a field whose possible values are true and false.
+
+[Spec: ECMAScript language types](https://tc39.es/ecma262/#sec-ecmascript-language-types)
+
+**ECMAScript specification types** are types that occur only in the ECMAScript spec, but not in the JavaScript language, and which might or might not occur as internal types in a JavaScript engine. In this blog post, we'll get to know the ECMAScript specification type _Record_ (and its subtype _Completion Record_).
```
syg:
```suggestion
In addition to language types and values, the specification also uses types and values occur only in the ECMAScript spec. These are called **ECMAScript specification types**, and a JavaScript engine does not need to (but is free to) implement them. In this blog post, we'll get to know the ECMAScript specification type _Record_ (and its subtype _Completion Record_).
```

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+> O.\[\[GetOwnProperty\]\]
+>
+> 1. Let temp be OrdinaryGetOwnProperty(O, P)
+> 2. Assert: temp is not an abrupt completion
+> 3. Let temp be temp.\[\[Value\]\]
+> 4. Return NormalCompletion(temp)
+
+Here we assume that _temp_ is a brand new temporary variable which doesn't collide with anything else.
+
+We've also used the knowledge that when a return statement returns something else than a Completion Record, it's implicitly wrapped inside a NormalCompletion.
```
syg: Sorry to harp on this point. I'd add a note here that even though the implicit rule is in place, the spec itself isn't always consistent. Often predicate-like AOs are called and their bare return values are used without unwrapping via `?` and `!`. A note to the reader that this is usually clear from context, though it is being worked on upstream.

src/blog/understanding-ecma-part-1.md
marjakh: I'm trying to maintain a style where quotes from the spec are inside >. Did you mean the code block should also be inside > (is that even possible?) or that the > should end before the code starts (confusing!)?

marjakh: Ah I read your comment again, I see what you mean. So not wrapping the whole thing inside a code block but individually. I might argue thought that the whole lines should be inside the code span, like. `Let desc be ...`

marjakh: Experimented a bit.

Option 1:

> When the hasOwnProperty method is called with argument V, the following steps are taken:
>
> `1. Let P be ? ToPropertyKey(V).`
> `2. Let O be ? ToObject(this value).`
> `3. Return ? HasOwnProperty(O, P).`

-> I think this looks ugly

Option 2:

> When the hasOwnProperty method is called with argument V, the following steps are taken:
>
> 1. Let `P` be ? `ToPropertyKey(V)`.
> 2. Let `O` be ? `ToObject(this value)`.
> 3. Return ? `HasOwnProperty(O, P)`.

-> I think this is confusing, also it's hard to decide what should be inside the code spans and what shouldn't. Let? Return? Question marks?

Option 3:

> When the hasOwnProperty method is called with argument V, the following steps are taken:
>
> 1. Let P be **?** ToPropertyKey(V).
> 2. Let O be **?** ToObject(this value).
> 3. Return **?** HasOwnProperty(O, P).

This is what I have now, the pros are that it allows using emphasis inside these blocks and it matches what the spec does. I still like this version best...

src/blog/understanding-ecma-part-1.md
marjakh: Did this suggestion *except* I kept the terms *bold* because they stand out better that way. (I'd like to be able to skim the blog post, look for a certain term and its explanation.)

src/blog/understanding-ecma-part-1.md
marjakh: Haha, I think I painted myself in a corner here :) I started changing "spec code snippets" inside the text to be inside `code spans`, and I think it's a good idea. But now they look different inside the text and inside the spec quotes. Argh.

mathiasbynens: I consistently use option 2, with the difference that `?` is wrapped in the code span as well. It’s not an ordinary part of an English sentence, so it should be marked accordingly. IMHO this is the only semantically correct way of representing this content in Markdown (keeping in mind what HTML it produces).

I agree that option 1 is overkill (and ugly). Option 3 is… not great, semantically speaking.

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+**Abstract operations** are functions defined in the ECMAScript spec; they are defined for the purpose of writing the spec elegantly and concisely. A JavaScript engine doesn't have to implement them as separate functions inside the engine. They cannot be directly called from JavaScript.
+
+[Spec: Abstract Operations](https://tc39.es/ecma262/#sec-abstract-operations)
+
+## Internal slots and internal methods
+
+**Internal slots** and **internal methods** use names enclosed in \[\[ \]\].
+
+Internal slots are (assumed) data members (of a JavaScript object or a spec-internal data type) and internal methods are (assumed) member functions (of a JavaScript object or a spec-internal data type).
```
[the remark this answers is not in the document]
marjakh: That they're there only for spec purposes, e.g., a JavaScript object as it's represented in the JS engine might not have them. But it's confusing, so, removed (assumed).

src/blog/understanding-ecma-part-1.md
marjakh: I'm convinced, going with option 2.

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+```javascript
+let o = {'foo': 1};
+o.hasOwnProperty('foo'); // true
+o.hasOwnProperty('bar'); // false
+```
+
+In the example, _o_ doesn't have a property called _hasOwnProperty_, so we walk up the prototype chain and look for it. We find it in _o_'s prototype, which is _Object.prototype_.
+
+When trying to find out how _Object.prototype.hasOwnProperty_ is defined, I bumped into pseudocode-like descriptions like this:
```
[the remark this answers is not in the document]
marjakh: I unified this a bit; kept the "I" only in the first paragraph where I tell how daunted **I** was; for the rest of the blog post, I now use "we".

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,232 @@
...
+
+[Spec: ECMAScript language types](https://tc39.es/ecma262/#sec-ecmascript-language-types)
+
+**ECMAScript specification types** are types that occur only in the ECMAScript spec, but not in the JavaScript language, and which might or might not occur as internal types in a JavaScript engine. In this blog post, we'll get to know the ECMAScript specification type _Record_ (and its subtype _Completion Record_).
+
+[Spec: ECMAScript specification types](https://tc39.es/ecma262/#sec-ecmascript-specification-types)
+
+## Abstract operations
+
+**Abstract operations** are functions defined in the ECMAScript spec; they are defined for the purpose of writing the spec elegantly and concisely. A JavaScript engine doesn't have to implement them as separate functions inside the engine. They cannot be directly called from JavaScript.
```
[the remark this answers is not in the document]
marjakh: Dropped "elegantly". Only concisely.

src/blog/understanding-ecma-part-1.md
marjakh: (Did a slightly modified version of this suggestion.)

src/blog/understanding-ecma-part-1.md
marjakh: Added a note (not here but above where this "implicit completion" thing is introduced for the first time).

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,230 @@
...
+
+**Abstract operations** are functions defined in the ECMAScript spec; they are defined for the purpose of writing the spec concisely. A JavaScript engine doesn't have to implement them as separate functions inside the engine. They cannot be directly called from JavaScript.
+
+[Spec: Abstract Operations](https://tc39.es/ecma262/#sec-abstract-operations)
+
+## Internal slots and internal methods
+
+**Internal slots** and **internal methods** use names enclosed in \[\[ \]\].
+
+Internal slots are data members (of a JavaScript object or a specification type). They are used for storing the state of the object. Internal methods are member functions (of a JavaScript object or a specification type).
```
syg: Nit: Only JS objects have internal methods. Internal methods are a pre-defined vtable that all JS objects have in the spec. Slots on the other hand are used more liberally.

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,230 @@
...
+> 1. Let `temp` be `Foo()`
+> 2. If `temp` is an abrupt completion, return `temp`
+> 3. Set `temp` to `temp.[[Value]]`
+> 4. Return `NormalCompletion(temp)`
+
+Which is the same as `Return Foo()`; it behaves the same way for both abrupt and normal completions.
+
+## Asserts
+
+Asserts in the spec assert invariant conditions of the algorithms. They are added for clarity, but don't add any requirements to the implementation - the implementation doesn't need to check them.
```
syg: Formatting nit: I was taught in American schools an em dash be used in situations like these.

src/blog/understanding-ecma-part-1.md
marjakh: I'm deciding against a table of contents, I think it would break the flow; as all those terms would be just nonsense before they're explained.

src/blog/understanding-ecma-part-1.md
marjakh: Ah, thanks for pointing out this error!

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,230 @@
...
+>
+> 1. Assert: `Type(O)` is `Object`.
+> 2. Assert: `IsPropertyKey(P)` is `true`.
+> 3. Let `desc` be `? O.[[GetOwnProperty]](P)`.
+> 4. If `desc` is `undefined`, return `false`.
+> 5. Return `true`.
+
+[Spec: HasOwnProperty](https://tc39.es/ecma262#sec-hasownproperty)
+
+But what's an "abstract operation"? What are the things inside \[\[ \]\]? Why is there a "?" in front of a function? What do the asserts mean?
```
mathiasbynens:
```suggestion
But what's an "abstract operation"? What are the things inside `[[ ]]`? Why is there a `?` in front of a function? What do the asserts mean?
```

mathiasbynens: same for other inline code spans below

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Abstract operations
+
+**Abstract operations** are functions defined in the ECMAScript spec; they are defined for the purpose of writing the spec concisely. A JavaScript engine doesn't have to implement them as separate functions inside the engine. They cannot be directly called from JavaScript.
+
+[Spec: Abstract Operations](https://tc39.es/ecma262/#sec-abstract-operations)
+
+## Internal slots and internal methods
+
+**Internal slots** and **internal methods** use names enclosed in \[\[ \]\].
```
mathiasbynens:
```suggestion
**Internal slots** and **internal methods** use names enclosed in `[[ ]]`.
```

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,230 @@
...
+> 1. Let `temp` be `OrdinaryGetOwnProperty(O, P)`
+> 2. Assert: `temp` is not an abrupt completion
+> 3. Let `temp` be `temp.[[Value]]`
+> 4. Return `NormalCompletion(temp)`
+
+Here we assume that `temp` is a brand new temporary variable which doesn't collide with anything else.
+
+We've also used the knowledge that when a return statement returns something else than a Completion Record, it's implicitly wrapped inside a `NormalCompletion`.
+
+### Side track: Return ? Foo()
```
mathiasbynens:
```suggestion
### Side track: `Return ? Foo()`
```

src/blog/understanding-ecma-part-1.md
```diff
@@ -0,0 +1,230 @@
...
+## Asserts
+
+Asserts in the spec assert invariant conditions of the algorithms. They are added for clarity, but don't add any requirements to the implementation &mdash; the implementation doesn't need to check them.
+
+## Moving on
+
+We have built the understanding needed for reading the spec for simple methods like `Object.prototype.hasOwnProperty` and abstract operations like `HasOwnProperty`. They still delegate to other abstract operations, but based on this blog post we should be able to figure out what they do. We'll encounter Property Descriptors, which is just another specification type.
+
+<figure>
+  <img src="/_img/understanding-ecma-part-1-1.svg" height="306" width="1082" alt="Function call graph starting from Object.prototype.hasOwnProperty">
```
mathiasbynens: this needs `loading=lazy` (`npm run img`)

PR #322, opened 2020-01-27
Update useful commands in README.

planglois:
Instead of running `npm start` followed by `npm run watch`, one can do both
using `npm run serve` so mention that in the README.

Also, how to run the markdown linter is useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,5 +11,5 @@ This repository hosts the source code of [v8.dev, the official website of the V8
 `npm run` shows the full list of supported commands. Highlights:
 
 - `npm run build` builds the site into `dist`.
-- `npm run watch` builds the site into `dist` and watches for changes.
-- `npm start` kicks off a local HTTP server.
+- `npm run serve` kicks off a local HTTP server and watches for changes.
+- `npm run lint` to run markdownlint.


planglois:
Hi there!

I'd like to contribute to the v8 website, especially updating some of our Arm pages around debugging and building. So I thought I'd try out the contribution process using a quite README update.

WDYT?

Thanks,
Pierre

README.md
```diff
@@ -11,5 +11,5 @@ This repository hosts the source code of [v8.dev, the official website of the V8
 `npm run` shows the full list of supported commands. Highlights:
 
 - `npm run build` builds the site into `dist`.
-- `npm run watch` builds the site into `dist` and watches for changes.
-- `npm start` kicks off a local HTTP server.
+- `npm run serve` kicks off a local HTTP server and watches for changes.
```
mathiasbynens: Please undo these changes. If you're working on CSS/JS changes for v8.dev, you don't want `watch`.

README.md
```diff
@@ -11,5 +11,5 @@ This repository hosts the source code of [v8.dev, the official website of the V8
...
+- `npm run lint` to run markdownlint.
```
mathiasbynens: This runs before every commit, so I don't think it needs to be called out.

README.md
planglois: Ah, I see `npm run serve` only works when only editing markdown files.

planglois:
Right, these changes were not as useful as I had hoped, I'll drop this PR. At least I got to test the contribution process :-), I'll come back with more!

mathiasbynens:
Thanks for helping out regardless :) We look forward to your next patch!

PR #323, opened 2020-01-30
Make sure figures never get too wide

tomayac:
Currently figures can [break out of the container](https://t.co/tVCsK2rJNI):

6500F1A0-1B4A-4B79-B2FE-3D7E6391677D

This PR hopefully fixes this.

### DIFF
--- a/src/_css/main.css
+++ b/src/_css/main.css
@@ -347,29 +347,9 @@ figcaption {
 }
 iframe {
   border: 0;
-  width: 100%;
 }
 iframe + h2, iframe + h3 {
   margin-top: 1em;
-}
-/* https://alistapart.com/article/creating-intrinsic-ratios-for-video */
-.video {
-  position: relative;
-  padding-bottom: 56.25%; /* 16:9 */
-  padding-top: 25px;
-  margin-bottom: 0.5em;
-  height: 0;
-}
-/* If we ever need other ratios, add them like so:
-.video-16\:9 {
-  padding-bottom: 56.25%;
-}
-*/
-.video iframe {
-  position: absolute;
-  top: 0;
-  left: 0;
-  height: 100%;
 }
 .table-wrapper {
   overflow-x: auto;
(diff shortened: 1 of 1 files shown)


tomayac:
Please try this before merging, I have edited the patch on mobile.

tomayac:
OK, [fixed](https://github.com/v8/v8.dev/pull/323/commits/9d2dcd0748deef65628331f82c98b4960977dfc4) the [fix](https://github.com/v8/v8.dev/pull/323/commits/e47651da0839f1324ab6c84c525327f2bc8764b5). This is now good to be merged, I've had a chance to verify on a real computer.

mathiasbynens:
This is already a great improvement, thanks!

I wonder if we can re-use the pre-existing `.video` styles, which we use for YouTube embeds. That way we solve the overflow issue + have a means to set aspect ratio as well, just like for our non-`<video>` videos.

mathiasbynens:
Re: the above, @RReverser said he'd take a look today.

mathiasbynens:
Here's the fix I had in mind: https://github.com/v8/v8.dev/pull/324

tomayac:
@mathias, PTAL. Until `aspect-ratio` is supported, we fall back to the previous hack. This will allow regular HTML5 `video` and `img` to be hassle-free, and YouTube `iframe` embeds—for the time being—still need the `div` wrapper.

mathiasbynens:
I see. We'd still need the changes to the Markdown to add the wrappers (like in #324) then.

tomayac:
>  We'd still need the changes to the Markdown to add the wrappers 

No, this is all the markup you need (exemplary):

```html
<figure>
  <video autoplay="" loop="" muted="" playsinline="" src="/_img/simd/credit_card.mp4"></video> 
  <figcaption>Card reading</figcaption>
</figure>
```

The CSS makes sure it never gets too wide and is responsive. Try it :-)

mathiasbynens:
@tomayac I might be misunderstanding something. From your latest patch, the code path for `@supports not (aspect-ratio: attr(width) / 1)` clearly relies on the wrappers, so... we need to add them for the fallback to work.

mathiasbynens:
Ooh never mind, it's only for `iframe`s, for which the wrappers are already present. Ignore me! :P

mathiasbynens:
Oh right, because we only need the fallback for the `iframe`s, all of which are already wrapped. Got it!

RReverser:
To confirm - do latest fixes here supersede https://github.com/v8/v8.dev/pull/324 or is that one still necessary as well? This PR definitely looks cleaner and quite neat.

tomayac:
> To confirm - do latest fixes here supersede #324 or is that one still necessary as well?

#324 can be closed, Mathias and I chatted off-Issue and we're happy with the way it turned out here.

RReverser:
Perfect, thanks.

mathiasbynens:
This patch just needs the `width`/`height` attributes on the `<video>` elements, like #324 did.

tomayac:
> This patch just needs the width/height attributes on the <video> elements

It works better when with and height are known, but it also works without (causing a layout shift). I have just opened https://github.com/v8/v8.dev/issues/325 to add the dimensions for videos that lack them.

tomayac:
Alrighty, https://github.com/v8/v8.dev/pull/326 and the present #323 are hopefully ready to be merged.

PR #324, opened 2020-01-31
Fix video styling

mathiasbynens:
This makes sure that https://v8.dev/features/simd renders correctly on narrow viewports, but also that the embedded videos are responsive while maintaining their aspect ratio (similar to other videos on v8.dev, which previously were all YouTube embeds).

Closes #323.

### DIFF
--- a/src/features/simd.md
+++ b/src/features/simd.md
@@ -120,25 +120,33 @@ As per their description, MediaPipe is a framework for building multimodal (eg. 
 One of the most visually appealing demos where it’s easy to observe the difference in performance SIMD makes, is a following hand-tracking system. Without SIMD, you can get only around 3 frames per second on a modern laptop, while with SIMD enabled you get a much smoother experience at 15-16 frames per second.
 
 <figure>
-  <video autoplay muted playsinline loop src="/_img/simd/hand.mp4"></video>
+  <div class="video video-25:9">
+    <video autoplay muted playsinline loop width="600" height="216" src="/_img/simd/hand.mp4"></video>
+  </div>
 </figure>
 
 Visit the [link](https://pursuit.page.link/MediaPipeHandTrackingSimd) in Chrome Canary with SIMD enabled to try it!
 
-Another interesting set of demos that makes use of SIMD for smooth experience, come from OpenCV - a popular computer vision library that can also be compiled to WebAssembly. They’re available by [link](https://bit.ly/opencv-camera-demos), or you can check out the pre-recorded versions below:
+Another interesting set of demos that makes use of SIMD for smooth experience come from OpenCV, a popular computer vision library that can also be compiled to WebAssembly. They’re available by [link](https://bit.ly/opencv-camera-demos), or you can check out the pre-recorded versions below:
 
 <figure>
-  <video autoplay muted playsinline loop src="/_img/simd/credit-card.mp4"></video>
+  <div class="video video-25:9">
+    <video autoplay muted playsinline loop width="256" height="512" src="/_img/simd/credit-card.mp4"></video>
+  </div>
   <figcaption>Card reading</figcaption>
 </figure>
 
 <figure>
-  <video autoplay muted playsinline loop src="/_img/simd/invisibility-cloak.mp4"></video>
+  <div class="video">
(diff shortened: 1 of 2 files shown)


src/features/simd.md
```diff
@@ -120,25 +120,33 @@ As per their description, MediaPipe is a framework for building multimodal (eg.
...
 
 Visit the [link](https://pursuit.page.link/MediaPipeHandTrackingSimd) in Chrome Canary with SIMD enabled to try it!
 
-Another interesting set of demos that makes use of SIMD for smooth experience, come from OpenCV - a popular computer vision library that can also be compiled to WebAssembly. They’re available by [link](https://bit.ly/opencv-camera-demos), or you can check out the pre-recorded versions below:
+Another interesting set of demos that makes use of SIMD for smooth experience come from OpenCV, a popular computer vision library that can also be compiled to WebAssembly. They’re available by [link](https://bit.ly/opencv-camera-demos), or you can check out the pre-recorded versions below:
 
 <figure>
-  <video autoplay muted playsinline loop src="/_img/simd/credit-card.mp4"></video>
+  <div class="video video-25:9">
+    <video autoplay muted playsinline loop width="256" height="512" src="/_img/simd/credit-card.mp4"></video>
```
mathiasbynens: Side note: we should make these available in open formats, too. I can't view them in my custom Chromium build, for example.

tomayac: Not particularly enthused about the additional wrapper… It's a problem that's [going away](https://groups.google.com/a/chromium.org/g/blink-dev/c/GePU9T8UpEc/m/bET6Mi60CgAJ).

mathiasbynens:
I love keeping it simple! :) This patch just builds upon what is already there for v8.dev.

I'm definitely open to dropping the hacky aspect ratio CSS + removes the wrappers, once stable browsers support it natively. IMHO that's separate to this patch which only aims to address the currently broken layout.

src/features/simd.md
mathiasbynens: I never liked that wrapper! Excited that we'll be able to drop it soon :)

tomayac:
You can drop it now! 

```css
img,
video {
  height: auto;
  max-width: 100%;
}
```

…and… 

```html
<video width="640" height="480">[…]</video>
<img src="foo.png" width="640" height="480">
```

…does the job. Today. Since your videos are typically wider than the container anyway, you get the fluidity for free.

mathiasbynens:
@tomayac Do you want to update your PR #323 to remove the wrappers then?

tomayac:
Actually, add `iframe[src*="youtube"]` for YouTube embeds. 

```css
img,
video,
iframe[src*="youtube"] {
  height: auto;
  max-width: 100%;
}

iframe[src*="youtube"] {
  aspect-ratio: attr(width) / attr(height);
}
```

The user-agent stylesheet covers the aspect ratios for images and videos, for YouTube iframes we have to do it manually.

tomayac:
> @tomayac Do you want to update your PR #323 to remove the wrappers then?

Will do.

RReverser:
Closing in favour of #323.

mathiasbynens:
Note that we'd still want to add `width`/`height` to the `<video>`. #323 doesn't do that, currently.

tomayac:
> Note that we'd still want to add width/height to the <video>.

Let's [automate](https://www.npmjs.com/package/get-video-dimensions) that, similar to [`add-image-dimensions.js`](https://github.com/v8/v8.dev/blob/master/add-image-dimensions.js). 

I open a separate issue.

src/features/simd.md
RReverser: We'll probably want to kep it for other browsers (although I agree it's ugly :( ).

src/features/simd.md
```diff
@@ -120,25 +120,33 @@ As per their description, MediaPipe is a framework for building multimodal (eg.
...
+  </div>
   <figcaption>Card reading</figcaption>
 </figure>
 
 <figure>
-  <video autoplay muted playsinline loop src="/_img/simd/invisibility-cloak.mp4"></video>
+  <div class="video">
```
RReverser: Why do we include `video-25:9` in other videos but not here? (and, more generally, how do we choose class for non-standard dimensions)

tomayac: This is obsolete. #323 doesn't need the wrapper for `video`.

PR #326, opened 2020-01-31
Automatically add video dimensions

### LINKED ISSUE #325: Automatically add video dimensions
Let's [automate](https://www.npmjs.com/package/get-video-dimensions) that, similar to [`add-image-dimensions.js`](https://github.com/v8/v8.dev/blob/master/add-image-dimensions.js).

tomayac:
Fixes #325

Note that this requires ffmpeg to be installed

```bash
brew install ffmpeg
```


add-image-and-video-dimensions.js
```diff
@@ -0,0 +1,46 @@
+const fs = require('fs');
+
+const glob = require('glob');
+const imageSize = require('image-size');
+const videoSize = require('get-video-dimensions');
+
+(async () => {
+  const files = glob.sync('src/**/*.{html,md}');
+  console.log(files)
```
mathiasbynens:
```suggestion

```

add-image-and-video-dimensions.js
```diff
@@ -0,0 +1,45 @@
...
+      console.log('>>');
+      console.log(updatedLine);
+      console.log('------------');
+      updatedContents = updatedContents.replace(oldLine, updatedLine);
+    }
+
+    // TODO: This needs updating once we use `<source>`.
+    const videos = contents.matchAll(/^\s*<video.*$/gm);
+    if (!videos) continue;
+    for await (const result of videos) {
```
mathiasbynens:
```suggestion
    for (const result of videos) {
```

add-image-and-video-dimensions.js
```diff
@@ -0,0 +1,45 @@
...
+const glob = require('glob');
+const imageSize = require('image-size');
+const videoSize = require('get-video-dimensions');
+
+(async () => {
+  const files = glob.sync('src/**/*.{html,md}');
+  for (const file of files) {
+    const contents = fs.readFileSync(file, 'utf8').toString();
+    let updatedContents = contents;
+    const images = contents.matchAll(/^\s*<img.*$/gm);
```
RReverser: Looks like copy-paste mistake. I suppose this should check for videos?

RReverser: Ah, I see, you merged both scripts. Then it should check for both images and videos, rather than continuing right away. Otherwise, a page with just videos and no images would currently be skipped altogether.

add-image-and-video-dimensions.js
```diff
@@ -0,0 +1,45 @@
+const fs = require('fs');
+
+const glob = require('glob');
+const imageSize = require('image-size');
+const videoSize = require('get-video-dimensions');
+
+(async () => {
+  const files = glob.sync('src/**/*.{html,md}');
+  for (const file of files) {
+    const contents = fs.readFileSync(file, 'utf8').toString();
```
RReverser: Since you're in `async` context anyway, might be better to use `fs.promises.*` here and everywhere else.

You can just import them at the top as

```js
const fs = require('fs').promises;
```

and then use async APIs everywhere.

Additionally, you could extract file processing into a function to be able to handle files in parallel.

add-image-and-video-dimensions.js
tomayac: 🤦‍♂️Duh…

add-image-and-video-dimensions.js
tomayac: Correct. Thanks.

add-image-and-video-dimensions.js
tomayac: Very good catch. I have changed the logic now.

add-image-and-video-dimensions.js
tomayac: I use the async file methods now, but have refrained from optimizing further. It's really just fast enough :-)

PR #327, opened 2020-02-04
Add a blog post: Understanding the ECMAScript spec, part 2

marjakh:


src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,305 @@
...
+
+The prototype chain walk is inside step 3: if we don't find the property as an own property, we recurse into the prototype's `[[Get]]` method.
+
+## Where's the Receiver coming from?
+
+`OrdinaryGet` also passes the original `Receiver` throughout the recursion, unchanged. The `Receiver` is only used if the property is an accessor property, in which case it's passed as the **this value** when calling the getter function.
+
+Let's find out where the `Receiver` is coming from!
+
+Searching for places where `[[Get]]` is called we find an abstract operation `GetValue` which operates on References. Reference is a specification type, consisting of a base value (in this case, an Object), the reference name (in this case, a String), and a strict mode flag (bool).
```
gahaas: maybe `(in this case, a String)` => `(in this case, the string 'foo')`?

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,305 @@
...
+> 5. If `IsPropertyReference(V)` is `true`, then
+> a. If `HasPrimitiveBase(V)` is `true`, then
+> i. Assert: In this case, `base` will never be `undefined` or `null`.
+> ii. Set `base` to `! ToObject(base)`.
+> b. Return `? base.[[Get]](GetReferencedName(V), GetThisValue(V))`.
+> 6. Else,
+> a. Assert: `base` is an Environment Record.
+> b. Return `? base.GetBindingValue(GetReferencedName(V), IsStrictReference(V))`
+
+For property references such as `o.foo`, we take the branch 5 b. The `Receiver` we pass is `GetThisValue(V)`. In this case, it's just the base of the Reference:
```
gahaas: Do you think it would be good to change the format of `base` here and before to show clearly that this is the name of a field, not a name that makes sense?

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,305 @@
...
+
+> [`GetThisValue( V )`](https://tc39.es/ecma262/#sec-getthisvalue)
+>
+> 1. Assert: `IsPropertyReference(V)` is `true`.
+> 2. If `IsSuperReference(V)` is `true`, then
+> a. Return the value of the `thisValue` component of the reference `V`.
+> 3. Return `GetBase(V)`.
+
+Interesting! The `this` inside the getter refers to the original object where we tried to get the property from, not the one where we found the property during the prototype chain walk. Let's try it out!
```
gahaas: I don't get it. Do we take `2.a` here? What is `IsSuperReference`?

src/blog/understanding-ecmascript-part-2.md
gahaas: I got more and more confused below. Could you make the example super explicit here? 
In our example, the reference consists of the following values: base = o1 (or o2? I don't know); reference name = `foo`, strict mode flag = true.

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,305 @@
...
+```javascript
+const o = { x: 10, get sneaky() { return this.x; } };
+let o2 = {};
+o2.__proto__ = o;
+o2.sneaky; // will return 10
+o2.x = 50;
+o2.sneaky; // will return 50
+o.sneaky; // will return 10
+```
+
```
gahaas: Could you walk the reader through this example?

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,305 @@
...
+> 5. Perform `? DestructuringAssignmentEvaluation` of `assignmentPattern` using `rval` as the argument.
+> 6. Return `rval`.
+
+Now the `LeftHandSideExpression` is not an object literal or an array literal, so we take the if branch in step 1. The assignment expression is `o.foo`, and it's not a function definition, so we take the else branch in 1 d. There we evaluate the `AssignmentExpression` and call `GetValue` on it.
+
+In any case, `GetValue` will be called on the Reference which is the result of evaluating the `MemberExpression`. Thus, we know that the Object internal method `[[Get]]` will get invoked when accessing a property on an Object, and the prototype chain walk will occur.
+
+### Why is `o2.foo` an `AssignmentExpression`?
+
+Just one more thing. Previously, we used `o2.foo` as an `AssignmentExpression`. It doesn't look like one though. Why is it an `AssignmentExpression`?
```
gahaas: What does this "Previously" refer to? Isn't o2.foo used the last time before `AssignmentExpression` is even defined?

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,305 @@
...
+> c. Return `? parent.[[Get]](P, Receiver)`.
+> 4. If `IsDataDescriptor(desc)` is `true`, return `desc.[[Value]]`.
+> 5. Assert: `IsAccessorDescriptor(desc)` is `true`.
+> 6. Let `getter` be `desc.[[Get]]`.
+> 7. If `getter` is `undefined`, return `undefined`.
+> 8. Return `? Call(getter, Receiver)`.
+
+The prototype chain walk is inside step 3: if we don't find the property as an own property, we recurse into the prototype's `[[Get]]` method.
+
+## Where's the Receiver coming from?
```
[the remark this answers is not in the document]
mathiasbynens: +1 I don't think many JS devs use (or understand) the term "receiver". Let's briefly explain it.

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,305 @@
...
+
+We know that properties are looked up in the prototype chain: if an object doesn't have the property we're trying to read, we walk up the prototype chain until we find it (or find an object which no longer has a prototype).
+
+For example:
+
+```javascript
+const o1 = {'foo' : 2511};
+const o2 = {};
+console.log(o2.foo); // undefined
+o2.__proto__ = o1;
```
mathiasbynens: Let's use `Object.setPrototypeOf` instead of `__proto__` (throughout the entire article). `__proto__` is a legacy web compat wart, and `Object.setPrototypeOf` is more readable.

src/blog/understanding-ecmascript-part-2.md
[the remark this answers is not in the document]
marjakh: It's explained above: "The `Receiver` is only used if the property is an accessor property, in which case it's passed as the **this value** when calling the getter function."

--> I added a reminder of that here.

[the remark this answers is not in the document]
marjakh: s/above/below, it's actually explained right in the first paragraph of this section :)

--> Reworded the paragraph so that the explanation comes first.

I also added an explanation to the place where the term Receiver occurs the first time.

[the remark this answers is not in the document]
marjakh: Also changed the title to: What's Receiver and where is it coming from? to make it clear that we don't need to understand what Receiver is at this point of the post.

src/blog/understanding-ecmascript-part-2.md
marjakh: Clarified: For `o2.foo`, we don't take the branch in step 2, since it's not a super reference (such as `super.foo`), but we take step 3 and return the base value of the reference which is `o2`.

src/blog/understanding-ecmascript-part-2.md
marjakh: Added

src/blog/understanding-ecmascript-part-2.md
marjakh: Oops, this was supposed to say o.foo. However, I fixed the examples so that all access o2.foo instead of o.foo, so it's more consistent (the blog post talks about accessing o2.foo throughout).

marjakh: Also added a forward-reference to the place where I say o2.foo is an AssignmentExpression.

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,342 @@
...
+>         1. Let `rval` be `? GetValue(rref)`.
+>     1. Perform `? PutValue(lref, rval)`.
+>     1. Return `rval`.
+> 1. Let `assignmentPattern` be the `AssignmentPattern` that is covered by `LeftHandSideExpression`.
+> 1. Let `rref` be the result of evaluating `AssignmentExpression`.
+> 1. Let `rval` be `? GetValue(rref)`.
+> 1. Perform `? DestructuringAssignmentEvaluation` of `assignmentPattern` using `rval` as the argument.
+> 1. Return `rval`.
+
+Now the `LeftHandSideExpression` (`x`) is not an object literal or an array literal, so we take the if branch in step 1. In step 1 a, we evaluate the left-hand side (`x`) and store the result into `lref`. `o2.foo` is not a function definition, so we don't take the if branch in 1 c, but the else branch in 1 d. There we evaluate `o2.foo` and call `GetValue` on it.
```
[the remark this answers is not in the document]
marjakh: This is a known issue - Mathias said he'll take care of the custom formatting. (The problem is making markdown do "1. foo a. subfoo i. subsubfoo" - style formatting. I have no idea how to, I tried but failed...)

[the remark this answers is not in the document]
RReverser: Looks like this will require either changing our CSS or using inline HTML like:

```html
<ol type="a">
<li>123
<li>456
</ol>
```

which is not as "good" as CSS version, but consistently renders as

<ol type="a">
<li>123
<li>456
</ol>

even on Github.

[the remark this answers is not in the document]
RReverser: (update based on the chat: found a better solution using `<style>` on the page to get consistent formatting; still not perfect, because, well, it uses `<style>` on the page, so looking into something more viable in `markdown-it`)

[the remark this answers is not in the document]
RReverser: @mathiasbynens On the second thought, do you think it would be useful to change styles by default on entire blog?

There already is (and probably will be) lots of quotes from the ECMAScript spec (duh, given the nature of the blog) and it could be nicer to keep nested list styles consistent everywhere.

FWIW this is what I did locally to style this one page and could add this to global CSS instead:
```css
ol > li > ol { list-style-type: lower-alpha; }
ol > li > ol > li > ol { list-style-type: lower-roman; }
```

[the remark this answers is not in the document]
mathiasbynens: I was thinking an explicit wrapper a la `:::ecmascript-algorithm` (like `:::table-wrapper`) would be the best solution. We could then scope the styles to that wrapper, and we'd still be free to change the global styles. WDYT?

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,342 @@
...
+> [`GetThisValue( V )`](https://tc39.es/ecma262/#sec-getthisvalue)
+>
+> 1. Assert: `IsPropertyReference(V)` is `true`.
+> 2. If `IsSuperReference(V)` is `true`, then
+> a. Return the value of the `thisValue` component of the reference `V`.
+> 3. Return `GetBase(V)`.
+
+For `o2.foo`, we don't take the branch in step 2, since it's not a super reference (such as `super.foo`), but we take step 3 and return the base value of the reference which is `o2`.
+
+Piecing everything together, we find out that we set the `Receiver` to be the base of the original reference, and then we keep it unchanged during the prototype chain walk. Finally, if the property we find is an accessor property, we use the `Receiver` as the **this value** when calling it.
```
syg: What happened to explaining how `o2.foo` gets `2511` before switching gears to getters? There should be a section about the proto walk bottoming out and ultimately going through `o.[[GetOwnProperty]]`.

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,342 @@
...
+o2.foo; // will return 50
+```
+
+In this example, we have an accessor property called `foo` and we define a getter for it. The getter returns `this.x`.
+
+Then we access `o2.foo` - what does the getter return?
+
+We found out that when we call the getter, the **this value** is the object where we originally tried to get the property from, not the object where we found it. In this case the **this value** is `o2`, not `o1`. We can verify that by checking whether the getter returns `o2.x` or `o1.x`, and indeed, it returns `o2.x`.
+
+(Note that setting `o2.x = 50` adds a property called `x` in `o2` and doesn't overwrite the property `x` in `o1`.)
```
syg: Setters on the proto are actually super confusing when following through the spec. If `o1.x` was non-writable, this would error/silently nop. If `o1.x` was a setter, that would invoke the setter and not create an `o2.x`. I think for the scope of the tutorial we should remove mentions of property sets.

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,342 @@
...
+
+and
+
+```javascript
+x = (y = 5);
+x; // 5
+y; // 5
+```
+
+`o2.foo` is a "degenerate" `AssignmentExpression` which doesn't assign anything. This follows from the following grammar productions, each one taking the "simplest" or "most degenerate" case until the last one:
```
syg: What does degenerate mean here? The productions below and the layering encodes operator precedence and how closely expressions bind, not about simplicity.

src/blog/understanding-ecmascript-part-2.md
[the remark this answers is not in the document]
mathiasbynens: PR: https://github.com/v8/v8.dev/pull/332

src/blog/understanding-ecmascript-part-2.md
marjakh: Right, removed the setting, it's much simpler that way.

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,342 @@
...
+console.log(o2.foo); // 2511
+```
+
+## Where's the prototype walk defined?
+
+Let's try to find out where this behavior is defined. A good place to start is a list of [Object Internal Methods](https://tc39.es/ecma262/#sec-object-internal-methods-and-internal-slots).
+
+There's both `[[GetOwnProperty]]` and `[[Get]]` &mdash; now we're interested in the version that isn't restricted to _own_ properties, so we'll go with `[[Get]]`.
+
+Unfortunately, the [Property Descriptor specification type](https://tc39.es/ecma262/#sec-property-descriptor-specification-type) also has a field called `[[Get]]`, so while browsing the spec for `[[Get]]`, we need to carefully distinguish between the two independent usages.
```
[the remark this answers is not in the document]
marjakh: Added this:

`[[Get]]` is an **essential internal method**. **Ordinary objects** implement the default behavior for essential internal methods, whereas **exotic objects** can define their own internal method `[[Get]]` which deviates from the default behavior. In this post, we focus on ordinary objects.

The default implementation for `[[Get]]` delegates to `OrdinaryGet`:

...

src/blog/understanding-ecmascript-part-2.md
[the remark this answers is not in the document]
RReverser: We've merged #332; @marjakh can you please update the post using the mentioned wrapper?

syg:
@googlebot I fixed it.

src/blog/understanding-ecmascript-part-2.md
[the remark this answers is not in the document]
marjakh: Done!

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,342 @@
...
+
+The reference in our example is `o2.foo` which is a property reference. So we take branch 5. We don't take the branch in 5 a, since the base (`o2`) is not a primitive value (a Boolean, String, Symbol, BigInt or Number).
+
+Then we call `[[Get]]` in step 5 b. The `Receiver` we pass is `GetThisValue(V)`. In this case, it's just the basevalue of the Reference:
+
+> [`GetThisValue( V )`](https://tc39.es/ecma262/#sec-getthisvalue)
+>
+> 1. Assert: `IsPropertyReference(V)` is `true`.
+> 2. If `IsSuperReference(V)` is `true`, then
+> a. Return the value of the `thisValue` component of the reference `V`.
```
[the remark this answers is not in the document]
marjakh: This was fixed by the formatting change.

src/blog/understanding-ecmascript-part-2.md
marjakh: As discussed, removed the word "generate" and explained it differently

syg:
@googlebot I consent.

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,342 @@
...
+
+> [`[[Get]] ( P, Receiver )`](https://tc39.es/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots-get-p-receiver)
+>
+> When the `[[Get]]` internal method of `O` is called with property key `P` and ECMAScript language value `Receiver`, the following steps are taken:
+>
+> 1. Return `? OrdinaryGet(O, P, Receiver)`.
+
+We'll see shortly that `Receiver` is the value which is used as the **this value** when calling a getter function of an accessor property.
+
+`OrdinaryGet` is defined like this:
```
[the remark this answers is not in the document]
marjakh: Added

> Warning! This episode contains copy-pasted algorithms from the [ECMAScript spec](https://tc39.es/ecma262/) as of February 2020. They'll eventually be out of date.

src/blog/understanding-ecmascript-part-2.md
marjakh: Added the following paragraphs here:

Let's look at how this algorithm works when we access `o2.foo`. First we invoke `OrdinaryGet` with `O` being `o2` and `P` being `"foo"`. `O.[[GetOwnProperty]]` returns `undefined`, since `o2` doesn't have an own property called `"foo"`, so we take the if branch in step 3. In step 3.a, we set `parent` to the prototype of `o2` which is `o1`. `parent` is not `null`, so we don't return in step 3.b. In step 3.c, we call the parent's `[[Get]]` method with property key `"foo"`, and return whatever it returns.

The parent (`o1`) is an ordinary object, so its `[[Get]]` method invokes `OrdinaryGet` again, this time with `O` being `o1` and `P` being `"foo"`. In step 2, `O.[[GetOwnProperty]]("foo")` returns the property descriptor associated to `"foo"`. The property descriptor is not `undefined`, so we don't take the if branch in step 3. Next we execute step 4. The property descriptor is a data descriptor, so we return its value, `99`, in step 4, and we're done.

---- but I didn't change the overall structure of the post, so, the "Receiver" side track is still where it was :)

marjakh: (Also simplified the example and changed the number value because I started liking 99 more.)

src/blog/understanding-ecmascript-part-2.md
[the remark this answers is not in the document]
marjakh: > Warning! This episode contains copy-pasted algorithms from the [ECMAScript spec](https://tc39.es/ecma262/) as of February 2020. They'll eventually be out of date.

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,342 @@
...
+> 5. If `IsPropertyReference(V)` is `true`, then
+> a. If `HasPrimitiveBase(V)` is `true`, then
+> i. Assert: In this case, `base` will never be `undefined` or `null`.
+> ii. Set `base` to `! ToObject(base)`.
+> b. Return `? base.[[Get]](GetReferencedName(V), GetThisValue(V))`.
+> 6. Else,
+> a. Assert: `base` is an Environment Record.
+> b. Return `? base.GetBindingValue(GetReferencedName(V), IsStrictReference(V))`
+
+The reference in our example is `o2.foo` which is a property reference. So we take branch 5. We don't take the branch in 5 a, since the base (`o2`) is not a primitive value (a Boolean, String, Symbol, BigInt or Number).
```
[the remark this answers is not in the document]
marjakh: Made this consistent; using Reference like the spec does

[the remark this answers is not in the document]
marjakh: ... using the term "property reference" (capitalized like that) like the spec does too.

src/blog/understanding-ecmascript-part-2.md
marjakh: Added this:


Let's look at how this algorithm works when we access `o2.foo`. First we invoke `OrdinaryGet` with `O` being `o2` and `P` being `"foo"`. `O.[[GetOwnProperty]]("foo")` returns `undefined`, since `o2` doesn't have an own property called `"foo"`, so we take the if branch in step 3. In step 3.a, we set `parent` to the prototype of `o2` which is `o1`. `parent` is not `null`, so we don't return in step 3.b. In step 3.c, we call the parent's `[[Get]]` method with property key `"foo"`, and return whatever it returns.

The parent (`o1`) is an ordinary object, so its `[[Get]]` method invokes `OrdinaryGet` again, this time with `O` being `o1` and `P` being `"foo"`. `o1` has a property called `"foo"`, so in step 2, `O.[[GetOwnProperty]]("foo")` returns the associated Property Descriptor.

[Property Descriptor](https://tc39.es/ecma262/#sec-property-descriptor-specification-type) is a specification type. Data Property Descriptors store the value of the property directly in the `[[Value]]` field. Accessor Property Descriptors store the accessor functions in fields `[[Get]]` and/or `[[Set]]`.

The Property Descriptor returned by `O.[[GetOwnProperty]]("foo")` in step 2 is not `undefined`, so we don't take the if branch in step 3. Next we execute step 4. The Property Descriptor is a Data Descriptor, so we return its `[[Value]]` field, `99`, in step 4, and we're done.

marjakh: (Also changed to 99 because it's a nicer number.)

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,342 @@
...
+> a. Let `parent` be `? O.[[GetPrototypeOf]]()`.
+> b. If `parent` is `null`, return `undefined`.
+> c. Return `? parent.[[Get]](P, Receiver)`.
+> 4. If `IsDataDescriptor(desc)` is `true`, return `desc.[[Value]]`.
+> 5. Assert: `IsAccessorDescriptor(desc)` is `true`.
+> 6. Let `getter` be `desc.[[Get]]`.
+> 7. If `getter` is `undefined`, return `undefined`.
+> 8. Return `? Call(getter, Receiver)`.
+
+The prototype chain walk is inside step 3: if we don't find the property as an own property, we recurse into the prototype's `[[Get]]` method.
```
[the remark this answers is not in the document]
marjakh: Reworded this. As an non-native English speaker, it would never occur to me to use the word "recur" here, and I think it would be very surprising for other non-native speakers. Trying to minimize linguistic surprises here, so rewording it like this:

The prototype chain walk is inside step 3: if we don't find the property as an own property, we call the prototype's `[[Get]]` method. If we don't find the property there either, we call it's prototype's `[[Get]]` method, and so on, until we reach an object without a prototype.

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,358 @@
...
+> 1. Assert: `identifierName` is an `IdentifierName`
+> 1. Let `bv` be `? RequireObjectCoercible(baseValue)`.
+> 1. Let `propertyNameString` be `StringValue` of `identifierName`.
+> 1. Return a value of type Reference whose base value component is `bv`, whose referenced name component is `propertyNameString`, and whose strict reference flag is `strict`.
+
+That is: `EvaluatePropertyAccessWithIdentifierKey` constructs a Reference which uses the provided `baseValue` as the base, the string value of `identifierName` as the property name, and `strict` as the strict mode flag.
+
+Eventually this Reference gets passed to `GetValue`. This is defined in several places in the spec, depending on how the Reference ends up being used.
+
+### Property access as a parameter
```
syg: I found the first section compelling and cohesive: let's take a behavior we all know and thoroughly trace through how it's specified. However, I'm missing the transition to these next two sections. It feels a bit abrupt.

Is the idea to highlight a few places where property gets can appear, and then work those out? Why were function arguments and RHS of assignments chosen as the examples?

I actually think this episode should end here with the grammar parts left for the next episode. The narrative is "let's learn how a feature is specified from the bottom up". Episode 1 was the spec mechanism underpinnings: the definitions you need to even navigate the spec. Episode 2 builds on Episode 1 to completely explain how, given an expression like `o.p`, how that expression is specified. Episode 3 can then focus on the grammar: now that we know how `o.p` works, how does the spec get a `o.p`? You can then flesh out a full, but very simple program like `x = o2.foo` or even just `o2.foo`. Then you can dig into the grammar productions a bit more, with its recursive cases and so on. WDYT?

marjakh: As discussed, moved part to an extra content, got rid of one of the "property access" subsections.

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,358 @@
...
+> 3. Let `base` be `GetBase(V)`.
+> 4. If `IsUnresolvableReference(V)` is `true`, throw a `ReferenceError` exception.
+> 5. If `IsPropertyReference(V)` is `true`, then
+>     1. If `HasPrimitiveBase(V)` is `true`, then
+>         1. Assert: In this case, `base` will never be `undefined` or `null`.
+>         1. Set `base` to `! ToObject(base)`.
+>     1. Return `? base.[[Get]](GetReferencedName(V), GetThisValue(V))`.
+> 6. Else,
+>     1. Assert: `base` is an Environment Record.
+>     1. Return `? base.GetBindingValue(GetReferencedName(V), IsStrictReference(V))`
```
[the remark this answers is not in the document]
marjakh: Historical leftovers (formatting change). It shouldn't matter, both mean the same. Changed it now anyway.

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -0,0 +1,234 @@
...
+> 1. Let `arg` be `? GetValue(ref)`.
+> 1. Return a List whose sole item is `arg`.
+
+`o2.foo` doesn't look like an `AssignmentExpression` but it is one, so this production is applicable. To find out why, you can check out this [extra content](/blog/extras/understanding-ecmascript-part-2-extra), but it's not strictly necessary at this point.
+
+The `AssignmentExpression` in step 1 is `o2.foo`. `ref`, the result of evaluating `o2.foo`, is the above mentioned Reference. In step 2 we call `GetValue` on it. Thus, we know that the Object internal method `[[Get]]` will get invoked, and the prototype chain walk will occur.
+
+## Summary
+
+In this episode, we looked at how a language feature, in this case prototype lookup, in defined across all the different layers: the syntactic constructs that trigger the feature and the algorithms defining it.
```
chicoxyzzy:
```suggestion
In this episode, we looked at how a language feature, in this case prototype lookup, is defined across all the different layers: the syntactic constructs that trigger the feature and the algorithms defining it.
```

marjakh: (Reworded)

PR #331, opened 2020-02-14
Add Intl.DisplayNames feature explainer

syg:


src/features/intl-displaynames.md
```diff
@@ -0,0 +1,112 @@
+---
+title: '`Intl.DisplayNames`'
+author: 'Shu-yu Guo ([@_shu](https://twitter.com/_shu)) and Frank Tang'
+avatars:
+  - 'shu-yu-guo'
+  - 'frank-tang'
+date: 2020-02-13
+tags:
+  - Intl
+  - Node.js 14
```
RReverser: I don't think this is a valid tag (with spaces), and on the previous [Intl post](https://v8.dev/features/intl-numberformat) we just included `Intl` - let's do the same here?

src/features/intl-displaynames.md
```diff
@@ -0,0 +1,112 @@
...
+languageNames.of('zh');
+// → '中文'
+languageNames.of('de');
+// → '德文'
+```
+
+The following example gets currency names in Simplified Chinese using [ISO-4217 3-letter currency codes](https://www.iso.org/iso-4217-currency-codes.html).
+
+```js
+const currencyNames = new Intl.DisplayNames(['zh-Hans'], {type: 'currency'});
```
RReverser: Are these for singular currency names of plural ones? Is there a way to customise this via options? (seems like an obvious question reader would ask)

src/features/intl-displaynames.md
mathiasbynens: It's a valid tag (search for `Node.js 12`, for example). But generally we only add these tags for Node.js versions that have been released already.

RReverser: If you open any of the posts containing that tag in the source, you can actually see that such tags with a space also seem to have silently broken the parser and both them and any following tags are not rendered on the pages.

For example: https://github.com/v8/v8.dev/blob/f880711d0ebd70413e3a3cecb1e7f3dbaaa969ab/src/features/intl-relativetimeformat.md#L7-L10

contains `Intl`, `Node.js 12` and `io19` in the source, but

https://v8.dev/features/intl-relativetimeformat

renders only `Intl`.

mathiasbynens: This is not a bug, in fact we have special logic just to achieve this: https://github.com/v8/v8.dev/blob/b93cbfc168b4b19ecebf5a8c93ada7ec8cf3bdd2/src/_includes/features.njk#L7 We don't want #io and Node.js tags to show up on the overview, but we do want to be able to link to them.

RReverser: That... wasn't very obvious, but okay :)

src/features/intl-displaynames.md
syg: Well, depends on the language, right? For those that have singular/plural split, it's singular. For languages that have cases, it's nominative IIRC. This API doesn't return plurals for languages that have them.

syg: What's the action item you're asking for here? Clarification that it's only singular names in the text?

syg: Oh, and also, to get at the plural long-form currency names, [Intl.NumberFormat](https://v8.dev/features/intl-numberformat) can be used.

RReverser: > Clarification that it's only singular names in the text?

Yeah, if that's the case. Probably worth including a note about `NumberFormat` (and a link to that other explainer) as well.

RReverser: ...which you already did :)

src/features/intl-displaynames.md
```diff
@@ -40,7 +40,7 @@ languageNames.of('de');
 // → '德文'
 ```
 
-The following example gets currency names in Simplified Chinese using [ISO-4217 3-letter currency codes](https://www.iso.org/iso-4217-currency-codes.html).
+The following example gets currency names in Simplified Chinese using [ISO-4217 3-letter currency codes](https://www.iso.org/iso-4217-currency-codes.html). In languages that have distinct singular and plural forms, the currency names are singular. For plural forms, [Intl.NumberFormat](https://v8.dev/features/intl-numberformat) may be used.
```
RReverser:
```suggestion
The following example gets currency names in Simplified Chinese using [ISO-4217 3-letter currency codes](https://www.iso.org/iso-4217-currency-codes.html). In languages that have distinct singular and plural forms, the currency names are singular. For plural forms, [`Intl.NumberFormat`](https://v8.dev/features/intl-numberformat) may be used.
```

src/features/intl-displaynames.md
```diff
@@ -0,0 +1,112 @@
...
+- When `type` is `"currency"`, `code` must be a [ISO-4217 3-letter currency code](https://www.iso.org/iso-4217-currency-codes.html).
+- When `type` is `"script"`, `code` must be a [ISO-15924 4-letter script code](https://unicode.org/iso15924/iso15924-codes.html).
+
+## Conclusion
+
+Like other `Intl` APIs, as `Intl.DisplayNames` become more widely available, libraries and applications will opt to drop packaging and shipping their own translation data in favor of using the native functionality.
+
+## `Intl.DisplayNames` support { #support }
+
+<feature-support chrome="81 /blog/v8-release-81#intl.displaynames"
```
mathiasbynens: Note that the V8 v8.1 release blog post hasn't happened yet. @natorion What's the status on that?

RReverser:
> LGTM. Let's publish this next Tuesday?

I was initially thinking tomorrow. Do you want to wait for V8 release post first?

RReverser:
I'm fine with Tuesday, and then we can publish https://github.com/v8/v8.dev/pull/334 either same day or Wednesday (since it references this feature explainer).

mathiasbynens:
I'd suggest spacing things out a bit. Publishing #334 on Wednesday sounds great!

RReverser:
> I'd suggest spacing things out a bit. Publishing #334 on Wednesday sounds great!

My motivation for suggesting posting them together is that the release post is pretty much entirely this feature's announcement. But I guess it doesn't harm putting them a day apart.

mathiasbynens:
We'd be competing with our own announcement :/ Better to space things out.

RReverser:
Ugh not again :(

RReverser:
@syg I've manually amended last commit to have your Chromium email like other commits instead. Please configure git to use same email at least per-repo for Google projects.

src/features/intl-displaynames.md
RReverser: Missed that this one produces a broken table; on it.

PR #332, opened 2020-02-19
Add `:::ecmascript-algorithm` wrapper with styles

mathiasbynens:
Ref. https://github.com/v8/v8.dev/pull/331#discussion_r380717839.

cc @marjakh

### DIFF
--- a/src/blog/understanding-ecmascript-part-1.md
+++ b/src/blog/understanding-ecmascript-part-1.md
@@ -27,6 +27,7 @@ In the example, `o` doesn’t have a property called `hasOwnProperty`, so we wal
 
 To describe how `Object.prototype.hasOwnProperty` works, the spec uses pseudocode-like descriptions:
 
+:::ecmascript-algorithm
 > **[`Object.prototype.hasOwnProperty(V)`](https://tc39.es/ecma262#sec-object.prototype.hasownproperty)**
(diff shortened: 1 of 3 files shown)


src/blog/understanding-ecmascript-part-1.md
```diff
@@ -117,29 +123,42 @@ If an algorithm throws an exception, it means returning a Completion Record with
 
 [`ReturnIfAbrupt(argument)`](https://tc39.es/ecma262/#sec-returnifabrupt) means taking the following steps:
 
+:::ecmascript-algorithm
+<!-- markdownlint-disable blanks-around-lists -->
```
mathiasbynens: We only need these when `:::ecmascript-algorithm` contains nothing but a list.

RReverser: Is there any way to configure markdownlint to ignore all these instances?

mathiasbynens: Not that I know of. I don't think we want it for all of them though.

RReverser: Okay, let's go with this for now, we can always iterate later.

RReverser:
Thanks @mathiasbynens!

PR #336, opened 2020-02-26
Add blog post: Understanding ecmascript part 3

marjakh:


src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,320 @@
...
+
+If you're not familiar with [context-free grammars](https://en.wikipedia.org/wiki/Context-free_grammar), now it's a good idea to check out the basics, since the spec uses context-free grammars to define the language.
+
+## ECMAScript grammars
+
+ECMAScript source text is a sequence of Unicode code points. Each Unicode code point is an integral value between `U+0000` and `U+10FFFF`. The actual encoding (for example, UTF-8 or UTF-16) is not important &mdash; we assume that the source code has already been converted into a sequence of Unicode code points according to the encoding it was in.
+
+The spec contains several grammars which we'll briefly describe next.
+
+### Lexical grammar
```
gahaas: I think the order of paragraphs is off here in this section. I think it should either be top down or bottom up.

Even though you refer to the Wikipedia article, you can still have a quick introduction here:
As most grammars the lexical grammar consists of a set of production rules which generate tokens for the syntactic grammar. For the lexical grammar, these production rules are called _goal symbols_.

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,320 @@
...
+> `RightBracePunctuator`
+> `RegularExpressionLiteral`
+
+As we see from the productions, it's possible that this produces the `RegularExpressionLiteral` input element, but producing `DivPunctuator` is not possible.
+
+Similarly, there is another goal symbol, `InputElementRegExpOrTemplateTail`, for contexts where `TemplateMiddle` and `TemplateTail` are permitted, in addition to `RegularExpressionLiteral`. And finally, `InputElementTemplateTail` is the goal symbol for contexts where only `TemplateMiddle` and `TemplateTail` are permitted but `RegularExpressionLiteral` is not permitted.
+
+We can imagine the syntactic grammar analyzer ("parser") calling the lexical grammar analyzer ("tokenizer" or "lexer"), passing the goal symbol as a parameter and asking for the next input element suitable for that goal symbol.
+
+### Other grammars
```
gahaas: `Other grammars` seems like a weird header. However, I think you can fix this by introducing the "table of contents" I proposed above.

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,320 @@
...
+> `VariableStatement_Yield` :
+> `var VariableDeclarationList_In_Yield;`
+>
+> `VariableStatement_Await` :
+> `var VariableDeclarationList_In_Await;`
+>
+> `VariableStatement_Yield_Await` :
+> `var VariableDeclarationList_In_Yield_Await;`
+
+Ultimately, we'll need to find out two things: 1) Where is it decided whether we're in the case with `_Await` or without `_Await`? 2) Where does it make a difference &mdash; where do the productions for `Something_Await` and `Something` (without `_Await`) diverge?
```
gahaas: You could make a list with two entries here. It would give more structure.

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,320 @@
...
+On the other hand, if we're inside a non-async function, the relevant production is:
+
+> [<code>FunctionDeclaration<sub>[Yield, Await, Default]</sub>](https://tc39.es/ecma262/#prod-FunctionDeclaration) :</code>
+> <code>function BindingIdentifier<sub>[?Yield, ?Await]</sub> ( FormalParameters<sub>[~Yield, ~Await]</sub> ) { FunctionBody<sub>[~Yield, ~Await]</sub> }</code>
+
+(`FunctionDeclaration` has another production, but it's not relevant for our code example.)
+
+To avoid combinatorial expansion, let's ignore the `Default` parameter which is not used in this particular production.
+
+The expanded form of the production is:
```
gahaas: Either there is a typo here, or I don't get it. I don't see a `FunctionBody_Await` here.

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Previous episodes
+
+In [part 2](/blog/understanding-ecmascript-part-2), we examined a simple grammar production and how its runtime semantics are defined.  In [the extra content](/blog/extra/understanding-ecmascript-part-2-extra), we also followed a long grammar production chain from `AssignmentExpression` to `MemberExpression`. In this episode, we'll go deeper in the definition of the ECMAScript (or JavaScript) language and its syntax.
+
+If you're not familiar with [context-free grammars](https://en.wikipedia.org/wiki/Context-free_grammar), now it's a good idea to check out the basics, since the spec uses context-free grammars to define the language.
+
+## ECMAScript grammars
+
+ECMAScript source text is a sequence of Unicode code points. Each Unicode code point is an integral value between `U+0000` and `U+10FFFF`. The actual encoding (for example, UTF-8 or UTF-16) is not important &mdash; we assume that the source code has already been converted into a sequence of Unicode code points according to the encoding it was in.
```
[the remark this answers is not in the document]
marjakh: I don't understand this comment :) I don't want to talk about "letters and symbols" since, what is a letter and a symbol? (A unicode code point!) Also I didn't say unicode is not important, I said the encoding is not important.

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,320 @@
...
+
+The spec contains several grammars which we'll briefly describe next.
+
+### Lexical grammar
+
+The [lexical grammar](https://tc39.es/ecma262/#sec-ecmascript-language-lexical-grammar) describes how Unicode code points are translated into a sequence of **input elements** (tokens, line terminators, comments, white space).
+
+There are several cases where the next token cannot be identified purely by looking at the Unicode code point stream, but we need to know where we are in the syntactic grammar. A classic example is `/`. To know whether it's a division or the start of the RegExp, we need to know which one is allowed in the syntactic context we're currently in.
```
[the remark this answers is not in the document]
marjakh: Right, I'll think about how to re-write this to be more clear about the point that it's not possible to tokenize in advance.

src/blog/understanding-ecmascript-part-3.md
marjakh: The next paragraph just after the productions explains it:

The important thing in this production is that FunctionBody is parameterized with [~Yield, ~Await], meaning that we always take the version without _Yield and without _Await, no matter what the parameters of FunctionDeclaration were.

marjakh: Rephrased:

In this production we always get FunctionBody (without _Yield and without _Await), since the FunctionBody in the non-expanded production is parameterized with [~Yield, ~Await].

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,322 @@
...
+  - ECMAScript
+description: 'Tutorial on reading the ECMAScript specification'
+tweet: ''
+---
+
+... where we dive deep in the syntax!
+
+## Previous episodes
+
+In [part 2](/blog/understanding-ecmascript-part-2), we examined a simple grammar production and how its runtime semantics are defined.  In [the extra content](/blog/extra/understanding-ecmascript-part-2-extra), we also followed a long grammar production chain from `AssignmentExpression` to `MemberExpression`. In this episode, we'll go deeper in the definition of the ECMAScript (or JavaScript) language and its syntax.
```
chicoxyzzy: extra space
```suggestion
In [part 2](/blog/understanding-ecmascript-part-2), we examined a simple grammar production and how its runtime semantics are defined. In [the extra content](/blog/extra/understanding-ecmascript-part-2-extra), we also followed a long grammar production chain from `AssignmentExpression` to `MemberExpression`. In this episode, we'll go deeper in the definition of the ECMAScript (or JavaScript) language and its syntax.
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,322 @@
...
+For example:
+```javascript
+const x = 10 / 5;
+//           ^ this is a DivPunctuator
+
+const r = /foo/;
+//        ^ this is the start of a RegularExpressionLiteral
+```
+
+A similar thing happens with templates &mdash; the interpretation of <code>{`</code> depends on the context we're in:
```
chicoxyzzy: Should be closing bracket?
```suggestion
A similar thing happens with templates &mdash; the interpretation of <code>}`</code> depends on the context we're in:
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,322 @@
...
+
+What do the subscripts (`[Yield, Await]`) and prefixes (`+` in `+In` and `?` in `?Async`) mean?
+
+The notation is explained in section [Grammar Notation](https://tc39.es/ecma262/#sec-grammar-notation).
+
+The subscripts are a shorthand for expressing a set of productions, for a set of left-hand side symbols, all at once. The left-hand side symbol has two parameters, so the "real" left-hand side symbols we're defining are `VariableStatement`, `VariableStatement_Yield`, `VariableStatement_Await` and `VariableStatement_Yield_Await`.
+
+Note that here the plain `VariableStatement` means "`VariableStatement` without `_Await` and without `_Yield`" and shoud not be confused with <code>VariableStatement<sub>[Yield, Await]</sub></code>.
+
+On the right-hand side of the production, we see the shorthand `+In`, meaning "use the version with `_In`", and `?Await`, meaning "use the version with `_Await` iff the left-hand side symbol has `_Await` (similarly with `?Yield`).
```
chicoxyzzy:
```suggestion
On the right-hand side of the production, we see the shorthand `+In`, meaning "use the version with `_In`", and `?Await`, meaning "use the version with `_Await` iff the left-hand side symbol has `_Await`" (similarly with `?Yield`).
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,322 @@
...
+> `function BindingIdentifier_Await ( FormalParameters_Await ) { FunctionBody }`
+
+> `FunctionDeclaration_Yield_Await :`
+> `function BindingIdentifier_Yield_Await ( FormalParameters_Yield_Await ) { FunctionBody }`
+
+In this production we always get `FunctionBody` (without `_Yield` and without `_Await`), since the `FunctionBody` in the non-expanded production is parameterized with `[~Yield, ~Await]`.
+
+Function name and formal parameters are treated differently: they get the parameters `_Await` and `_Yield` if the left-hand side symbol has them.
+
+To summarize: Async functions have a `FunctionBody_Await` and non-async functions have a `FunctionBody` (without `_Await`). You can think of the `_Await` parameter meaning "`await` is a keyword.
```
chicoxyzzy:
```suggestion
To summarize: Async functions have a `FunctionBody_Await` and non-async functions have a `FunctionBody` (without `_Await`). You can think of the `_Await` parameter meaning "`await` is a keyword".
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,322 @@
...
+
+In this case, the [static semantics for BindingIdentifier](https://tc39.es/ecma262/#sec-identifiers-static-semantics-early-errors) define the following syntax-directed rule:
+
+>`BindingIdentifier : await`
+>
+> It's a Syntax Error if this production has an <code><sub>[Await]</sub></code> parameter.
+
+Effectively, this forbids the `BindingIdentifier_Await : await` production.
+
+The spec explains that the reason for having this production but defining it as a Syntax Error by the static semantics is because of interference with automatic semicolon insertion. If the production was missing, automatic semicolon insertion might kick in and insert a semicolon into a program which is syntactically incorrect only because it uses `await` or `yield` as an identifier, changing the meaning of the program.
```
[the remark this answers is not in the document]
marjakh: Actually, do you think this paragraph above is correct? :) I'm now doubting it. The part "incorrect **only** because it uses...". Is it possible that the *only* error is that await was used as a keyword, and if the production was disallowed, ASI would turn the program into something correct?

The spec uses this as an example:

let
await 0;

and ASI turns it into 

let;
await 0;

But here it's not the case that the only error was "await" as an identifier: let await 0; is also incorrect as a let statement.

However, I don't want to use that example since it has "let" as an identifier too and I don't want to go there. So if what I'm saying is correct, that it's possible to have a program where the *only* error is await as an identifier, and ASI changing the meaning, that would be an easier to understand example.

However, I think it's maybe not possible...

let
await;

is not correct even after ASI so hmm.

(Sorry, totally rubberducking here, I think the answer will be clear to me once I post this comment.)

src/blog/understanding-ecmascript-part-3.md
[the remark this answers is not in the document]
gahaas: I guess my comment is mostly bike-shedding. You use terminology I'm not use to, and a level of detail that I don't consider particularly interesting.
Terminology: You write "Unicode code points" instead of "Unicode characters" or just "characters". After years of training I have an understanding what a character is. Why is a different notion used here?
Why is it interesting that Unicode code points are integrals value between `U+0000` and `U+10FFFF`? Do you want to say that there are a lot of them? 
Why is the encoding (UTF-8 or UTF-16) interesting? As @syg writes, maybe you can just link to a wikipedia article and be done with it."

syg:
@googlebot I consent.

src/blog/understanding-ecmascript-part-3.md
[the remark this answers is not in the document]
syg: Good call, I didn't think it through that closely.

Here is the original bug: https://github.com/tc39/test262/pull/956#issue-114912013

The problem was that previously, ASI was being made dependent on whether or not `await` was allowed as an identifier. So in async functions, where identifiers can't be named `await`, `let \n await 0;` ended up being parsed instead of raising an Early Error. This was deemed super confusing. 

*With* the static semantics rule, the error is that `await` is used as an identifier. *Without* the static semantics rule, it'll just a `let; await 0;` which is legal syntax. So I think that sentence is inverted. Maybe something like:

"If the production was missing, automatic semicolon insertion will kick in and transform the program into the legal program `let; await 0;` in contexts where `await` is disallowed as an identifier, like inside async functions. The rule aims to prevent this transformation, which TC39 judged to be egregiously confusing."

src/blog/understanding-ecmascript-part-3.md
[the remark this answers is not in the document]
marjakh: As discussed, rewrote this:

The spec defines ECMAScript source text as a sequence of Unicode characters. This means that variable names are not limited to ASCII characters but can also include other Unicode characters, such as emojis. The spec doesn't talk about the actual encoding (for example, UTF-8 or UTF-16). We assume that the source code has already been converted into a sequence of Unicode characters according to the encoding it was in.

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,320 @@
...
+
+In [part 2](/blog/understanding-ecmascript-part-2), we examined a simple grammar production and how its runtime semantics are defined.  In [the extra content](/blog/extra/understanding-ecmascript-part-2-extra), we also followed a long grammar production chain from `AssignmentExpression` to `MemberExpression`. In this episode, we'll go deeper in the definition of the ECMAScript (or JavaScript) language and its syntax.
+
+If you're not familiar with [context-free grammars](https://en.wikipedia.org/wiki/Context-free_grammar), now it's a good idea to check out the basics, since the spec uses context-free grammars to define the language.
+
+## ECMAScript grammars
+
+ECMAScript source text is a sequence of Unicode code points. Each Unicode code point is an integral value between `U+0000` and `U+10FFFF`. The actual encoding (for example, UTF-8 or UTF-16) is not important &mdash; we assume that the source code has already been converted into a sequence of Unicode code points according to the encoding it was in.
+
+The spec contains several grammars which we'll briefly describe next.
```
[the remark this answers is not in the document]
marjakh: Made this chapter an introductory one, with one paragraph about each grammar.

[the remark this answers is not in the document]
marjakh: Still no table of contents though, sorry ;) Maybe I'll regret this decision later? But I'd like to keep the blog post informal and the table of contents somehow feels like contradicting that.

src/blog/understanding-ecmascript-part-3.md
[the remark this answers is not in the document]
marjakh: Restructured. New text:

It's not possible to tokenize ECMAScript source code in advance, which makes defining the lexical grammar slightly more complicated.

For example, we cannot determine whether `/` is the division operator or the start of a RegExp without looking at the larger context it occurs:

src/blog/understanding-ecmascript-part-3.md
marjakh: Added:

Each grammar is defined as a context-free grammar, consisting of a set of production rules.

----

(The suggestion " For the lexical grammar, these production rules are called goal symbols." is not true; goal symbol is not the a production; each context free grammar has one "main" symbol, such as "Program", which is the goal symbol.)

src/blog/understanding-ecmascript-part-3.md
marjakh: I moved them into the introduction chapter at the top. The long explanations of the lexical & syntactic grammars comes after that.

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,320 @@
...
+tweet: ''
+---
+
+... where we dive deep in the syntax!
+
+## Previous episodes
+
+In [part 2](/blog/understanding-ecmascript-part-2), we examined a simple grammar production and how its runtime semantics are defined.  In [the extra content](/blog/extra/understanding-ecmascript-part-2-extra), we also followed a long grammar production chain from `AssignmentExpression` to `MemberExpression`. In this episode, we'll go deeper in the definition of the ECMAScript (or JavaScript) language and its syntax.
+
+If you're not familiar with [context-free grammars](https://en.wikipedia.org/wiki/Context-free_grammar), now it's a good idea to check out the basics, since the spec uses context-free grammars to define the language.
```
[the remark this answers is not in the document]
marjakh: Offline discussion: I use the word "context", e.g., to say that "/" is division in some contexts and regexp in others. This might be confusing because of the overloaded word "context". I tried to think of a better word but couldn't find any... I'll prob add a note that "context" here doesn't mean the same as "context" in the "context-free grammar" sense.

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,322 @@
...
+
+The [syntactic grammar](https://tc39.es/ecma262/#sec-syntactic-grammar) describes how syntactically correct programs are composed of tokens.
+
+The notation used for different grammars differs slightly. For example, the syntactic grammar uses `Symbol :` whereas the lexical grammar and the RegExp grammar use `Symbol ::` and the numeric string grammar uses `Symbol :::`.
+
+For the rest of this episode, we'll focus on the syntactic grammar.
+
+## Example: Allowing legacy identifiers
+
+In some contexts, `await` and `yield` are allowed identifiers. Finding out when exactly they are allowed can be a bit involved, so let's dive right in!
```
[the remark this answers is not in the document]
marjakh: Added explanation about the history here.

Example: Allowing legacy identifiers

Introducing new keywords to the grammar is a possibly breaking change — what if existing code already uses the keywords as identifiers?

For example, before await was a keyword, someone might have written the following code:

function old() {
  var await;
}
The ECMAScript grammar carefully added the await keyword in such a way that this code will continue to work. Inside async functions, await is a keyword, so this doesn't work:

async function modern() {
  var await; // Syntax error
}
Allowing yield as an identifier in non-generators and disallowing it in generators works similarly; let's focus on the await case for the rest of the post.

Understanding how await is allowed as an identifier requires understanding ECMAScript-specific syntactic grammar notation. Let's dive right in!

src/blog/understanding-ecmascript-part-3.md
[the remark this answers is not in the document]
marjakh: Rewrote this part.

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,322 @@
...
+```
+
+A similar thing happens with templates &mdash; the interpretation of <code>}`</code> depends on the context we're in:
+
+```javascript
+const what1 = 'temp';
+const what2 = 'late';
+const t = `I am a ${ what1 + what2 }`;
+// `I am a ${ is TemplateHead
+// }` is TemplateTail
```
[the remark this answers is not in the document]
marjakh: Took this comments out of the source code, hopefully they're easier to read now.

syg:
Sorry about the CLA thing -- I even changed my primary email to my chromium one. @marjakh I think you'll again have to rebase and get rid of my committer info to get it to pass.

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,362 @@
...
+  - ECMAScript
+description: 'Tutorial on reading the ECMAScript specification'
+tweet: ''
+---
+
+... where we dive deep in the syntax!
+
+## Previous episodes
+
+In [part 1](/blog/understanding-ecmascript-part-1), we read through a simple method — `Object.prototype.hasOwnProperty` — and **abstract operations** it invokes. We familiarized ourselves with the shorthands `?` and `!` related to error handling. We encountered **language types**, **specification types**, **internal slots**, and **internal methods**.
```
syg:
```suggestion
In [part 1](/blog/understanding-ecmascript-part-1), we read through a simple method — `Object.prototype.hasOwnProperty` — and the **abstract operations** it invokes. We familiarized ourselves with the shorthands `?` and `!` related to error handling. We encountered **language types**, **specification types**, **internal slots**, and **internal methods**.
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,362 @@
...
+
+In [part 2](/blog/understanding-ecmascript-part-2), we examined a simple grammar production and how its runtime semantics are defined. In [the extra content](/blog/extra/understanding-ecmascript-part-2-extra), we also followed a long grammar production chain from `AssignmentExpression` to `MemberExpression`. In this episode, we'll go deeper in the definition of the ECMAScript language and its syntax.
+
+If you're not familiar with [context-free grammars](https://en.wikipedia.org/wiki/Context-free_grammar), now it's a good time to check out the basics, since the spec uses context-free grammars to define the language.
```
syg:
```suggestion
If you're not familiar with [context-free grammars](https://en.wikipedia.org/wiki/Context-free_grammar), now is a good time to check out the basics, since the spec uses context-free grammars to define the language.
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,362 @@
...
+
+Next we'll look into the lexical grammar and the syntactic grammar in more detail.
+
+## Lexical grammar
+
+The spec defines ECMAScript source text as a sequence of Unicode code points. For example, variable names are not limited to ASCII characters but can also include other Unicode characters, such as emojis. The spec doesn't talk about the actual encoding (for example, UTF-8 or UTF-16). It assumes that the source code has already been converted into a sequence of Unicode code points according to the encoding it was in.
+
+It's not possible to tokenize ECMAScript source code in advance, which makes defining the lexical grammar slightly more complicated.
+
+For example, we cannot determine whether `/` is the division operator or the start of a RegExp without looking at the larger context it occurs:
```
syg:
```suggestion
For example, we cannot determine whether `/` is the division operator or the start of a RegExp without looking at the larger context it occurs in:
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,362 @@
...
+
+Here `/` is a DivPunctuator.
+
+```js
+const r = /foo/;
+```
+
+Here the first `/` is the start of a RegularExpressionLiteral.
+
+Templates introduce a similar ambiguity &mdash; the interpretation of <code>}`</code> depends on the context it occurs:
```
syg:
```suggestion
Templates introduce a similar ambiguity &mdash; the interpretation of <code>}`</code> depends on the context it occurs in:
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,362 @@
...
+### Productions and shorthands
+
+Let's look at how the productions for `VariableStatement` are defined. At the first glance, the grammar can look a bit scary:
+
+> [<code>VariableStatement<sub>[Yield, Await]</sub> :</code>](https://tc39.es/ecma262/#prod-VariableStatement)
+> <code>var VariableDeclarationList<sub>[+In, ?Yield, ?Await]</sub>;</code>
+
+What do the subscripts (`[Yield, Await]`) and prefixes (`+` in `+In` and `?` in `?Async`) mean?
+
+The notation is explained in section [Grammar Notation](https://tc39.es/ecma262/#sec-grammar-notation).
```
syg:
```suggestion
The notation is explained in the section [Grammar Notation](https://tc39.es/ecma262/#sec-grammar-notation).
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,362 @@
...
+
+The subscripts are a shorthand for expressing a set of productions, for a set of left-hand side symbols, all at once. The left-hand side symbol has two parameters, so the "real" left-hand side symbols we're defining are `VariableStatement`, `VariableStatement_Yield`, `VariableStatement_Await` and `VariableStatement_Yield_Await`.
```
syg:
```suggestion
The subscripts are a shorthand for expressing a set of productions, for a set of left-hand side symbols, all at once. The left-hand side symbol has two parameters, which expands into four "real" left-hand side symbols: `VariableStatement`, `VariableStatement_Yield`, `VariableStatement_Await`, and `VariableStatement_Yield_Await`.
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,362 @@
...
+
+Note that here the plain `VariableStatement` means "`VariableStatement` without `_Await` and `_Yield`". It should not be confused with <code>VariableStatement<sub>[Yield, Await]</sub></code>.
+
+On the right-hand side of the production, we see the shorthand `+In`, meaning "use the version with `_In`", and `?Await`, meaning "use the version with `_Await` if and only if the left-hand side symbol has `_Await`" (similarly with `?Yield`).
+
+(The third shorthand, `~Foo`, meaning "use the version without `_Foo`", is not used in this production.)
```
syg:
```suggestion
The third shorthand, `~Foo`, meaning "use the version without `_Foo`", is not used in this production.
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,362 @@
...
+
+Function name and formal parameters are treated differently: they get the parameters `_Await` and `_Yield` if the left-hand side symbol has them.
+
+To summarize: Async functions have a `FunctionBody_Await` and non-async functions have a `FunctionBody` (without `_Await`). Since we're talking about non-generator functions, both our async example function and our non-async example function are parameterized without `_Yield`.
+
+Maybe it's hard to remember which one is `FunctionBody` and which `FunctionBody_Await`. Is `FunctionBody_Await` for a function where `await` is an identifier, or for a function where `await` is a keyword?
+
+You can think of the `_Await` parameter meaning "`await` is a keyword". This approach is also future proof. Imagine a new keyword, `blob` being added, but only inside "blobby" functions. Non-blobby non-async non-generators would still have `FunctionBody` (without `_Await`, `_Yield` or `_Blob`), exactly like they have now. Blobby functions would have a `FunctionBody_Blob`, async blobby functions would have `FunctionBody_Await_Blob` and so on.
```
syg: Well, future proof up to still having to add ~Blob to FunctionBody, I guess?

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,362 @@
...
+> `yield`
+> `await`
+
+(We're omitting the productions for `BindingIdentifier_Yield` and `BindingIdentifier_Yield_Await` which are not needed in our example.)
+
+This looks like `await` and `yield` would be always allowed as identifiers. What's up with that? Is the whole blog post for nothing?
+
+### Statics semantics to the rescue
+
+Turns out **static semantics** are needed for forbidding `await` as an identifier inside async functions.
```
syg:
```suggestion
It turns out that **static semantics** are needed for forbidding `await` as an identifier inside async functions.
```

src/blog/understanding-ecmascript-part-3.md
```diff
@@ -0,0 +1,362 @@
...
+> It is a Syntax Error if this production has an <code><sub>[Await]</sub></code> parameter and `StringValue` of `Identifier` is `"await"`.
+
+This might be confusing at first. `Identifier` is defined like this:
+
+> [`Identifier :`](https://tc39.es/ecma262/#prod-Identifier)
+> `IdentifierName` but not `ReservedWord`
+
+`await` is a `ReservedWord`, so how can an `Identifier` ever be `await`?
+
+Turns out, `Identifier` cannot be `await`, but it can be something else whose `StringValue` is `"await"` &mdash; a different representation of the character sequence `await`.
```
syg:
```suggestion
As it turns out, `Identifier` cannot be `await`, but it can be something else whose `StringValue` is `"await"` &mdash; a different representation of the character sequence `await`.
```

src/blog/understanding-ecmascript-part-3.md
marjakh: Right, I meant that the expanded forms won't change.

marjakh: Added: We'd still need to add the Blob subscript to the rules, but the expanded forms of FunctionBody for already existing functions stay the same.

PR #344, opened 2020-03-10
Pointer Compression blog post

santiaboy:
Pointer Compression deep dive blog post


santiaboy:
There is a todo in the markdown file, since I am not sure if we can do that (and if so, how to do it).
The rest of the file is ready for publishing

src/blog/pointer-compression.md
```diff
@@ -0,0 +1,478 @@
...
+```
+## Results
+Let’s take a look at Pointer Compression’s final numbers!
+
+We use a set of web pages for measuring our memory and performance metrics that reflect popular real-world websites. In them, we observed that Pointer Compression reduces V8 heap size up to 43%!
+
+In turn, it reduces Chrome’s renderer process memory by 20% on Desktop and 10% in Mobile. This difference appears due to V8 being a lower percentage of the renderer process memory on Mobile.
+
+<figure>
+  <img src="/_img/pointer-compression/V8-heap-memory.svg" width="600" height="371" alt="Memory savings when browsing in Windows 10" loading="lazy">
```
mathiasbynens: `V8-heap-memory.svg` → `v8-heap-memory.svg`

santiaboy: Done!

RReverser:
Please fix issues marked by markdownlint - you can see them either in Travis CI logs or by running `npm run lint` locally.

RReverser:
@santiaboy Ping.

santiaboy:
There are still some svgs to be updated and minified, but I had a question about a lint failure.
We want to show code inside a table but it is complaining saying "no-inline-html Inline HTML [Element: br]". How do we solve this?

RReverser:
@santiaboy Hmm yeah for that one you'd probably want to just disable the lint around the table using

```
<!-- markdownlint-disable no-inline-html -->
...
<!-- markdownlint-enable no-inline-html -->
```

RReverser:
Update: Turns out we already use a plugin that allows for multiline cells, so I enabled the corresponding option, transformed first table in the post to use it as an example, and fixed style issue for code blocks in a table.

Transforming the rest following this one as an example should be fairly straightforward (even though I'm not saying "easy", because syntax is a bit weird and not supported by standard table generators unfortunately).

Let me know if you run into any issues!

RReverser:
Btw, I suggest to use ```` ```asm ```` on other code blocks in the post as well - currently they don't seem to use any highlighting even outside the tables, e.g.

image

src/blog/pointer-compression.md
```diff
@@ -0,0 +1,514 @@
...
+movlsxlq rax, rax // sign extend
+```
+
+Once we fixed that to sign extend the value loaded from memory directly:
+
+```
+movlsxlq rax, <mem>
+```
+
+swe got yet another 2% improvement.
```
RReverser: Looks like a typo. Not sure but was this supposed to be

```suggestion
so got yet another 2% improvement.
```

?

src/blog/pointer-compression.md
```diff
@@ -0,0 +1,514 @@
...
+
+There are cases where the decompression is not actually necessary. For example, if a compressed value is loaded from somewhere only to be then stored to a new location.
+
+In order to optimize unnecessary operations, we implemented a new “Decompression Elimination” phase in TurboFan. Its job is to eliminate decompressions directly followed by compressions. Since these nodes might not be directly next to each other it also tries to propagate decompressions through the graph, with the hope of encountering a compress down the line and eliminating them both. This gave us a 2% improvement of Octane’ score.
+
+#### Bump (3), +2%
+
+While we were looking at the generated code, we noticed that the decompression of a value that had just been loaded produced code that was a bit too verbose:
+
+```
```
RReverser:
```suggestion
```asm
```

src/blog/pointer-compression.md
```diff
@@ -0,0 +1,514 @@
...
+movl rax, <mem>   // load
+movlsxlq rax, rax // sign extend
+```
+
+Once we fixed that to sign extend the value loaded from memory directly:
+
+```
```
RReverser:
```suggestion
```asm
```

santiaboy:
I updated the tabled with the new version.
There is still the matter of the first graph that I am updating, but I wanted to push since I couldn't get rid of the npm lint errors. It is throwing for some reason.

src/blog/pointer-compression.md
```diff
@@ -206,24 +207,40 @@ Let’s take a look at the x64 assembly.
 |               | andl r10,0x1            | jz done                      \
 |               | negq r10                | addq r11,r13                 \
 |               | andq r10,r13            | done:                        \
-|               | addq r11,r10            | ```                          \
-|               | ```                     |                              |
+|               | addq r11,r10            |  ```                         \
+|               | ```                     |  <br>                        |
```
RReverser: Hmm what happened here / why does this need `<br>` again?

src/blog/pointer-compression.md
```diff
@@ -322,12 +339,25 @@ Also, since we don’t care about sign extending the Smi anymore, this change al
...
 :::table-wrapper
-| Decompression | Branchless              | Branchful                    |
-|---------------|-------------------------|------------------------------|
-| Code          | `movsxlq r11,[...]`<br>`testb r11,0x1`<br>`jz done`<br>`addq r11,r13`<br>`done:` | `movl r11,[rax+0x13]`<br>`addq r11,r13`<br><br><br><br> |
-| Summary       | 13 bytes<br>3 or 4 instructions executed<br>1 branch | 7 bytes<br>2 instructions executed<br>no branches |
+<!-- markdownlint-disable no-space-in-code -->
+| Decompression | Branchful                    | Smi-corrupting               |
+|---------------|------------------------------|------------------------------|
+| Code          | ```asm                       | ```asm                       \
+|               |   movsxlq r11,[...]          |   movl r11,[rax+0x13]        \
```
RReverser: There seems to be extra unnecessary spacing in the beginning of each code block line here, unlike in others.

src/blog/pointer-compression.md
RReverser: (same for other tables - we shouldn't need it anymore)

santiaboy: br is there so that both code blocks start at the top

src/blog/pointer-compression.md
santiaboy: Done

src/blog/pointer-compression.md
RReverser: Hmm I'm conflicted about it, but I guess fine. Just as another option, would leaving a newline inside the code block help? (then both code blocks would have the same size, which might be even better visually for showing the difference)

santiaboy: That's what I tried first, but it does not work (the new line is removed)

RReverser: Hmm it works on my side. To be clear - I'm talking about an extra newline before the closing ```` ``` ````.

santiaboy:
Updated the graph. If we can fix the linter issue we are set.
It is happening since we introduced the multiline code in the table. It is throwing with "Error: Property 'range' of onError parameter is incorrect."

src/blog/pointer-compression.md
RReverser: @santiaboy Ping - what do you think?

RReverser: Nvm, this one doesn't work.

src/blog/pointer-compression.md
```diff
@@ -0,0 +1,544 @@
...
+|               | movl r10,r11            | testb r11,0x1                \
+|               | andl r10,0x1            | jz done                      \
+|               | negq r10                | addq r11,r13                 \
+|               | andq r10,r13            | done:                        \
+|               | addq r11,r10            |  ```                         \
+|               | ```                     |  <br>                        |
+| Summary       | 20 bytes                | 13 bytes                     \
+|               | 6 instructions executed | 3 or 4 instructions executed \
+|               | no branches             | 1 branch                     \
+|               | 1 additional register   | <br>                         |
```
RReverser: Wait, so you want all the table cells to be aligned to the top, not just cells with code?

If so, we could add a table class (similarly to `table-wrapper`) that would do just that without any `<br>`s, but it makes table style inconsistent across the blog, as all the other ones are aligned by the middle.

Is there any particular reason you want that alignment?

@mathiasbynens do you have thoughts on this as well maybe?

mathiasbynens: No opinion, other than the usual "let's keep it as simple as possible". I trust you'll make the right decision here 👍

santiaboy: Yes, aligning to the top is what we want. Both sides start with loading and then they differ. It is easier to see the difference(e.g in code size) if they are both aligned at the top

RReverser: > No opinion, other than the usual "let's keep it as simple as possible". I trust you'll make the right decision here 👍

I just meant an opinion on concistency. But if you don't have one, that's fine.

RReverser: @santiaboy I see. If the goal is to show clear matching between results, I wonder if it would make more sense to use proper cells (and rowspan for summary here) then?

Here's what it would look like:
image

You can check table in `v8-release-80.md` for markdown code example.

mathiasbynens: Strong +1 to this suggestion. This seems perfect!

mathiasbynens: Btw, this is the first time we're adding `<pre>` within tables on v8.dev, so let's make sure this works well on narrow viewports etc. before landing.

santiaboy: Yes, that looks perfect.

RReverser: > within tables on v8.dev, so let's make sure this works well on narrow viewports etc. before landing.

Yeah I double-checked that already, table-wrapper seems to work fine here too and just introduces horizontal scroll when necessary.

src/blog/pointer-compression.md
RReverser: Found how this one worked - you just need to use `| \` in the end of the line with spaces instead of `\`.

@santiaboy can you please try again?

RReverser:
Added a custom hacky way (or, in fancy words, a CSS4 feature polyfill!) to render table cells with just code inside a bit differently.

Before:
image


After:
image

mathiasbynens:
@rreverser Looking slick!

RReverser:
@mathiasbynens I was wondering whether you'll say that or "oh nooo this is way too hacky" 😅

mathiasbynens:
> @mathiasbynens I was wondering whether you'll say that or "oh nooo this is way too hacky" 😅

Not at all :) I’m totally down with this. As long as we can hide the “hackiness” in the build process as opposed to paying the cost on the client-side, I consider it a win!

src/blog/pointer-compression.md
```diff
@@ -0,0 +1,537 @@
...
+
+Back in 2014 Chrome switched from being a 32-bit process to a 64-bit process. This gave Chrome better [security, stability and performance](https://blog.chromium.org/2014/08/64-bits-of-awesome-64-bit-windows_26.html), but it came at a memory cost since each pointer now occupies eight bytes instead of four. We took on the challenge to reduce this overhead in V8 to try and get back as many wasted 4 bytes as possible.
+
+Before diving into the implementation, we need to know where we are standing to correctly assess the situation. To measure our memory and performance we use a set of [web pages](https://v8.dev/blog/optimizing-v8-memory) that reflect popular real-world websites. The data showed that V8 contributes up to 60% of Chrome’s [renderer process](https://www.chromium.org/developers/design-documents/multi-process-architecture) memory consumption on desktop, with an average of 40%.
+
+<figure>
+  <img src="/_img/pointer-compression/memory-chrome.svg" width="600" height="371" alt="V8 memory consumption percentage in Chrome’s renderer memory" loading="lazy">
```
RReverser: Here and on other images: set `alt` to empty with `alt=""`. When we use `figcaption`, there is no point in duplicating same text inside the `alt`, because it will only worsen UX for users on low-speed connections or impaired users, who will now see or hear same text twice.

src/blog/pointer-compression.md
```diff
@@ -0,0 +1,537 @@
...
+  <img src="/_img/pointer-compression/perf-octane-2.svg" width="859" height="178" alt="Second round of Octane's improvements" loading="lazy">
+  <figcaption>Second round of Octane's improvements</figcaption>
+</figure>
+
+#### Bump (5), +0.5%
+
+While implementing the Decompression Elimination in TurboFan we learned a lot. The explicit Decompression/Compression node approach had the following properties:
+
+Pros:
+    - Explicitness of such operations allowed us to optimize unnecessary decompressions by doing canonical pattern matching of sub-graphs.
```
RReverser: These lists don't render as actual lists, you might want to add a newline before the first item.

src/blog/pointer-compression.md
santiaboy: Done.

src/blog/pointer-compression.md
```diff
@@ -0,0 +1,542 @@
...
+</pre>
+
+You may notice that unlike 32-bit architectures, on 64-bit architectures V8 can use 32 bits for the Smi value payload. The benefits of 32-bit Smis will be discussed in the following sections.
+
+You may notice that unlike 32-bit architectures, on 64-bit architectures V8 can use 32 bits for the Smi value payload. The implications of 32-bit Smis on pointer compression will be discussed in the following sections.
+
+## Compressed tagged values and new heap layout
+
+With Pointer Compression, our goal is to somehow fit both kinds of tagged values into 32 bits on 64-bit architectures. We can fit pointers into 32 bits by:
+    - making sure all V8 objects are allocated within a 4GB memory range
```
RReverser: @santiaboy Looks like this one got missed and still doesn't render like list (btw, the indentation is also a bit excessive).

santiaboy: Done.
The indentation was due to npm run lint. I was originally using " *" and I guess that confused it because it sugested using "-" and 4 spaces.

Issue #281, opened 2019-11-11
Fix styling for wide headings on narrow viewports


mathiasbynens:
e.g. https://v8.dev/features/string-replaceall or https://v8.dev/features/symbol-description

![](https://user-images.githubusercontent.com/81942/68602299-f3148f00-045a-11ea-938e-8ffc7861f765.png)

(Reported via Twitter: https://twitter.com/saschaplas/status/1193918922682032128)

RReverser:
Reopening to track issue on iOS Safari: https://twitter.com/tomayac/status/1234565673277370368

PR #345, opened 2020-03-12
Fix styling for wide content on Safari

tomayac:
Re-fixes https://github.com/v8/v8.dev/issues/281


src/_css/main.css
```diff
@@ -251,6 +251,8 @@ code {
   tab-size: 2;
   background: rgba(0, 0, 0, 0.07);
   overflow-wrap: break-word;
+  /* The below is still required for Safari */
+  word-wrap: break-word;
```
tomayac: It shouldn't be necessary [in theory](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#wikiArticle:~:text=The%20property%20was%20originally%20a%20nonstandard,overflow%2Dwrap%2C%20with%20word%2Dwrap%20being%20an%20alias.), but is in practice.

mathiasbynens:
Just like that huh? Thank you @tomayac!

PR #348, opened 2020-03-27
Link to the internal committer permissions process

backes:
We often forget that there is a streamlined process for v8 team members. Link to the internal process from the external page to increase visibility.

### DIFF
--- a/src/docs/become-committer.md
+++ b/src/docs/become-committer.md
@@ -2,6 +2,9 @@
 title: 'Becoming a committer'
 description: 'How does one become a V8 committer? This document explains.'
 ---
+
+*Note to googlers: There is a [slightly different approach for v8 team members](http://go/v8/setup_permissions.md).*
+
 Technically, committers are people who have write access to the V8 Git repository. Committers can submit their own patches or patches from others.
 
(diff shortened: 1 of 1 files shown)


backes:
PTAL, especially about linking an internal document externally. Via go link, I think it's OK. Chromium also does that, e.g. at https://www.chromium.org/developers/how-tos/get-the-code.

src/docs/become-committer.md
```diff
@@ -2,6 +2,9 @@
 title: 'Becoming a committer'
 description: 'How does one become a V8 committer? This document explains.'
 ---
+
+*Note to googlers: There is a [slightly different approach for v8 team members](http://go/v8/setup_permissions.md).*
```
mathiasbynens:
```suggestion
*Note to Googlers: There is a [slightly different approach for V8 team members](http://go/v8/setup_permissions.md).*
```

mathiasbynens:
```suggestion

*Note to Googlers: There is a [slightly different approach for V8 team members](http://go/v8/setup_permissions.md).*
```

backes:
Thanks, done.

src/docs/become-committer.md
```diff
@@ -2,6 +2,9 @@
 title: 'Becoming a committer'
 description: 'How does one become a V8 committer? This document explains.'
 ---
+
```
mathiasbynens: Nit: we don't usually have this newline here.

backes: Ack, removed.

PR #351, opened 2020-03-30
Add missing figcaption to a blog post

RReverser:
Looks like this was an accidental omission as we use figcaption on all other images in the blog, but sending as a PR to double-check with the author.


marjakh:
Yes, this was accidental! Thanks for fixing

PR #352, opened 2020-03-30
Use figcaption instead of title

RReverser:
This makes caption visible to all users and is more consistent with the rest of the blog.


RReverser:
I see. That seems unusual, but I guess fine in this case. Just a shame that e.g. mobile users won't see same funny text, which `figcaption` would solve.

mathiasbynens:
> Just a shame that e.g. mobile users won't see same funny text

They will if they interact with the image in the way that displays the `title` text. (E.g. on Chrome for Android, tap-and-hold on the image.)

RReverser:
> They will if they interact with the image in the way that displays the title text.

Sure, but for that they need to know that image offers some interaction. I'd imaging most users don't tap-and-hold on images in blogs.

mathiasbynens:
And that's fine for non-crucial captions like these.

RReverser:
I don't know... I think we should be at least marking these images somehow if we don't offer a text that is definitely visible. But not sure which alternative way would be good then. Let's keep this open for now in case we can think of something better.

LeszekSwirski:
Yeah, xkcd-style easter-egg was what I was going for here.

RReverser:
TBH until you all told me, I didn't know XKCD has hover titles either 😅

LeszekSwirski:
whole-new-world

RReverser:
I guess I'll just close it, I don't have that strong opinion.

PR #353, opened 2020-03-31
Headings Fix

santiaboy:
I am not sure why I have 49 commits in this pull request, but the end result is removing one # from two headings :)


backes:
You should be able to rebase your branch on v8:master and force-push to eliminate all the already-landed CLs.

santiaboy:
Rebased and forced push!
PTAL

PR #354, opened 2020-03-31
Add support for images in regular Markdown

RReverser:
One issue I've noticed contributors struggling with frequently is our custom embedding of images in a consistent format. Indeed, sometimes image descriptions go into `alt`, sometimes into `figcaption`, and sometimes into both.

Also, currently contributors need to either manually set image dimensions or use the provided script in the repo which fetches sizes and embeds them in Markdown.

While these things can be caught during a review, they still require manual work from both contributors and maintainers and sometimes can still slip by, e.g. with a script I used for this PR I've caught and fixed few missed instances like #351, https://github.com/v8/v8.dev/commit/a980a5a185d4cc727bac26b77de550b5e25fc7a5 and https://github.com/v8/v8.dev/commit/a2f284d079f885ce09094df88780c351c8feed07.

As such, I've decided to continue a spirit of automation started in https://github.com/v8/v8.dev/pull/344#issuecomment-604585411 and made a custom plugin that, in combination with added `markdown-it-implicit-figures` (or rather, its fork with a fix I had to make), allows to render Markdown images in our standard style.

I've verified that we didn't use Markdown images anywhere else so far, so this is not a breaking change, and now allows to reduce e.g.

```html
<figure>
  <img src="/_img/array-sort/runs-stack.svg" width="770" height="427" alt="" loading="lazy">
  <figcaption>Runs stack before and after merging <code>A</code> with <code>B</code></figcaption>
</figure>
```

to just

```markdown
![Runs stack before and after merging `A` with `B`](/_img/array-sort/runs-stack.svg)
```

The custom plugin will take care of inlining image sizes, marking images as lazy-loaded, and adding `srcset` when a `@2x` version of the image is present, and then `markdown-it-implicit-figures` will wrap the image into a `<figure>` and extract rich description into a `<figcaption>` if present.

In order to make sure that we handle all common cases and showcase the difference, I also wrote a one-off script and converted all existing repeititions of such `<figure>`s into Markdown as well. This is not strictly necessary, but I believe useful as an example for future contributors and ensures that we can easily update embedding for all such images should we want to, from a single place.

There are few cases of images that were deliberately skipped by script, and these include either `img` tags with custom `width` and `height` that don't match the image itself (presumably for scaling, but I've decided it's best not to touch them as they need to be checked on one-by-one basis) as well as a single post that uses `title` instead of `alt` or `figcaption` (#352).

---

All in all, I believe this should save significant time for both sides and make it easier to embed images in new posts.


md-embed-image.js
```diff
@@ -0,0 +1,29 @@
...
+      // Skip non-image tokens.
+      if (t.type !== 'image') continue;
+      let imgSrc = t.attrGet('src');
+      // We only embed self-hosted images.
+      if (imgSrc.startsWith('/_img/')) {
+        let { width, height } = imageSize('src' + imgSrc);
+        // Lazify image and embed its sizes to avoid layout jump.
+        t.attrs.push(['width', width], ['height', height], ['loading', 'lazy']);
+        // Check if `file@2x.ext` exists for `file.ext`.
+        let imgSrc2x = imgSrc.replace(/\.[^.]*$/, '@2x$&');
```
mathiasbynens:
```suggestion
        const imgSrc2x = imgSrc.replace(/\.[^.]*$/, '@2x$&');
```

md-embed-image.js
```diff
@@ -0,0 +1,29 @@
...
+module.exports = md => {
+  // Add a post-process rule for inline items.
+  md.inline.ruler2.push('embed_image', state => {
+    for (let t of state.tokens) {
+      // Skip non-image tokens.
+      if (t.type !== 'image') continue;
+      let imgSrc = t.attrGet('src');
+      // We only embed self-hosted images.
+      if (imgSrc.startsWith('/_img/')) {
+        let { width, height } = imageSize('src' + imgSrc);
```
mathiasbynens:
```suggestion
        const { width, height } = imageSize('src' + imgSrc);
```

md-embed-image.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const imageSize = require('image-size');
+const { existsSync } = require('fs');
+
+module.exports = md => {
+  // Add a post-process rule for inline items.
+  md.inline.ruler2.push('embed_image', state => {
+    for (let t of state.tokens) {
```
mathiasbynens:
```suggestion
    for (const t of state.tokens) {
```

RReverser:
> Can we also simplify the `img-video-dimensions` script now? (I guess we still want to keep it for `<video>`?)

Yeah, I want to tackle videos separately.

As for the script, it's still useful for cases where people use custom HTML e.g. for classes on images like `no-darkening`.

I might change that too in a separate PR by using https://www.npmjs.com/package/markdown-it-attrs, but for now I wanted to keep things simple and the original script is the easiest way to put dimensions on such images.

md-embed-image.js
RReverser: 🙄 okay, just for you :)

robdodson:
Inlining the image sizes seems really cool. @RReverser and I were chatting about this a bit today and I did want to mention that using an empty `alt=""` means that the images themselves will not be announced by assistive technology. In other words, it will read the figcaption but will not tell the user that an image preceded it.

Here you can see that using an empty alt hides the image from the accessibility tree:

image

Here's a quick demo using VoiceOver. Note the image is bypassed:

Kapture 2020-04-06 at 10 25 40

Since the images on v8.dev are often data visualizations, it might be ok that this happens. It would be hard to accurately describe them using alt text. But I did want to flag it so y'all are aware of the tradeoffs.

Ingvar also shared [this StackOverflow comment](https://stackoverflow.com/questions/58447538/accessibility-difference-between-img-alt-and-figcaption/58468470#58468470) from a blind user who said that they found it annoying when folks duplicated the alt and the figcaption, so he suggested leaving the alt text out. I left a comment asking if he found it weird that the caption would be announced without the image.

PR #355, opened 2020-03-31
Change alt to figcaption in Liftoff post

RReverser:
As in other found cases, this likely was an accident, but tagging author to confirm just in case.


backes:
Both works for me. I wouldn't say that is was an accident, the text just doesn't add anything here. But putting it as a caption is also fine.

PR #356, opened 2020-03-31
Convert few more images with attributes

RReverser:
Follow-up to #354. 

Unfortunately, overriding width or height via markdown-it-attrs syntax is not yet possible due to conflicts between plugins and undocumented APIs, but at least we can convert few more images meanwhile.


mathiasbynens:
LGTM assuming this results in `title="..."` :)

RReverser:
> LGTM assuming this results in title="..." :)

If you're referring to the first port, yeah, that's standard Markdown syntax for `title` on images and links apparently.

PR #364, opened 2020-04-02
Fix figure markup

tomayac:
Without the paragraph, this renders as `` instead of `<figure>`


RReverser:
Thanks!

mathiasbynens:
> Without the paragraph, this renders as  instead of <figure>

I'm surprised the linter didn't catch that. It certainly complained about lack of newline before `<figure>` back when that was how we did it. Any ideas, @rreverser? Maybe there's a markdownlint rule we can just enable?

RReverser:
Not sure there is a built-in markdownlint rule, because technically it's possible and allowed to have inline images as well, for which this is a valid syntax.

However, if we never use inline images, it should be fairly straighforward to add such custom check in `md-embed-image.js`.

mathiasbynens:
> However, if we never use inline images, it should be fairly straighforward to add such custom check in `md-embed-image.js`.

We never do, and I don't imagine ever needing to do so. Until a use case comes up, adding that check sounds great!

PR #369, opened 2020-04-03
Disallow inline images in a custom plugin

### LINKED ISSUE #368: Add a check to make sure we don't allow inline images
This is usually an accident, and results in a different styling than we want.

RReverser:
Also fixing couple of cases caught by the added check.

Fixes #368.

### DIFF
--- a/md-embed-image.js
+++ b/md-embed-image.js
@@ -7,7 +7,7 @@ const assert = require('assert');
 module.exports = md => {
   md.core.ruler.push('check_img_in_figure', state => {
     let inFigure = false;
-    for (let t of state.tokens) {
+    for (const t of state.tokens) {
       switch (t.type) {
         case 'figure_open':
           assert(!inFigure);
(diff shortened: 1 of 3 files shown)


md-embed-image.js
```diff
@@ -2,8 +2,34 @@
 
 const imageSize = require('image-size');
 const { existsSync } = require('fs');
+const assert = require('assert');
 
 module.exports = md => {
+  md.core.ruler.push('check_img_in_figure', state => {
+    let inFigure = false;
+    for (let t of state.tokens) {
```
mathiasbynens, line 10:
```suggestion
    for (const t of state.tokens) {
```

md-embed-image.js
```diff
@@ -2,8 +2,34 @@
...
+          assert(!inFigure);
+          inFigure = true;
+          break;
+        case 'figure_close':
+          assert(inFigure);
+          inFigure = false;
+          break;
+        case 'inline':
+          if (!inFigure) {
+            let image = t.children.find(t => t.type === 'image');
```
mathiasbynens, line 22:
```suggestion
            const image = t.children.find(t => t.type === 'image');
```

RReverser: Whyyyy do you like `const` so much? 😂

PR #373, opened 2020-04-06
Define a `grammar` highlighter

### LINKED ISSUE #362: Automate and/or document language-grammar
We have a special style for grammar excerpts - a `language-grammar` class - that needs to be used manually in `<pre><code class="language-grammar">...</code></pre>` blocks.

It should be either documented somewhere, or, ideally, we should add a custom highlighter that would support and convert ```` ```language-grammar ```` blocks automatically.

RReverser:
This is intentionally simple for now, but already makes authoring grammar pieces in posts *much* easier than before, and gives us more flexibility with styling.

For the most part it looks like before, and I made only couple of small adjustments. First, only production names are now italic (like in actual spec). Second, `opt` ES grammar keyword and `list+` / `list*` Torque keywords are now highlighted differently, like keywods in other languages.

E.g. before:
image

After:
image

Another example, before:
image

After:
image

Fixes #362.

(cc @marjakh - FYI)

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -73,6 +73,20 @@ module.exports = (eleventyConfig) => {
   eleventyConfig.addPlugin(pluginSyntaxHighlight, {
     init({ Prism }) {
       installPrismLanguages(Prism);
+      Prism.languages.grammar = {
+        // matches the `[...]` part in `ProductionName[...]`
+        'production-params': /(?<=[a-z])\[.*?\]/,
+        // matches `ProductionName`
+        'production-name': /\b[A-Z][A-Za-z_]*\b/,
+        // "but not" and "one of" are special human-readable words in
+        // the ES spec that shouldn't be formatted as literals
+        'skip': /but not|one of/,
+        // `opt` is an ES grammar keyword to be formatted as a subscript
+        // `list+` and `list*` are Torque grammar keywords
+        'keyword': /\bopt\b|\blist[+*]/,
+        // anything else - numbers, words, punctuation - is a literal
+        'literal': /\S+/,
+      };
     },
   });
 
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -73,20 +73,6 @@ module.exports = (eleventyConfig) => {
   eleventyConfig.addPlugin(pluginSyntaxHighlight, {
     init({ Prism }) {
       installPrismLanguages(Prism);
-      Prism.languages.grammar = {
-        // matches the `[...]` part in `ProductionName[...]`
-        'production-params': /(?<=[a-z])\[.*?\]/,
-        // matches `ProductionName`
-        'production-name': /\b[A-Z][A-Za-z_]*\b/,
-        // "but not" and "one of" are special human-readable words in
-        // the ES spec that shouldn't be formatted as literals
-        'skip': /but not|one of/,
-        // `opt` is an ES grammar keyword to be formatted as a subscript
-        // `list+` and `list*` are Torque grammar keywords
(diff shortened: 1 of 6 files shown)


.eleventy.js
```diff
@@ -73,6 +73,20 @@ module.exports = (eleventyConfig) => {
...
+        'production-name': /\b[A-Z][A-Za-z_]*\b/,
+        // "but not" and "one of" are special human-readable words in
+        // the ES spec that shouldn't be formatted as literals
+        'skip': /but not|one of/,
+        // `opt` is an ES grammar keyword to be formatted as a subscript
+        // `list+` and `list*` are Torque grammar keywords
+        'keyword': /\bopt\b|\blist[+*]/,
+        // anything else - numbers, words, punctuation - is a literal
+        'literal': /\S+/,
+      };
```
mathiasbynens, line 89: Would it make sense to move this to its own function outside of `.eleventy.js`, similar to `installPrismLanguages`? Maybe `installPrismExtras()` which then does the two?

RReverser: Ah, I haven't noticed it's a custom module and not a 3rd-party. Looks like we already define some custom grammars like `torque` in that file, so I suppose I could just add this one to the bottom, no need to split out into more files yet.

RReverser: Done.

mathiasbynens: Yeah, adding it to the existing file is what I was suggesting 👍 Thanks!

prism-languages.js
```diff
@@ -111,6 +111,22 @@ const installPrismLanguages = (Prism) => {
...
+    'skip': /but not|one of/,
+    // `opt` is an ES grammar keyword to be formatted as a subscript
+    // `list+` and `list*` are Torque grammar keywords
+    'keyword': /\bopt\b|\blist[+*]/,
+    // anything else - numbers, words, punctuation - is a literal
+    'literal': /\S+/,
+  };
 };
 
 module.exports = installPrismLanguages;
```
mathiasbynens, line 132: Echo'ing my earlier comment: since it's now no longer just about languages, `installPrismLanguages` -> `installPrismExtras`

mathiasbynens: and rename the file accordingly

src/blog/understanding-ecmascript-part-2.md
```diff
@@ -165,14 +165,16 @@ We’ll take a deeper look into the grammar rules in a later episode, let’s ke
...
-  MemberExpression <b>[</b> Expression <b>]</b>
-  MemberExpression <b>.</b> IdentifierName
+  MemberExpression [ Expression ]
+  MemberExpression . IdentifierName
   MemberExpression TemplateLiteral
   SuperProperty
   MetaProperty
-  <b>new</b> MemberExpression Arguments</code></pre>
+  new MemberExpression Arguments
+```
```
mathiasbynens, line 177: This is so awesome!

prism-languages.js
mathiasbynens: Oh, I guess this is a `language` technically from Prism's point of view. Never mind then :)

RReverser: > since it's now no longer just about languages

I'm not sure what you mean. Are you saying that grammar syntax is not a language, so we should rename the function?

If so, then in Prism terms, it's still a "language" as we still add it to `Prism.languages.*`, it still generates a class `language-grammar` and so on, so I'm a bit worried that the naming mismatch will be more confusing not less, as we're not adding any other extras aside from extending `languages`...

PR #374, opened 2020-04-06
Revert "Add syntax highlighting support for numeric separators"

RReverser:
A new version of PrismJS with native support has been since released.

This reverts commit 50078bd57e65809d0f33c1522a4458cef41c14e0.

### DIFF
--- a/prism-languages.js
+++ b/prism-languages.js
@@ -13,11 +13,6 @@
 
 const installPrismLanguages = (Prism) => {
 
-  // Add support for numeric separators.
-  // TODO: Remove this once a `prism-js` version containing
-  // https://github.com/PrismJS/prism/pull/1895 is released.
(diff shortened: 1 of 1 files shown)


RReverser:
@mathiasbynens I'm thinking for cases like this (where a dependency of a dependency has been released) it would be useful for us to commit `package-lock.json` in the repo to make sure we indeed resolve to the desired version.

I usually don't commit `package-lock.json` in most of my projects, because they're libraries, but `v8.dev` is essentially an app and stricter dependency tracking could be useful.

mathiasbynens:
I'm not a fan of `package-lock.json` to be honest, but it's up to you. Thanks for cleaning this up!

RReverser:
> I'm not a fan of package-lock.json to be honest, but it's up to you.

I see. I'll think about it a bit more.

PR #376, opened 2020-04-07
Fix responsive figures for good

### LINKED ISSUE #370: Explicit image dimensions are ignored and result in significant layout jumps
Currently, we set (or, now, generate) explicit `width` and `height` on all images, presumably to avoid layout jumps.

However, these dimensions are getting completely ignored and overridden by https://github.com/v8/v8.dev/blob/cf7598164fdc7a881f93f4a6f058ac95bb5a110e/src/_css/main.css#L412-L416 and result in different sizes once image is loaded.

This results in a significant amount of layout jumps while scrolling through the document, especially due to combination with our extensive usage of `loading="lazy"`.

In order to verify, go to an image-heavy post, e.g. https://v8.dev/blog/pointer-compression/, open DevTools and paste:

```js
[document.body.clientHeight, ...$$('img').map(img => `${img.offsetWidth}x${img.offsetHeight}`)]
```

On my computer, it results in:

```json
[22875, "600x371", "764x240", "764x336", "764x764", "764x764", "764x764", "764x764", "764x764", "764x764", "764x764", "764x764", "600x600", "600x600", "96x96", "96x96"]
```

You can see how any images outside the immediate viewport resulted in 1:1 dimensions (squares) with explicit `height` completely ignored.

Then, scroll through the document to the end and try same expression again:

```json
[18141, "600x371", "764x240", "764x336", "764x336", "764x183", "764x158", "764x336", "764x267", "764x213", "764x103", "764x241", "600x371", "600x371", "96x96", "96x96"]
```

Now that images are loaded, all the heights got updated to actual dimensions from the image itself and the total document height shrunk by 4,734 pixels.

cc @mathiasbynens and @tomayac as it looks like you gave this some thought as part of removal of `intrinsicsize` in #231 / #232 and might have ideas on how to fix this.

tomayac:
For responsive `figure`s (with children `img` or `video`) with calculated aspect ratio to work, their `width` needs to be `100%`, which is something that was missing (we only had `max-width`), so the `width` attribute on the `img` or `video` was used under some circumstances.  

Fixes #370

### DIFF
--- a/src/_css/main.css
+++ b/src/_css/main.css
@@ -325,9 +325,13 @@ figure {
   padding: 0;
   text-align: center;
 }
-figure img {
+figure img,
+figure video {
   display: block;
   margin: 0 auto;
+  width: 100%;
+  max-width: 100%;
+  height: auto;
 }
 figure a {
   border-bottom: 0;
@@ -408,11 +412,6 @@ main .meta {
   border: 0;
   background: var(--tag-background);
   color: var(--tag-color);
-}
-img,
-video {
-  height: auto;
-  max-width: 100%;
 }
 @supports (aspect-ratio: attr(width) / 1) {
   /* https://drafts.csswg.org/css-sizing-4/#example-2476aa5b */


src/_css/main.css
```diff
@@ -325,9 +325,13 @@ figure {
   padding: 0;
   text-align: center;
 }
-figure img {
+figure img,
+figure video {
   display: block;
   margin: 0 auto;
+  width: 100%;
```
mathiasbynens, line 332: I'm confused. On wide viewports, we wouldn't want images that are smaller than the available width to take up the entire width, would we?

tomayac: If we don't want this, then we have the problem that it uses `height: auto`, but the `width` attribute, which breaks the aspect ratio calculation again:

Screen Shot 2020-04-07 at 12 45 11

In practice I think it doesn't matter, since the `body` is capped at `860px`. One way around it would be to also cap `figure` as the container to a maximum width, say `500px`.

RReverser: > One way around it would be to also cap figure as the container to a maximum width, say 500px

This sounds the most promising so far. Are there any downsides to doing that? (aside from hardcoded number)

mathiasbynens: > In practice I think it doesn't matter, since the body is capped at `860px`.

Now I'm even more confused :) Just to be sure, what I meant was that e.g. the image here https://v8.dev/blog/code-caching-for-devs#don%E2%80%99t-change-code gets stretched to the entire width when you give it `width: 100%`, as you'd expect. This is undesirable.

tomayac:
```html
<style>
  body { max-width: 860px;}
  figure { max-width: 400px; } /* container */
  figure > img, figure > video {
    width: 100%;
    max-width: 100%;
    height: auto;
  }
</style>

<figure>
  <!-- The container rule would make sure the image is capped at 400px -->
  <img width=1600 height=400 src=foo>
</figure>
```

mathiasbynens: That seems suboptimal; ideally we'd cap images at their actual size, which is what we currently do.

tomayac: The only way that I'm aware of to make this happen while keeping the aspect ratio calculation *during* the loading phase effective would be to add a manual `max-width` with the individual pixel width to each image.

RReverser: > That seems suboptimal; ideally we'd cap images at their actual size, which is what we currently do.

Don't we currently cap them at the container (page) width as well?

tomayac: > Don't we currently cap them at the container (page) width as well?

This is what I was saying [above](https://github.com/v8/v8.dev/pull/376#discussion_r404715511).

mathiasbynens: > > That seems suboptimal; ideally we'd cap images at their actual size, which is what we currently do.
> 
> Don't we currently cap them at the container (page) width as well?

Well, yes, I was talking about the case where images are smaller (not larger) than the container, e.g. on a wide viewport. What I'm really saying is: IMHO we shouldn't stretch images, and hardcoding a number like that `400px` is always gonna result in images smaller than that getting stretched.

tomayac: Are there many non-vector images?

RReverser: > What I'm really saying is: IMHO we shouldn't stretch images, and hardcoding a number like that 400px is always gonna result in images smaller than that getting stretched.

Ah I misunderstood what that code does. Makes sense. Given that we specify dimensions on all images, can we use CSS `min()` + `attr()` perhaps?

> Are there many non-vector images?

There are some screenshots and such, yeah.

tomayac: > Given that we specify dimensions on all images, can we use CSS min() + attr() perhaps?

[Not yet](https://drafts.csswg.org/css-values-4/#attr-notation) 😔

> There are some screenshots and such, yeah.

We could work with a selector that applies the `width: 100%` rule to only SVGs: `img[src$=".svg"]`.

mathiasbynens: > We could work with a selector that applies the width: 100% rule to only SVGs: `img[src$=".svg"]`.

There's SVGs that we don't want to upscale though, e.g. https://v8.dev/blog/liftoff
(Search `dist/` for `width=2**`, `width=3**`, etc. to find more examples.

tomayac: Fair. How about having the best of both worlds? In a pre-processing step before deployment you can for each image with a width ≤ 300px (or whatever threshold you come up with) apply an inline style. This makes sure "wide enough" images get 100% width and auto height, and the smaller images are capped by the inline style. Both will get you the aspect-ratio-dependent space pre-reservation in rendering.

Screen Shot 2020-04-08 at 10 19 37

mathiasbynens: @tomayac Wouldn't that break on viewports narrower than whatever the inline `width` style says?

I was pointing out extreme examples before, but I still think we don't want to stretch images in _any_ case (not even when they're SVG and large enough).

Taking a step back, while I appreciate the creative brainstorming here, I think we might be over-engineering this. Responsive images really shouldn't be this hard. Maybe we should go back to https://github.com/v8/v8.dev/issues/370#issuecomment-609744033 and consider that this might be a Chrome bug? @RReverser do you want to file a crbug with your initial findings?

tomayac: > @tomayac Wouldn't that break on viewports narrower than whatever the inline width style says?

True, we'd need to select the threshold accordingly (maybe aimed at KaiOS as the smallest).

> Maybe we should go back to #370 (comment) and consider that this might be a Chrome bug? 

I do not think this is a Chrome bug, since we didn't have `width: 100%`: https://youtu.be/4-d_SoCHeWE?t=438.

tomayac: Hmmm, wait. It works fine in Canary (84) and also Firefox Nightly (77). Could it be that this simply hasn't landed (correctly) in Beta (81)?

tomayac: Yuck, this really bugged me so I dug deeper. Turns out we're far less advanced in the implementation than I thought: https://crbug.com/1045668. It *will* eventually be good, but still isn't.

tomayac:
I close this PR. After further investigation, turns out what we already have *eventually* will be sufficient. We're now waiting for https://crbug.com/1045668 to be closed.

mathiasbynens:
Thanks for digging in, @tomayac!

RReverser:
Sigh, but thanks for all the work! Glad to hear this is being fixed on the browser side.

PR #377, opened 2020-04-07
No more flashing of white in dark mode

### LINKED ISSUE #335: FOUC with the dark theme on first page load or refresh
A dark theme that always flashes a white frame on first page load or refresh isn't really working and I would even recommend disabling it until this is fixed, provided you don't want to blind your users.

image

image

Hopefully, it'd be simple to move the script/style to an earlier time slot. Otherwise, a general solution is either to use `<link>` with `media="(prefers-color-scheme: dark)"` in `<head>` or check the theme name in an inline `<script>` in `<head>` and toggle the corresponding `<link>` or `<style>` element if for whatever reason you don't want to rely on a media query.

tomayac:
This re-establishes the previous `media`-based architecture and as a by-product also reduces the JavaScript size.

Fixes #335

### DIFF
--- a/src/_css/dark.css
+++ b/src/_css/dark.css
@@ -1,0 +1,40 @@
+/* Copyright 2018 Google Inc.
+**
+** Licensed under the Apache License, Version 2.0 (the “License”);
+** you may not use this file except in compliance with the License.
+** You may obtain a copy of the License at
+** <https://apache.org/licenses/LICENSE-2.0>.
+**
+** Unless required by applicable law or agreed to in writing, software
+** distributed under the License is distributed on an “AS IS” BASIS,
+** WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+** See the License for the specific language governing permissions and
+** limitations under the License.
+*/
+
+/* Dark mode */
+:root {
+  --html-color: var(--nearwhite);
+  --html-background: var(--black);
+  --main-color:var(--nearwhite);
+  --main-background: var(--darkestgray);
+  --footer-color: var(--nearwhite);
+  --footer-background: var(--darkgray);
+  --main-and-footer-link-color: var(--yellow);
+  --avatar-img-filter: grayscale(50%);
+  --icon-no-support-filter: brightness(0.5) grayscale(100%);
+  --note-color: var(--darkestgray);
+  --note-background: var(--yellow);
+  --tag-color: var(--nearwhite);
+  --tag-background: var(--darkgray);
+  --table-border: var(--lightgray);
+  --logo-hover-background: var(--darkgray);
+  --alt-table-row-background: #111;
+
+  -webkit-font-smoothing: antialiased;
+}
+
+figure img[src*=".png"]:not(.no-darkening),
+figure img[src*=".svg"]:not(.no-darkening) {
+  filter: invert(0.862745) hue-rotate(180deg);
+}
--- a/src/_css/dark.css
+++ b/src/_css/dark.css
@@ -1,4 +1,4 @@
-/* Copyright 2018 Google Inc.
+/* Copyright 2020 Google Inc.
 **
 ** Licensed under the Apache License, Version 2.0 (the “License”);
 ** you may not use this file except in compliance with the License.
--- a/src/_css/light.css
+++ b/src/_css/light.css
@@ -1,0 +1,32 @@
+/* Copyright 2018 Google Inc.
+**
+** Licensed under the Apache License, Version 2.0 (the “License”);
+** you may not use this file except in compliance with the License.
+** You may obtain a copy of the License at
+** <https://apache.org/licenses/LICENSE-2.0>.
+**
+** Unless required by applicable law or agreed to in writing, software
+** distributed under the License is distributed on an “AS IS” BASIS,
+** WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+** See the License for the specific language governing permissions and
+** limitations under the License.
+*/
+
+/* Light mode */
+:root {
+  --html-color: var(--black);
+  --html-background: var(--darkgray);
+  --main-color: var(--black);
+  --main-background: var(--white);
+  --footer-background: var(--lightestgray);
+  --main-and-footer-link-color: var(--darkblue);
+  --avatar-img-filter: none;
+  --icon-no-support-filter: grayscale(100%);
+  --note-color: var(--main-color);
+  --note-background: var(--lightyellow);
+  --tag-color: var(--black);
+  --tag-background: var(--lightestgray);
+  --table-border: var(--lightestgray);
+  --logo-hover-background: var(--lightestgray);
+  --alt-table-row-background: #EEE;
+}
--- a/src/_css/light.css
+++ b/src/_css/light.css
@@ -1,4 +1,4 @@
-/* Copyright 2018 Google Inc.
+/* Copyright 2020 Google Inc.
 **
 ** Licensed under the Apache License, Version 2.0 (the “License”);
 ** you may not use this file except in compliance with the License.
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -5,6 +5,27 @@
     <title>{% if '/' === page.url %}V8 JavaScript engine{% else %}{{ (renderData.title or title) | markdown | striptags | decodeHtmlEntities }} · V8{% endif %}</title>
     <meta name="viewport" content="width=device-width,initial-scale=1">
     <meta name="color-scheme" content="dark light">
+    <script>
+      // If `prefers-color-scheme` is not supported, fall back to light mode.
+      // In this case, light.css will be downloaded with `highest` priority.
+      if (window.matchMedia('(prefers-color-scheme: dark)').media === 'not all') {
+        document.documentElement.style.display = 'none';
+        document.head.insertAdjacentHTML(
+            'beforeend',
+            '<link rel="stylesheet" href="/_css/light.css" onload="document.documentElement.style.display = \'\'">'
+        );
+      }
+    </script>
+    <!--
+      Conditionally either load the light or the dark stylesheet. The matching file
+      will be downloaded with `highest`, the non-matching file with `lowest`
+      priority. If the browser doesn't support `prefers-color-scheme`, the media
+      query is unknown and the files are downloaded with `lowest` priority (but
+      above I already force `highest` priority for my default light experience).
+    -->
+    <link rel="stylesheet" href="/_css/dark.css" media="(prefers-color-scheme: dark)">
+    <link rel="stylesheet" href="/_css/light.css" media="(prefers-color-scheme: no-preference), (prefers-color-scheme: light)">
+    <!-- The main stylesheet -->
     <link rel="stylesheet" href="/_css/main.css">
 {% if '/features' in page.url %}
     <link rel="stylesheet" href="/_css/feature-support.css">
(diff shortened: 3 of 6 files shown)


src/_css/dark.css
```diff
@@ -0,0 +1,40 @@
+/* Copyright 2018 Google Inc.
```
mathiasbynens, line 1:
```suggestion
/* Copyright 2020 Google Inc.
```

src/_css/light.css
```diff
@@ -0,0 +1,32 @@
+/* Copyright 2018 Google Inc.
```
mathiasbynens, line 1:
```suggestion
/* Copyright 2020 Google Inc.
```

src/_includes/layouts/base.njk
```diff
@@ -5,6 +5,27 @@
     <title>{% if '/' === page.url %}V8 JavaScript engine{% else %}{{ (renderData.title or title) | markdown | striptags | decodeHtmlEntities }} · V8{% endif %}</title>
     <meta name="viewport" content="width=device-width,initial-scale=1">
     <meta name="color-scheme" content="dark light">
+    <script>
+      // If `prefers-color-scheme` is not supported, fall back to light mode.
+      // In this case, light.css will be downloaded with `highest` priority.
+      if (window.matchMedia('(prefers-color-scheme: dark)').media === 'not all') {
```
mathiasbynens, line 11: Let's not do this. This only affects old IE, which we don't actively support anyway, and the site works fine in dark mode.

mathiasbynens: i.e. can we just remove this entire `<script>`?

RReverser: > and the site works fine in dark mode

I actually wonder what will happen to styles below if `prefers-color-scheme` is not supported. Are you saying the dark one will be loaded? I would think neither one will, but I don't know much about how unsupported media selectors work.

src/_includes/layouts/base.njk
```diff
@@ -5,6 +5,27 @@
...
+    </script>
+    <!--
+      Conditionally either load the light or the dark stylesheet. The matching file
+      will be downloaded with `highest`, the non-matching file with `lowest`
+      priority. If the browser doesn't support `prefers-color-scheme`, the media
+      query is unknown and the files are downloaded with `lowest` priority (but
+      above I already force `highest` priority for my default light experience).
+    -->
+    <link rel="stylesheet" href="/_css/dark.css" media="(prefers-color-scheme: dark)">
+    <link rel="stylesheet" href="/_css/light.css" media="(prefers-color-scheme: no-preference), (prefers-color-scheme: light)">
```
mathiasbynens, line 27: Once we remove the above `<script>`, we can simplify things further by moving the contents of `light.css` back into `main.css`, and then only conditionally loading `dark.css` when dark mode is enabled.

mathiasbynens: Should we load these below `main.css`?

src/_includes/layouts/base.njk
```diff
@@ -5,6 +5,27 @@
...
+      }
+    </script>
+    <!--
+      Conditionally either load the light or the dark stylesheet. The matching file
+      will be downloaded with `highest`, the non-matching file with `lowest`
+      priority. If the browser doesn't support `prefers-color-scheme`, the media
+      query is unknown and the files are downloaded with `lowest` priority (but
+      above I already force `highest` priority for my default light experience).
+    -->
+    <link rel="stylesheet" href="/_css/dark.css" media="(prefers-color-scheme: dark)">
```
RReverser, line 26: Given how small these files are and the fact that we download them anyway, is it even worth splitting them up into separate requests? We probably add more overhead than gain anything.

Also, with pure CSS we could use `@supports not (prefers-color-scheme: dark)` as one of the rules for e.g. the light theme so it would nicely fallback without the JS above. Arguably, we could do the same with `@import` inside such condition, but at that point it seems even better to keep it all in one file.

src/_includes/layouts/base.njk
RReverser: I've added a suggestion below on just keeping it all in one file. IMO that would be simpler (including for fallback logic) and even faster than separate requests given the size of these files.

src/_includes/layouts/base.njk
mathiasbynens: > Given how small these files are and the fact that we download them anyway, is it even worth splitting them up into separate requests? We probably add more overhead than gain anything.

+1

> Also, with pure CSS we could use `@supports not (prefers-color-scheme: dark)` as one of the rules for e.g. the light theme so it would nicely fallback without the JS above.

I don't think `@supports` works with media queries. But +1 to doing as much as we can with CSS instead of JS.

src/_includes/layouts/base.njk
mathiasbynens: The less files the better!

src/_includes/layouts/base.njk
tomayac: They get loaded, but [both with low priority](https://blog.tomayac.com/2018/11/08/why-browsers-download-stylesheets-with-non-matching-media-queries-180513/). 

All up for removing, but one other browser we *might* want to consider is KaiOS.

src/_includes/layouts/base.njk
tomayac: We've had this discussion before. Just proceed with what you think is the best solution. There're no new arguments to be exchanged. The PR as it was addresses the issue. In one possible way of many.

tomayac: Things [my solution](https://web.dev/prefers-color-scheme/#dark-mode-in-practice) tries to consider:
- Avoiding unused CSS
- Loading files with the right priority
- Making the theme switching easy
- Avoiding flashes of white
- Being backward compatible without causing harm

src/_includes/layouts/base.njk
RReverser: Loaded, yes, but neither one is applied IIUC?

src/_includes/layouts/base.njk
RReverser: > I don't think @supports works with media queries.

Ah, you mean when combining both in the same rule?

src/_includes/layouts/base.njk
tomayac: Correct. This is why the inline script injects one of them (`light.css`) manually, so it gets loaded with highest priority and is applied, since `main.css` relies on the variable definition made therein. Try my blog on a KaiOS device to see it in action (or a [VM](https://www.browserling.com/browse/win/7/firefox/48/https%3A%2F%2Fblog.tomayac.com%2F)).

src/_includes/layouts/base.njk
tomayac: `@import` has the huge disadvantage of late discovery. We shouldn't advocate this.

RReverser: > Avoiding unused CSS
> Loading files with the right priority

I understand, but I think metrics not considered here are overhead of individual request (headers etc.), size of individual files after minification (our theme-specific styles are very tiny) as well as compression ratio (it's better when everything is in one file).

In the end, for this particular case it's IMO still benefitial performance-wise to keep things in one file, but I'm happy to be proven wrong by numbers.

tomayac: I’ve been countering request overhead with HTTP/2.

I agree this file architecture might be overkill for v8.dev in the concrete case, but in the general case, people who might take inspiration from v8.dev’s architecture might have waaaay more CSS. 

Again, proceed with what you think is best, my PR just fixed the original issue and makes the toggling work painlessly.

RReverser: > people who might take inspiration from v8.dev’s architecture might have waaaay more CSS

I suppose, but while that's nice, that's not the end goal IMO. We still care about UX for our real users more :)

> Again, proceed with what you think is best, my PR just fixed the original issue and makes the toggling work painlessly.

That's awesome and definitely appreciate that! Just discussing various options to improve before merging.

tomayac: Just to clarify *why* the toggling is painless: the `media` attribute gets changed to activate or deactivate the different stylesheets.

RReverser: Yup, that's clear, thanks! 👍🏻

tomayac: This article just crossed my stream: https://webcloud.se/blog/2020-04-06-flash-of-unstyled-dark-theme/.

RReverser: So could we merge these files back together and proceed like that?

tomayac: So let's take a step back:

Independent from `prefers-color-scheme`, we need to wait until `<dark-mode-toggle>` has [figured out the preferred theme](https://github.com/v8/v8.dev/blob/f9a62400eab6565b2a45eaf1683e38d00bb211c6/src/_js/main.mjs#L54) and potentially applies the `dark` class. Since we render before this happens, the user sees a flash of white. This is avoidable by simply not rendering before the toggle code has run (`root: { display: none; }` and once the toggle has done its work programmatically setting `root: { display: block; }`). Or, and this is what this PR suggests, making the CSS files load conditionally and blockingly with the `media` attribute.

RReverser: I see. I suppose we could instead have dark-mode-toggle add some class to the top element like it's often done (`.loaded` or something) and use that class in CSS condition in addition to `prefers-color-scheme`?

Then we can load CSS like we do today, but it would be applied only when dark-mode-toggle is ready.

tomayac: This would work, too, but you would still need to block the rendering.

RReverser: That's true, even though it unblocks layout at least. Too bad we don't have some sort of stats from the toggle - we could know whether people use it often enough or whether we can render `prefers-color-scheme` variant first and switch later.

Actually, maybe we should do that anyway? As in, while the toggle is not loaded, show variant that depends on `prefers-color-scheme` right away, and switch to the class-based one afterwards?

Then at least first-time visits won't see the flash (because they don't have toggled option yet), and neither will users whose option matches the system-wide theme.

This still leaves users with explicit override, but for them JS should be usually cached already, and the flash would be quick. Overall this is still not ideal but better than current state of things as it reduces or eliminates flash for majority of users?

tomayac: > Actually, maybe we should do that anyway? As in, while the toggle is not loaded, show variant that depends on `prefers-color-scheme` right away, and switch to the class-based one afterwards?

This would not fly with whoever has used the toggle to override their system preference. 

Here's a crazy idea: what if we inline `light.css` and `dark.css` as data URLs (since they're too small to justify an HTTP request), so the `media` attribute-based switching still works?

RReverser: > This would not fly with whoever has used the toggle to override their system preference.

Yes, that's what I said that for them flashing will still occur, but at least everyone else will get non-blocking rendering of the correct theme right away.

> Here's a crazy idea: what if we inline light.css and dark.css as data URLs (since they're too small to justify an HTTP request), so the media attribute-based switching still works?

How is that different from the `.loading` class in terms of behaviour? This seems a bit more complicated to maintain if anything...

tomayac: > How is that different from the `.loading` class in terms of behaviour? This seems a bit more complicated to maintain if anything...

This approach would inherit the native render-blocking behavior of CSS just without the request overhead, and without the need for additional JavaScript.

RReverser: Ah I see what you mean. I think I'd still prefer keeping them together - aside from actual request overhead, splitting provides worse compression and, as mentioned, more complicated maintainability. Let's keep things simple.

tomayac: I close this PR now, since it's definitely not going in the direction that you wanted it to go. It is easier to start from scratch with a "block rendering until toggle has figured out the color scheme" approach.

PR #379, opened 2020-04-08
Fix typo in GN name: v8_untrusted_code_mitigations

harikb:
Current documentation states it as `v8_enable_untrusted_code_mitigations` (extra word `enable` in the middle) in one of the two places. The build will ignore this flag and the user wouldn't benefit from the important untrusted mitigations.


harikb:
@googlebot I signed it!

natorion:
Thanks for the fix!

PR #385, opened 2020-04-15
Update Torque types documentation

sethbrenith:
Remove some things that are no longer true:
- Torque namespaces correspond to generated files and assembler classes
- dirtyInstantiatedAbstractClass
- Verifiers implicitly allow Undefined in some cases

Add some things that were undocumented:
- A bunch of class annotations
- Indexed fields
- Const and weak fields
- Classes with no body definition
- Shapes
- Structs
- Bitfield structs

Other changes:
- Moved union types above classes because some class examples use them
- Changed example class from JSFunction to JSProxy for simplicity (and
  because JSFunction does some creepy things that we don't want anyone
  to copy)
- Used generateCppClass in the primary class example and showed the
  other option later, to encourage doing the right thing


src/docs/torque.md
```diff
@@ -207,112 +178,236 @@ type int32 generates 'TNode<Int32T>' constexpr 'int32_t';
...
 
-It is only allowed to form union types from tagged types because untagged types cannot be distinguished at runtime.
+`const` means that a field cannot be altered at runtime (or at least not easily; Torque will fail compilation if you attempt to set it). This is a good idea for length fields, which should only be reset with great care because they would require freeing any released space and might cause data races with a marking thread.
 
-When mapping union types to CSA, the most specific common supertype of all the types of the union type is selected, with the exception of `Number` and `Numeric`, which are mapped to the corresponding CSA union types.
+`weak` at the beginning of a field declaration means that the field is visited with some kind of custom weak object semantics. If the object may be a normal weak ref (with the second bit set), then `Weak<T>` should be used in the type instead. As an example, this field from `Map` can contain some strong and some weak types, and also requires custom weak visitation:
+
+```torque
+  weak transitions_or_prototype_info: Map|Weak<Map>|TransitionArray|
+      PrototypeInfo|Smi;
```
tebbi, line 326: The truth is: "weak" doesn't have any proper semantics and just affects the generation of the kStartOf...-style constants. As far as I can tell from the actual BodyDescriptor, the transitions_or_prototype_info field is a normal field with tagged weak references, but without custom weakness. The reason why removing the weak keyword would not work at the moment is that the body descriptor actually says:

    IteratePointers(obj, Map::kStartOfStrongFieldsOffset,
                    Map::kEndOfStrongFieldsOffset, v);
    IterateMaybeWeakPointer(obj, kTransitionsOrPrototypeInfoOffset, v);

So we need the weak keyword to keep the Map::kEndOfStrongFieldsOffset correct.
This should go away and we should stop using these constants by using Torque-generated body descriptors, hopefully soon....

src/docs/torque.md
```diff
@@ -207,112 +178,236 @@ type int32 generates 'TNode<Int32T>' constexpr 'int32_t';
...
+
+Some classes are not defined in Torque, but Torque must know about every class because it is responsible for assigning instance types. For this case, classes can be declared with no body, and Torque will generate nothing for them except the instance type. Example:
+
+```torque
+extern class OrderedHashMap extends HashTable;
+```
+
+#### Shapes
+
+Defining a `shape` looks just like defining a `class` except that it uses the keyword `shape` instead of `class`. A `shape` is a subtype of JSObject representing a point-in-time arrangement of properties on that object (in spec-ese, these are "data properties" rather than "internal slots"). It does not have its own instance type. An object with a particular shape may change and lose that shape at any time because the object might go into dictionary mode and move all of its properties out to a separate backing store.
```
tebbi, line 341: We should mention that shapes always define in-object properties.

src/docs/torque.md
```diff
@@ -207,112 +178,236 @@ type int32 generates 'TNode<Int32T>' constexpr 'int32_t';
...
+  raw_number_of_marked_descriptors: uint16;
+  filler16_bits: uint16;
+  enum_cache: EnumCache;
+  descriptors[number_of_all_descriptors]: DescriptorEntry;
+}
+```
+
+##### References and Slices
+
+`Reference<T>` and `Slice<T>` are special structs representing pointers to data held within heap objects. They both contain an object and an offset; `Slice<T>` also contains a length. Rather than constructing these structs directly, you can use special syntax: `&o.x` will create a `Reference` to the field `x` within the object `o`, or a `Slice` to the data if `x` is an indexed field. `Reference<T>` can be dereferenced with `*` or `->`, consistent with C++.
```
tebbi, line 395: This should probably also mention that Reference\<T\> is not used directly, instead it has two subtypes  MutableReference\<T\> and ConstReference\<T\>, which can be referred to using syntactic sugar: &T and const &T.

sethbrenith:
Thanks @tebbi for the great suggestions! Updated to correct the things you pointed out, and also to make the "hello world" example compile.

PR #386, opened 2020-04-16
Point wasm_simd128.h to the LLVM repo

RReverser:
The actual header has been moved upstream, and Emscripten one just forwards the declaration. (cc @tlively @dtig for reviews - looks like you're not part of the V8 org, so I can't assign you as proper reviewers in Github UI)


src/features/simd.md
```diff
@@ -90,7 +90,7 @@ But when the `-msimd128` flag is used, the autovectorizer turns this into code t
 
 The loop body has the same structure but SIMD instructions are being used to load, multiply, and store four elements at a time inside the loop body.
 
-For finer grained control over the SIMD instructions generated by the compiler, include [this header file](https://github.com/emscripten-core/emscripten/blob/master/system/include/wasm_simd128.h), which defines a set of intrinsics. Intrinsics are special functions that, when called, will be turned by the compiler into the corresponding WebAssembly SIMD instructions, unless it can make further optimizations.
+For finer grained control over the SIMD instructions generated by the compiler, include [`wasm_simd128.h` header file](https://github.com/llvm/llvm-project/blob/master/clang/lib/Headers/wasm_simd128.h), which defines a set of intrinsics. Intrinsics are special functions that, when called, will be turned by the compiler into the corresponding WebAssembly SIMD instructions, unless it can make further optimizations.
```
tlively, line 93:
```suggestion
For finer grained control over the SIMD instructions generated by the compiler, include the [`wasm_simd128.h` header file](https://github.com/llvm/llvm-project/blob/master/clang/lib/Headers/wasm_simd128.h), which defines a set of intrinsics. Intrinsics are special functions that, when called, will be turned by the compiler into the corresponding WebAssembly SIMD instructions, unless it can make further optimizations.
```

PR #387, opened 2020-04-20
wasm-decompile blog post

aardappel:


src/blog/wasm-decompile.md
```diff
@@ -0,0 +1,233 @@
...
+author: 'Wouter van Oortmerssen'
+avatars:
+  - 'wouter-van-oortmerssen'
+date: 2020-04-20 12:35:00
+tags:
+  - WebAssembly
+  - tooling
+description: 'WABT gains a new decompilation tool that can make it easier to
+read the contents of Wasm modules'
+tweet: '?'
```
RReverser, line 12: Please leave empty for now, otherwise this will generate an invalid URL.

src/blog/wasm-decompile.md
```diff
@@ -0,0 +1,233 @@
...
+  local.get 0
+  f32.load offset=8
+  local.get 1
+  f32.load offset=8
+  f32.mul
+  f32.add))
+```
+
+That is a tiny bit of code, but already not great to read for many reasons.
+Besides the lack of an expression based syntax and general verbosity,
```
RReverser, line 70: To be completely fair, it might be helpful to compare with a folded WAT (`wasm2wat -f`) which makes it somewhat more readable.

aardappel: good point, will fix.

aardappel:
@RReverser I seem to have fixed the lint errors, but the remaining CI errors I have no idea what is going on.

src/blog/wasm-decompile.md
```diff
@@ -0,0 +1,225 @@
+---
+title: "What's in that .wasm? Introducing: wasm-decompile"
+author: 'Wouter van Oortmerssen'
```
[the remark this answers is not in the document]
aardappel: Ok, will do!

src/blog/wasm-decompile.md
```diff
@@ -0,0 +1,225 @@
...
+## Disassemble or.. decompile?
+
+You could use tools like `wasm2wat` (part of the
+[WABT](https://github.com/WebAssembly/wabt) toolkit), to transform a `.wasm`
+into Wasm's standard text format, `.wat`, which is a very faithful but not
+particularly readable representation.
+
+For example, a simple C function like a dot product:
+
+```C
```
RReverser, line 34: Looks like CI failed because of this: https://travis-ci.com/github/v8/v8.dev/jobs/321362009#L341

I think you need to use lowercase tag (`c`) or maybe even `cpp` in case that one doesn't work.

RReverser:
General comment: we don't split lines manually in Markdown - this results in actual newlines in output and broken layout:

image

aardappel:
@RReverser err.. pretty much all markdown out there is authored to fit 80 columns or whatever, since it also need to be read by humans, often alongside source code in commits etc. So newlines is the norm.

Not sure what markdown renderer this uses, but if it inserts newlines, that is very non-standard, and would mis-render pretty much all markdown documents out there. Would be worth fixing.

aardappel:
If you do "view file" on the `.md`, you see that github's own markdown renderer does not insert newlines. Nor does `pandoc` (one of the most used command-line rendering tools) etc etc.

RReverser:
> err.. pretty much all markdown out there is authored to fit 80 columns or whatever, since it also need to be read by humans, often alongside source code in commits etc. So newlines is the norm.

It's fairly common to enable word-wrap for Markdown, like in any word processor, when reading by humans. It's not code that needs to be structured and indented in some specific way, but rather a human-readable text, so there's no point in limiting its width.

aardappel:
I think most programmers would disagree :) Either way, is this a setting that can be turned off? What's the point of inserting these line breaks anyway? Paragraph breaks are still done with empty lines, so there is no use for it other than rendering most markdown incorrectly.

RReverser:
> I think most programmers would disagree :)

Well I'm a programmer too, but think it's reasonable (although it wasn't me who set it up initially) 😀 I think it's a preference discussion of which can get as heated as tabs vs spaces tbh :)

> Paragraph breaks are still done with empty lines

Yes, but regular newlines (non-paragraph ones) are not. You can also notice that e.g. if you write a comment with a newline on Github, then it does insert a newline, even if it doesn't in separate md files (yay consistency):

We have a growing number of compilers and other tools that generate or
manipulate `.wasm` files, and sometimes you might want to have a look inside.
Maybe you're a developer of such a tool, or more directly, you're simply a
programmer targeting Wasm, and wondering what the generated code looks like,
for performance or other reasons.

---

Anyway, let's just please follow existing conventions here. I don't think we want to change config for the entire blog at this point / as part of this PR.

src/blog/wasm-decompile.md
```diff
@@ -0,0 +1,225 @@
...
+avatars:
+  - 'wouter-van-oortmerssen'
+date: 2020-04-20 12:35:00
+tags:
+  - WebAssembly
+  - tooling
+description: 'WABT gains a new decompilation tool that can make it easier to
+read the contents of Wasm modules'
+---
+We have a growing number of compilers and other tools that generate or
```
mathiasbynens, line 13: Throughout the file: don't use hard line breaks unless you semantically want to insert a hard break.

src/blog/wasm-decompile.md
```diff
@@ -0,0 +1,225 @@
+---
+title: "What's in that .wasm? Introducing: wasm-decompile"
```
mathiasbynens, line 2:
```suggestion
title: 'What’s in that `.wasm`? Introducing: `wasm-decompile`'
```

src/blog/wasm-decompile.md
```diff
@@ -0,0 +1,225 @@
...
+author: 'Wouter van Oortmerssen ([@wvo](https://twitter.com/wvo))'
+avatars:
+  - 'wouter-van-oortmerssen'
+date: 2020-04-20 12:35:00
+tags:
+  - WebAssembly
+  - tooling
+description: 'WABT gains a new decompilation tool that can make it easier to
+read the contents of Wasm modules'
```
mathiasbynens, line 11: nit: this should read like a sentence, including a dot at the end.

src/blog/wasm-decompile.md
```diff
@@ -0,0 +1,225 @@
...
+---
+We have a growing number of compilers and other tools that generate or
+manipulate `.wasm` files, and sometimes you might want to have a look inside.
+Maybe you're a developer of such a tool, or more directly, you're simply a
```
mathiasbynens, line 15: drop "simply"

src/blog/wasm-decompile.md
aardappel: See discussion above.. this is not how anyone else uses markdown. But whatever.

src/blog/wasm-decompile.md
```diff
@@ -0,0 +1,150 @@
...
+float dot(const vec3 *a, const vec3 *b) {
+    return a->x * b->x +
+           a->y * b->y +
+           a->z * b->z;
+}
+```
+
+We use `clang dot.c -c -target wasm32 -O2` followed by `wasm2wat -f dot.o` to turn it into this `.wat`:
+
+```lisp
```
RReverser, line 35: We have a dedicated `wasm` highlighter:

```suggestion
```wasm
```

PR #389, opened 2020-04-24
Update wording in optional catch binding text

jridgewell:
It's not really a parameter. We refer to it as a "binding" in the text above, so I reused that.


RReverser:
Thanks, looks reasonable!

PR #390, opened 2020-04-27
Add logical assignment feature explainer

syg:


src/features/logical-assignment.md
```diff
@@ -0,0 +1,90 @@
...
+---
+title: 'Logical assignment'
+author: 'Shu-yu Guo ([@_shu](https://twitter.com/_shu))'
+avatars:
+  - 'shu-yu-guo'
+date: 2020-04-22
+tags:
+  - ECMAScript
+description: 'JavaScript now supports compound assignment with logical operations.'
+tweet: 'TODO'
```
RReverser, line 12: Please leave as empty string, otherwise this will generate invalid URL.

syg: Done.

RReverser:
Interesting. Maybe we should have a 3rd state `flagged` or something. @mathiasbynens do you have an opinion / how was this done before?

mathiasbynens:
We already have a third state, `partial`, which we should probably use for that. We already use it in a few places such as https://v8.dev/features/regexp-match-indices#support-for-regexp-match-indices.

RReverser:
Sounds good. From what I understand, the status would be `partial` for all 3 browsers at the moment?

RReverser:
I've made the changes and will publish this tomorrow now.

chicoxyzzy:
In that case many other feature pages on web.dev should also be updated since they have flagged features support as `no`. Wouldn't it be better to set the values to `no` here for consistency with other pages and update all pages later to `partial` (or even better `flagged` IMO, because no support by default is not the same thing as partial support) in next PR?

chicoxyzzy:
Partial was originally added for this case https://github.com/v8/v8.dev/pull/238#issuecomment-527658362

mathiasbynens:
I think having a single state that captures both "flagged" or "partial" is good enough for now: either of those effectively means you can experiment with the implementation but not rely on it fully. Adding more states would introduce even more maintenance overhead for relatively little value.

PR #391, opened 2020-04-29
Optional chaining support available in Firefox 74

Signez:
This awesome language feature is now available in Firefox. Let's update the support table to let developers know the good news!

See release notes at Mozilla:
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/74#JavaScript


Signez:
@googlebot I signed it!

PR #399, opened 2020-05-06
Add a blog post: Understanding the ECMAScript spec, part 4

marjakh:
This is not super polished yet, but sending this out for early comments.


src/blog/understanding-ecmascript-part-4.md
```diff
@@ -0,0 +1,280 @@
...
+
+In [part 3](/blog/understanding-ecmascript-part-3), we familiarized ourselves with the lexical grammar, the syntactic grammar, and the shorthands used for defining the syntactic grammar.
+
+## Meanwhile in other parts of the Web
+
+[Jason Orendorff](https://github.com/jorendorff) from Mozilla published [a great in-depth analysis of JS syntactic quirks](https://github.com/mozilla-spidermonkey/jsparagus/blob/master/js-quirks.md#readme). Even though the implementation details differ, every JS engine faces the same problems with these quirks.
+
+## Cover grammars
+
+In this episode, we take a deeper look into *cover grammars*. They are a way to specify grammar  rules for syntactic constructs where we don't know what we're looking at until we've seen the complete construct.
```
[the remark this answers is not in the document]
syg: Something feels off about the "until we've seen the complete construct" phrasing. I was tempted to say something like "where we don't know what we're looking at from the initial token alone", but that's less understandable...

src/blog/understanding-ecmascript-part-4.md
```diff
@@ -0,0 +1,280 @@
...
+
+PrimaryExpression :
+...
+CPEAAPL
+
+```
+
+Imagine we're again in the situation that we need to parse an `AssignmentExpression` and the next token is `(`. Now we can just decide to parse a `CPEAAPL` and figure out later what it actually is. It doesn't matter whether we're parsing an `ArrowFunction` or a `ParenthesizedExpression`, the next symbol to parse is `CPEAAPL` in any case!
+
+After we've parsed the `CPEAAPL`, we can decide whether the original `AssignmentExpression` is an `ArrowFunction` or a `ParenthesizedExpression` based on the token following the `CPEAAPL`.
```
syg, line 140: Pedantic aside: this isn't quite what the "covers" semantics in ecma262 does.

It doesn't parse CPEAAPL and then disambiguates based on the following token. It parses the entire program into an AST first. So, the AST is already something like (ArrowFunction (ArrowParameters (CPEAAPL ...)) `=>` (ConciseBody ...)). Then it attempts to refine the token list that compromises CPEAAPL into ArrowFormalParameters. If this "covering" of ArrowFormalParameters succeeds on that token list without any tokens left over, then we have a successful parse. 

In practice, the easiest way to understand this is to disambiguate using the immediately following non-terminal. But, it'd be good to call out this distinction, since the next section is all about restrictions. None of the Supplemental Syntax sections and covering restrictions talk about non-terminals outside of the production at all, so the reader might wonder "wait, I thought you said these were disambiguated by looking at what follows".

src/blog/understanding-ecmascript-part-4.md
```diff
@@ -0,0 +1,280 @@
...
+> [Static Semantics: Early Errors](https://tc39.es/ecma262/#sec-delete-operator-static-semantics-early-errors)
+>
+> `UnaryExpression: delete UnaryExpression`
+>
+> - It is a Syntax Error if the `UnaryExpression` is contained in strict mode code and the derived `UnaryExpression` is `PrimaryExpression : IdentifierReference`.
+> - It is a Syntax Error if the derived `UnaryExpression` is
+> `PrimaryExpression : CPEAAPL`
+> and `CPEAAPL` ultimately derives a phrase that, if used in place of `UnaryExpression`, would produce a  Syntax Error according to these rules. This rule is recursively applied.
+
+The first rule forbids `delete IdentifierReference` (for example, `delete foo`) in strict mode. The second rule forbids `CPEAAPL`s which would ultimately produce an `IdentifierReference`, such as `delete (foo)`, `delete ((foo))` and so on.
```
syg, line 225: I feel like this section doesn't add to the understanding of cover grammars, and the article would be tighter if we removed it.

src/blog/understanding-ecmascript-part-4.md
marjakh: I never said it was :) 

I''ll need to clarify this, I'll write my thoughts here.

There are two things:

1) the existence of CPEAAPL enables us to go ahead with parsing and pick the production to follow after having parsed the CPEAAPL

2) after we have the AST (having the AST requires having decided whether we have an ArrowFunction or ParenthesizedExpression), we apply the "covering" rules like you said.

-> How to makes this clearer in the text?

2 is the "cover semantics" but 1 is a needed step for constructing the AST in the first place.

src/blog/understanding-ecmascript-part-4.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Cover grammars
+
+In this episode, we take a deeper look into *cover grammars*. They are a way to specify grammar rules for syntactic constructs where we don't know what we're looking at until we've seen the complete construct.
+
+Again, we'll skip the subscripts for `[In, Yield, Await]` for brevity, as they aren't important for this blog post. See [part 3](/blog/understanding-ecmascript-part-3) for an explanation of their meaning and usage.
+
+## Parenthesized expression or an arrow parameter list?
+
+Typically, parsers decide which grammar production to follow based on finite lookahead.
```
RReverser, line 33: I think it would be useful to show a quick example of simple finite lookahead first (maybe something like the start of an `ExpressionStatement`?), and go into a complex one (that requires cover grammar) after.

src/blog/understanding-ecmascript-part-4.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+The parenthesized whatever-it-is can be arbitrarily long - we cannot know what it is based on a finite amount of tokens.
+
+If the productions were written like this:
+
+```grammar
+AssignmentExpression :
+...
+ConditionalExpression (eventually leading to PrimaryExpression)
```
RReverser, line 59: Can you please check how this renders locally? I suspect `(...)` will be treated as part of the syntax.

RReverser: Also, here and elsewhere the production variants should be indented with 2 spaces (see part 3 Markdown for an example).

src/blog/understanding-ecmascript-part-4.md
marjakh: Removed it

src/blog/understanding-ecmascript-part-4.md
marjakh: Ah, right, I might see what you mean. This sentence is confusing: "Once we know whether we were parsing an `ArrowFunction` or `ParenthesizedExpression`, we need to disallow the corresponding illegal construct"

-> From this one might think that this is done right away, after disambiguating... let's see how I can fix this.

marjakh: I tried to clarify this, by making the section "Using CPEAAPL" only talk about how to disambiguate during parsing, and the section "Restricting CPEAAPLs" only talk about the covering. PTAL

src/blog/understanding-ecmascript-part-4.md
```diff
@@ -0,0 +1,280 @@
...
+
+let x3 = async(a, b) => {};
+let x4 = async();
+```
+
+To this end, the grammar defines a cover grammar symbol `CoverCallExpressionAndAsyncArrowHead` which works similarly to `CPEAAPL`.
+
+## Summary
+
+In this episode we looked into how the spec defines the grammar in such a way that implementing a finite lookahead parser based on it is straightforward.
```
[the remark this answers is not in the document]
marjakh: Modified:

"In this episode we looked into how the spec defines a concise grammar for cases where we cannot identify the current syntactic construct based on a finite lookahead.

In particular, we looked into how the spec uses a cover grammars for first parsing ambiguous-looking constructs permissively and restricting them with static semantic rules later."

src/blog/understanding-ecmascript-part-4.md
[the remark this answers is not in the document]
marjakh: How about this: "In this episode, we take a deeper look into *cover grammars*. They are a way to specify grammar rules for syntactic constructs which look ambiguous based on a finite lookahead."

src/blog/understanding-ecmascript-part-4.md
marjakh: Added spaces, also took the "comment inside grammar" out of the grammar element, so now we have 2 grammar elements here, but prob that's the best we can do.

src/blog/understanding-ecmascript-part-4.md
marjakh: You mean the negative lookahead? I wouldn't want to confuse this blog post with negative lookaheads, as that's not related to cover grammars.

marjakh: ... if I'd like to have an example of a finite positive lookahead, UnaryExpression would be it. But I'm not sure if explaining that brings anything at this point. A more interesting concept would be the "we're not yet know which production to take, but we know the next symbol, so let's parse the next symbol and decide after that" which occurs e.g., for CallExpression, Expression, etc. Not sure if adding that would be beneficial either.

src/blog/understanding-ecmascript-part-4.md
syg: Hmm, I still think "we can decide whether the original `AssignmentExpression` is an `ArrowFunction` or a `ParenthesizedExpression` based on the token following the `CPEAAPL`," is misleading.

How about "in practice, parsers can decide"  instead of "we can decide"?

src/blog/understanding-ecmascript-part-4.md
RReverser: Positive lookahead sounds good too; I just think it's useful to show a simple lookahead example to go with that sentence.

src/blog/understanding-ecmascript-part-4.md
marjakh: I think this is resolved based on the offline discussion.

src/blog/understanding-ecmascript-part-4.md
marjakh: Added lookahead examples.

marjakh:
I uploaded a new version, PTAL.

src/blog/understanding-ecmascript-part-4.md
```diff
@@ -24,15 +24,33 @@ In [part 3](/blog/understanding-ecmascript-part-3), we familiarized ourselves wi
...
+```grammar
+UpdateExpression :
+  LeftHandSideExpression
+  LeftHandSideExpression ++
+  LeftHandSideExpression --
+  ++ UnaryExpression
+  -- UnaryExpression
+```
+
+If we're parsing an `UpdateExpression` and the next token is `++` or `--`, we know the production to use right away. If the next token is neither, it's still not too bad: we can parse a `LeftHandSideExpression` starting from the position we're at, and figure out what to do after we've parsed it. If the token following the `LeftHandSideExpression` is `++`, the production to use is `UpdateExpression : LeftHandSideExpression ++`. The case for `--` is similar. And if the token following the `LeftHandSideExpression` is neither `++` nor `--`, we use the production `UpdateExpression : LeftHandSideExpression`.
```
syg, line 46: Maybe a parenthetical saying that for the those familiar with formal grammars, the grammar is LR(1)? Up to you.

marjakh: I was deliberately leaving it out, afaics it would be inaccurate to say it's LR(1), it's rather "LR(1) with some reasonable extensions" and I don't want to go in the details there.

src/blog/understanding-ecmascript-part-4.md
```diff
@@ -0,0 +1,300 @@
...
+
+Trailing comma and the `...` can occur only in `ArrowParameterList`. Some constructs, like `b = 1` can occur in both, but they have different meanings: Inside `ParenthesizedExpression` it's an assignment, inside `ArrowParameterList` it's a parameter with a default value. Numbers and other `PrimaryExpressions` which are not valid parameter names (or parameter destructuring patterns) can only occur in `ParenthesizedExpression`. But they all can occur inside a `CPEAAPL`.
+
+### Using `CPEAAPL` in productions
+
+Now we can use the very permissive `CPEAAPL` in [`AssignmentExpression` productions](https://tc39.es/ecma262/#prod-AssignmentExpression):
+
+```grammar
+AssignmentExpression :
+  ConditionalExpression
```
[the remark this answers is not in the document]
marjakh: Moved the "note" explaining it up:

(`ConditionalExpression` leads to `PrimaryExpression` via a long production chain.)

PR #400, opened 2020-05-07
Update feature support statuses for different proposals

chicoxyzzy:


src/features/class-fields.md
```diff
@@ -191,14 +191,14 @@ class Cat extends Animal {
 
 <feature-support chrome="72 /blog/v8-release-72#public-class-fields"
                  firefox="partial https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/69#JavaScript"
-                 safari="no"
+                 safari="partial"
```
RReverser, line 194: Are there any links and/or version numbers?

chicoxyzzy: There are many bugzilla issues but only public class fields are available under --useClassFields JSC flag in Technology Preview for now.

RReverser: Hmm so they don't have a meta-bug to track all of these? In that case let's link to one for public class fields where the flag is documented.

RReverser: (same for other features / browsers in the PR)

PR #401, opened 2020-05-07
Up to 4GB of memory in WebAssembly

kripken:


src/blog/4gb-wasm-memory.md
```diff
@@ -0,0 +1,91 @@
...
+The solution is to do an *unsigned* shift, `>>>`:
+
+```javascript
+// This gives us 536870912, as we want!
+console.log((2 * 1024 * 1024 * 1024) >>> 2);
+```
+
+Emscripten knows at compile time whether you may use 2GB or more memory (depending on the flags you use; see later for details). If your flags make 2GB+ addresses possible then the compiler will automatically rewrite all memory accesses to use `>>>` instead of `>>`, which includes not just `HEAP32` etc. accesses as in the examples above but also operations like `.subarray()` and `.copyWithin()`. In other words, the compiler will switch to use unsigned pointers instead of signed ones.
+
+This transformation increases code size a little bit - one extra character in each shift - which is why we don’t do it if you aren’t using 2GB+ addresses. While the difference is typically less than 1%, it’s just unnecessary, and easy to avoid - and lots of small optimizations add up!.
```
jakobkummerow, line 69: nit: s/!./!/ at the end of this line

src/blog/4gb-wasm-memory.md
```diff
@@ -0,0 +1,91 @@
...
+
+Make sure to test on Chrome M83 (currently in Beta) or later. Please file issues if you find anything wrong!
+
+## Conclusion
+
+Support for up to 4GB memory is another step in making the web as capable as native platforms, allowing 32-bit programs to be able to use just as much memory as they would normally. By itself this doesn’t enable a completely new class of application, but it does enable higher-end experiences, such as a very large level in a game or manipulating large content in a graphical editor.
+
+As mentioned earlier, support for 64-bit memory is also planned, which will allow accessing even more than 4GB. However, wasm64 will have the same downside as 64-bit does on native platforms, that pointers take twice as much memory. That’s why 4GB support in wasm32 is so important: We can access twice as much memory while code size remains as compact as wasm has always been!
```
jakobkummerow, line 89: nit: maybe add "as before" after "We can access twice as much memory", just to clarify what the point of comparison is?

kripken:
Thanks @jakobkummerow , I pushed a commit with fixes for your comments now.

Btw when I build locally it looks like @gahaas and I have avatar pictures, but you don't - looks like there's no picture under `src/_img/avatars` yet for you?

RReverser:
>  looks like there's no picture under src/_img/avatars yet for you?

Ping @jakobkummerow - please add your avatar to images. Feel free to ping me or any of other post authors on guidance on resizing if you need any :)

PR #415, opened 2020-06-02
Adding more redirects

camillobruni:
The previous PR did not use regex as identifier in the rewrite rule.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -9,8 +9,13 @@
         "type": 301
       },
       {
-        "source": "/api/(?P<version>v[0-9]+.+)/:api*",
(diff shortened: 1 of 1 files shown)


mathiasbynens:
This broke deployment from CI. I've fixed the build by removing the regexp-based redirects here: 92a4079dbcfc4b89756ccaddb887ec0737f3612b @RReverser could you please help @camillobruni reland this in a way that works? You can use `npm run deploy` to test — it errors when there's something wrong with the config.

mathiasbynens:
@camillobruni No worries! It's a bit tricky since I don't think you can test `npm run deploy` locally, only owners of the Firebase project can. Hopefully @RReverser can take a look.

PR #416, opened 2020-06-04
Add top-level research grant page

RReverser:


src/research-grant.md
```diff
@@ -0,0 +1,22 @@
+---
+permalink: /research-grant/
```
mathiasbynens, line 2: We tend to keep URL slugs brief (`/blog`, `/docs`, `/logo`, `/features`, etc.) so how about just `/grant/`? Just a suggestion, not a blocker, feel free to do whatever you want :)

RReverser: That URL is what @hannespayer asked for, but if he agrees, I don't mind changing to shorter one as well :)

hannespayer: Good point regarding brief URLs. /grant sounds good. Maybe we should add /research at some point to advertise our research investment? Different topic...

RReverser:
Re-requested reviews because I added an item to the navigation pane as per further discussions. Please take another look.

src/_includes/layouts/base.njk
```diff
@@ -28,6 +28,7 @@
           <li{% if '/blog' in page.url %} class="current"{% endif %}><a href="/blog">Blog</a></li>
           <li{% if '/docs' in page.url %} class="current"{% endif %}><a href="/docs">Docs</a></li>
           <li{% if '/features' in page.url %} class="current"{% endif %}><a href="/features" title="JavaScript and WebAssembly language features">JS/Wasm features</a></li>
+          <li{% if '/grant' in page.url %} class="current"{% endif %}><a href="/grant" title="Research">Research</a></li>
```
mathiasbynens, line 31:
```suggestion
          <li{% if '/grant' in page.url %} class="current"{% endif %}><a href="/grant">Research</a></li>
```

mathiasbynens:
LGTM, with some final thoughts:

If the plan is to add a "research" section in the future (per Hannes' comments above), you could also go with `/research` instead of `/grant`. Then again, until now we've been advertising our research investment through blog posts like https://v8.dev/blog/spectre, so maybe we don't need a separate top-level section for it. We could add a `research` tag to the blog.

src/_includes/layouts/base.njk
RReverser: Oops, thanks, I think I just copied&modified the line above.

RReverser:
> If the plan is to add a "research" section in the future (per Hannes' comments above), you could also go with /research instead of /grant. 

As I mentioned above, the idea is to have a separate page for Research that will link out to this one. Which is why I reserved /research with a temporary redirect to /grant for now, but want to make sure that when we add this new page, /grant remains valid and points to what it does now.

PR #418, opened 2020-06-08
Update Code Caching for Devs post section on service worker cache.

wanderview:

PR #420, opened 2020-06-11
Add support for updated field

RReverser:
CI check is a bit trickier, so still in progress, but for now this should at least help with rendering the `updated:` date from frontmatter. In particular, should work well for #418.

### DIFF
--- a/src/_includes/layouts/feature.njk
+++ b/src/_includes/layouts/feature.njk
@@ -10,7 +10,7 @@ layout: layouts/base.njk
         · Updated <time datetime="{{ updated | htmlDateString }}" itemprop="dateModified">{{ updated | readableDate }}</time>
       {% endif %}
       {% if tags %}
-        · tagged with {% for tag in tags %}{% if not 'io' in tag and not 'Node.js' in tag %}
+        · Tagged with {% for tag in tags %}{% if not 'io' in tag and not 'Node.js' in tag %}
           <a href="/features/tags/{{ tag | slug }}" class="tag">{{ tag }}</a>
         {% endif %}{% endfor %}
       {% endif %}
--- a/src/_includes/layouts/feature.njk
+++ b/src/_includes/layouts/feature.njk
@@ -6,6 +6,9 @@ layout: layouts/base.njk
     <h1 itemprop="headline">{{ title | markdown | safe }}</h1>
     <p class="meta">
       Published <time datetime="{{ date | htmlDateString }}" itemprop="datePublished">{{ date | readableDate }}</time>
+      {% if updated %}
+        · Updated <time datetime="{{ updated | htmlDateString }}" itemprop="dateModified">{{ updated | readableDate }}</time>
+      {% endif %}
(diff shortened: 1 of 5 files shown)


src/_includes/layouts/feature.njk
```diff
@@ -6,6 +6,9 @@ layout: layouts/base.njk
     <h1 itemprop="headline">{{ title | markdown | safe }}</h1>
     <p class="meta">
       Published <time datetime="{{ date | htmlDateString }}" itemprop="datePublished">{{ date | readableDate }}</time>
+      {% if updated %}
+        · Updated <time datetime="{{ updated | htmlDateString }}" itemprop="dateModified">{{ updated | readableDate }}</time>
```
mathiasbynens, line 10: lowercase, since it’s no longer the first letter?

src/_includes/layouts/post.njk
```diff
@@ -6,6 +6,9 @@ layout: layouts/base.njk
     <h1 itemprop="headline">{{ title | markdown | safe }}</h1>
     <p class="meta">
       Published <time datetime="{{ date | htmlDateString }}" itemprop="datePublished">{{ date | readableDate }}</time>
+      {% if updated %}
+        · Updated <time datetime="{{ updated | htmlDateString }}" itemprop="dateModified">{{ updated | readableDate }}</time>
```
mathiasbynens, line 10:
```suggestion
        · updated on <time datetime="{{ updated | htmlDateString }}" itemprop="dateModified">{{ updated | readableDate }}</time>
```

RReverser:
Hmm lowercase looks a bit odd, no? Example:

image

It *feels* that " · " should be acting as a sentence separator, although I see now it's not true for `tagged`. Maybe we should sentence-case all of these instead?

image

mathiasbynens:
I guess it's personal preference. To me, the top example looks best, the second feels weird. They're not separate `<p>`s.

Also: can we _not_ show the "updated" part when it's the same as the published date? It doesn't really add any value in that case. (It's fine if you wanna tackle that separately in a future PR.)

RReverser:
> Also: can we not show the "updated" part when it's the same as the published date?

We don't, that was just copy-paste example made via DevTools to demonstrate the difference.

RReverser:
> I guess it's personal preference. To me, the top example looks best, the second feels weird. They're not separate `<p>`s.

No, but even within same `<p>` you can have several sentences. It kind of feels like that's what these are.

As another alternative, we can go all-lowercase:

image

RReverser:
> Hmm lowercase looks a bit odd, no? Example:
> 
> image
> 
> It _feels_ that " · " should be acting as a sentence separator, although I see now it's not true for `tagged`. Maybe we should sentence-case all of these instead?
> 
> image

Btw, this is what we do on web.dev:

image

I'd say let's go with that variant for consistency.

RReverser:
Sentence-cased "Tagged with" too to match other examples.

PR #421, opened 2020-06-12
Update WeakRefs feature to match upcoming web.dev LIVE talk

syg:
Leszek and I wanted to come up with a slightly more compelling example of WeakRefs than the cache example for the web.dev LIVE talk. We came up with one around event handlers.

Since we'll link to this article from the talk, we should update the feature explainer with the new example that's in the talk.

PTAL, thanks!


src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
 }
 ```
 
-But, there’s a problem here. `Map`s hold on to their keys and values strongly, and so the image names and data can never be garbage-collected. This steadily increases memory and causes a **memory leak**!
+We know that keeping all the server messages inside an instance `MovingAvg` uses a lot of memory, so we take care to null out `this.movingAvg` when monitoring is stopped to let the garbage collector reclaim memory.
 
-`WeakRef`s make it possible to solve the memory leak by creating a _weak reference_ to the image and storing _that_ in the cache (instead of the image itself). This way, the garbage collector can clean up the images that do not have a strong reference.
+However, after checking in the memory panel in DevTools, we found out that memory was not being reclaimed at all! The seasoned web developer may have already spotted the bug: event listeners are strong references and must be explicitly removed. Thus, the listener in `MovingAvg` instances, by referencing `this`, keeps the whole instance alive as long as the event listener isn't removed.
+
+Until now, the solution is to manually unregister the event listener via `dispose` method.
```
mathiasbynens, line 103:
```suggestion
Until now, the solution is to manually unregister the event listener via the `dispose` method.
```

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
-    if (deref !== undefined) return deref;
+  dispose() {
+    this.socket.removeEventListener('message', this.listener);
   }
-  const image = performExpensiveOperation(name);
-  const wr = new WeakRef(image);
-  cache.set(name, wr);
-  return image;
+
+  // ...
```
mathiasbynens, line 118: Nit:

```suggestion
  // …
```

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
 }
 ```
 
-But there’s still a problem here: the `Map` still holds on to the `name` strings forever, because those are the keys in the cache. Ideally, those strings would be removed too. The `WeakRef` proposal has a solution for this as well! With the new `FinalizationRegistry` API, we can register a callback to run when the garbage collector zaps a registered object. Such callbacks are known as _finalizers_.
+The downside to this approach is that it is manual memory management. `MovingAvgComponent`, and all other users of the `MovingAvg` class, must remember to call `dispose` or suffer memory leaks. Because the application behavior doesn't depend on the event listener of this diagnostic class, and that the listener is expensive in terms of memory use but not in computation, we really want the listener's lifetime to be logically tied to the `MovingAvg` instance. If its lifetime were tied to the `MovingAvg` instance, then `MovingAvg` could be used like any other JavaScript object whose memory is automatically reclaimed by the garbage collector.
```
mathiasbynens, line 122: “because […] and that the listener” sounds weird to my ESL ears. Maybe something like this?

```suggestion
The downside to this approach is that it is manual memory management. `MovingAvgComponent`, and all other users of the `MovingAvg` class, must remember to call `dispose` or suffer memory leaks. Because the application behavior doesn’t depend on the event listener of this diagnostic class, and the listener is expensive in terms of memory use but not in computation, we really want the listener's lifetime to be logically tied to the `MovingAvg` instance. If its lifetime were tied to the `MovingAvg` instance, then `MovingAvg` could be used like any other JavaScript object whose memory is automatically reclaimed by the garbage collector.
```

Feel free to reject if this is actually correct, of course.

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
-const cache = new Map();
-
-const finalizationRegistry = new FinalizationRegistry((name) => {
-  const ref = cache.get(name);
-  if (ref !== undefined && ref.deref() === undefined) {
-    cache.delete(name);
+class MovingAvg {
+  constructor(socket) {
+    this.events = [];
+    const listener = (ev) => { this.events.push(ev); }
```
mathiasbynens, line 130:
```suggestion
    const listener = (ev) => { this.events.push(ev); };
```

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
+    const weakRef = new WeakRef(listener);
+    const listenerWrapper = (ev) => {
+      const deref = weakRef.deref();
+      if (deref !== undefined) deref(ev);
+    });
+    socket.addEventListener("message", listenerWrapper);
```
mathiasbynens, line 136: Coding style consistency nit:

```suggestion
    socket.addEventListener('message', listenerWrapper);
```

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
   }
-});
+
+  // No more dispose!
```
mathiasbynens, line 139:
```suggestion
  // No more `dispose`!
```

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
+class MovingAvg {
+  constructor(socket) {
+    this.events = [];
+    const listener = (ev) => { this.events.push(ev); }
+    const weakRef = new WeakRef(listener); // 1
+    const listenerWrapper = (ev) => {
+      const deref = weakRef.deref(); // 2
+      if (deref !== undefined) deref(ev); // 3
+    });
+    socket.addEventListener("message", listenerWrapper);
```
mathiasbynens, line 168:
```suggestion
    socket.addEventListener('message', listenerWrapper);
```

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
+We wrap the event listener that does the work in a `WeakRef` to make garbage-collectible, and to not leak its reference to the `MovingAvg` instance via `this` (1). We make a wrapper that `deref` the `WeakRef` to check if it is still alive (2), then call it if so (3). After registering the listener wrapper on the socket, we also register the inner listener on the `FinalizationRegistry`, passing a _holding value_ `{ socket, wrapper }` to the registration (4). Sometime after the inner listener garbage-collected, the finalizer may run, with the holding value passed to it. Inside the finalizer, we unregister the listener wrapper as well, making all memory associated with a `MovingAvg` instance garbage-collectible (5).
```
mathiasbynens, line 175:
```suggestion
We wrap the event listener that does the work in a `WeakRef` to make garbage-collectible, and to not leak its reference to the `MovingAvg` instance via `this` (1). We make a wrapper that `deref` the `WeakRef` to check if it is still alive (2), then call it if so (3). After registering the listener wrapper on the socket, we also register the inner listener on the `FinalizationRegistry`, passing a _holding value_ `{ socket, wrapper }` to the registration (4). Sometime after the inner listener gets garbage-collected, the finalizer may run, with the holding value passed to it. Inside the finalizer, we unregister the listener wrapper as well, making all memory associated with a `MovingAvg` instance garbage-collectible (5).
```

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
+
+With all this, our original implementation of `MovingAvgComponent` neither leaks memory nor requires any manual disposal.
 
 ## Don’t overdo it { #progressive-enhancement }
 
 After hearing about these new capabilities, it might be tempting to `WeakRef` All The Things™. However, that’s probably not a good idea. Some things are explicitly _not_ good use cases for `WeakRef`s and finalizers.
 
-In general, avoid writing code that depends on the garbage collector cleaning up a `WeakRef` or calling a finalizer at any predictable time — [it can’t be done](https://github.com/tc39/proposal-weakrefs#a-note-of-caution)!
+In general, avoid writing code that depends on the garbage collector cleaning up a `WeakRef` or calling a finalizer at any predictable time — [it can’t be done](https://github.com/tc39/proposal-weakrefs#a-note-of-caution)! Moreover, whether an object is garbage-collectible at all may depend on implementation details, such as the representation of closures, that are both subtle and may differ between different versions of V8.
```
mathiasbynens, line 183: Let’s not just talk about V8 here but also mention different JavaScript engines. Maybe something like:

```suggestion
In general, avoid writing code that depends on the garbage collector cleaning up a `WeakRef` or calling a finalizer at any predictable time — [it can’t be done](https://github.com/tc39/proposal-weakrefs#a-note-of-caution)! Moreover, whether an object is garbage-collectible at all may depend on implementation details, such as the representation of closures, that are both subtle and may differ between different versions of V8 and across JavaScript engines.
```

mathiasbynens:
Thanks for the rewrite! LGTM with some nits/suggestions. Given that this is a substantial rewrite of the explainer, feel free to add yourselves as authors.

> Leszek and I wanted to come up with a slightly more compelling example of WeakRefs than the cache example for the web.dev LIVE talk. We came up with one around event handlers.

For background, Sathya and I came up with the old example [for our Google I/O ’19 presentation](https://www.youtube.com/watch?v=c0oy0vQKEZE&t=32m28s).

src/features/weak-references.md
mathiasbynens: For reference, we try to follow [the Google JS style guide](https://google.github.io/styleguide/jsguide.html) for any public-facing examples in blog posts and slides, which means: use semicolons, use single quotes instead of double quotes for string literals, prefer `const` over `let`, etc.

src/features/weak-references.md
marjakh, line 136: Doesn't this still leak the listenerWrapper? The socket doesn't know that the weakref died, and doesn't remove listenerWrapper.

In general, aren't the weak refs "off by one layer" here? The eventlistener list should be a list of weakrefs, so that we can clean up the eventlisteners which are no longer relevant.

Or did you mean the cleanup would be done by listenerwrapper when it gets an event and it notices its weakref has expeired?

src/features/weak-references.md
LeszekSwirski: I would say "via **a** dispose method"

src/features/weak-references.md
LeszekSwirski: I would split the sentence differently, e.g.

> The application behavior doesn't depend on the event listener of this diagnostic class, and the listener is expensive in terms of memory use but not in computation. What we really want is for the listener's lifetime to be logically tied to the `MovingAvg` instance, so that `MovingAvg` could be used like any other JavaScript object whose memory is automatically reclaimed by the garbage collector.

src/features/weak-references.md
LeszekSwirski: It does indeed leak the listenerWrapper, which is the point of the subsequent finalization registry note.

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
-  if (ref !== undefined && ref.deref() === undefined) {
-    cache.delete(name);
+class MovingAvg {
+  constructor(socket) {
+    this.events = [];
+    const listener = (ev) => { this.events.push(ev); }
+    const weakRef = new WeakRef(listener);
+    const listenerWrapper = (ev) => {
+      const deref = weakRef.deref();
+      if (deref !== undefined) deref(ev);
```
LeszekSwirski, line 134: Explicit for ease-of-reading? I did kinda like the optional chain function call for this.

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
-  const ref = cache.get(name);
-  if (ref !== undefined && ref.deref() === undefined) {
-    cache.delete(name);
+class MovingAvg {
+  constructor(socket) {
+    this.events = [];
+    const listener = (ev) => { this.events.push(ev); }
+    const weakRef = new WeakRef(listener);
+    const listenerWrapper = (ev) => {
+      const deref = weakRef.deref();
```
LeszekSwirski, line 133: nit: call this "listener", and we could even call the weakRef "weakListener"

src/features/weak-references.md
mathiasbynens: Right, that's better indeed! Edited the suggestion

src/features/weak-references.md
marjakh: Ah, right, I was confused by the "No more dispose!" below which I took to mean "dispose is not needed now, yay". Which is confusing, like, it's not needed in the first place if leaking is OK...

marjakh: (It might be less confusing to still have the dispose here because we still need to remove the event listener, like in the previous version. But I realize it is probably way too late to fix this, so pls feel free to ignore this comment.)

src/features/weak-references.md
RReverser: I'd remove the braces altogether since the return value doesn't matter:

```suggestion
    const listener = (ev) => this.events.push(ev);
```

mathiasbynens: Interestingly, for the exact same reason I would keep the braces :) (Only `return` something non-`undefined` when you need it.)

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
+  constructor(socket) {
+    this.events = [];
+    const listener = (ev) => { this.events.push(ev); }
+    const weakRef = new WeakRef(listener); // 1
+    const listenerWrapper = (ev) => {
+      const deref = weakRef.deref(); // 2
+      if (deref !== undefined) deref(ev); // 3
+    });
+    socket.addEventListener("message", listenerWrapper);
+    gListenersRegistry.register(listener,
```
RReverser, line 169: Nitpick: such split, while common in C / C++, looks a bit odd in typical JS. I'd either put it on the same line at the risk of running over max-length

```js
gListenersRegistry.register(listener, { socket, listenerWrapper });
```

or use block-style formatting

```js
gListenersRegistry.register(listener, {
  socket,
  listenerWrapper
});
```

src/features/weak-references.md
RReverser: Heh, usually when the caller expects `undefined`, I just return any value.

I know what the ultimate solution would be 😀
```suggestion
    const listener = (ev) => void this.events.push(ev);
```

src/features/weak-references.md
```diff
@@ -54,104 +54,133 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
-
-const finalizationRegistry = new FinalizationRegistry((name) => {
-  const ref = cache.get(name);
-  if (ref !== undefined && ref.deref() === undefined) {
-    cache.delete(name);
+class MovingAvg {
+  constructor(socket) {
+    this.events = [];
+    const listener = (ev) => { this.events.push(ev); }
+    const weakRef = new WeakRef(listener);
```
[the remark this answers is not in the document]
syg: Ah, doh, this is difficult to get right. I wonder how to keep it alive without making it uncollectible...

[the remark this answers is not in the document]
syg: FWIW since it's part of the WeakRef, at least it'll be guaranteed to stay alive for *1* turn. 😅

src/features/weak-references.md
syg: I was trying to stay period accurate to the time the original post was written.

RReverser: Since it's going to be updated anyway, I'd change to the new syntax. Just make sure to add `updated:` date to frontmatter, which will be rendered once #420 is merged.

src/features/weak-references.md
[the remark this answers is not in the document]
jridgewell: I think you need to `WeakRef(this)`:

```js
class MovingAvg {
  constructor(socket) {
    this.events = [];
    this.listener = (ev) => { this.events.push(ev); }

    const weakRef = new WeakRef(this);
    const listenerWrapper = (ev) => {
      const deref = weakRef.deref();
      if (deref !== undefined) deref.listener(ev);
    });
    socket.addEventListener("message", listenerWrapper);
  }
}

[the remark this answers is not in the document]
Comment (syg) [reply src/features/weak-references.md:131]: That's trickier to reason through. I plan to update with the following, which is easier to understand and side-steps the non-split function context subtlety:

```js
function makeWeakListener(listener) {
  const weakRef = new WeakRef(listener);
  const wrapper = (ev) => { weakRef.deref()?.(ev); };
  return wrapper;
}

class MovingAvg {
  constructor(socket) {
    this.events = [];
    this.listener = (ev) => { this.events.push(ev); };
    socket.addEventListener("message", makeWeakListener(this.listener));
  }
}
```

As long as you make the wrapper closure outside of the lexical scope of the constructor itself, you can sidestep the trickiness of accidentally entraining variables in the function context.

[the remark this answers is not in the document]
syg: To further clarify why `WeakRef(this)` doesn't work: once `FinalizationRegistry` comes into the picture, we need to pass the wrapper listener to it as a holding value so the finalizer can remove the event listener. That wrapper is strongly referenced as the holdings of the FR. If that wrapper is created inside the lexical scope of the constructor, it'll end up entraining all the other closed over variables. Since `this.listener` is an arrow and closes over `this`, the wrapper thus keeps the `MovingAvg` instance alive. It gets worse: since that `MovingAvg` instance is alive and that instance has a `this.listener` property, that arrow is also kept alive by the FR.

Unfortunately, function contexts is a pretty big footgun that devs will need to understand to use WeakRefs and FRs correctly with functions.

src/features/weak-references.md
syg: Made one line.

src/features/weak-references.md
syg: I refactored the wrapping and making of the new WeakRef into a separate function this hopefully is less confusing?

src/features/weak-references.md
syg: Removed the comment that said "No more dispose!"

src/features/weak-references.md
syg: Adopted.

src/features/weak-references.md
syg: Done.

src/features/weak-references.md
syg: Done.

src/features/weak-references.md
syg: Done.

src/features/weak-references.md
syg: Added semicolon.

src/features/weak-references.md
syg: Done.

src/features/weak-references.md
syg: Changed to "" everywhere.

src/features/weak-references.md
syg: Done.

src/features/weak-references.md
syg: Done.

src/features/weak-references.md
syg: Oops, changing it to '' everywhere.

syg:
Ping for more review. Hard deadline of July 1, when the talk goes live.

RReverser:
@syg Can you "Resolve conversation" on comments which have been addressed? It's not very clear what issues remain here / who you're waiting on.

If all the comments have been addressed, please bump the `updated` date again, probably to `2020-06-18` so that I could merge it tomorrow.

RReverser:
Looks like there is also some issue in CI logs (click "Details" on checks) with Yaml syntax...

syg:
I've marked most of the issues as resolved. Mainly waiting for another once-over from @LeszekSwirski. 

Not sure what the YAML error is, maybe I pushed it before the `updated:` PR was merged. I pushed with an update to the `updated:` frontmatter.

RReverser:
Idk, looks like it's not about `updated:`: https://travis-ci.com/github/v8/v8.dev/jobs/350577124#L269-L274

src/features/weak-references.md
```diff
@@ -1,10 +1,13 @@
 ---
 title: 'Weak references and finalizers'
-author: 'Sathya Gunasekaran ([@_gsathya](https://twitter.com/_gsathya)) and Mathias Bynens ([@mathias](https://twitter.com/mathias))'
+author: 'Sathya Gunasekaran ([@_gsathya](https://twitter.com/_gsathya)), Mathias Bynens ([@mathias](https://twitter.com/mathias)), Shu-yu Guo ([@_shu](https://twitter.com/_shu)), and Leszek Swirski ([@leszekswirski](https://twitter.com/leszekswirski))
```
RReverser, line 3: @syg This is probably the reason - someone removed the closing quote :)

syg: That someone would be this doofus. :)

gsathya:
Not sure if I'm missing something obvious but why is this overwriting (as opposed to appending) the previous example? Was that incorrect?

syg:
> Not sure if I'm missing something obvious but why is this overwriting (as opposed to appending) the previous example? Was that incorrect?

No, it wasn't incorrect. Leszek and I wanted a more compelling example. I assumed that having two examples would be confusing for folks coming from links on the video. I'm happy with appending as well. @RReverser WDYT?

src/features/weak-references.md
```diff
@@ -54,104 +57,138 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
-function getImageCached(name) {
-  const ref = cache.get(name); // 1
-  if (ref !== undefined) { // 2
-    const deref = ref.deref();
-    if (deref !== undefined) return deref;
+We can register a callback with a `FinalizationRegistry` to remove `listenerWrapper` from the socket when the inner event listener is garbage-collected. Our final implementation looks like this:
+
+```js
+const gListenersRegistry = new FinalizationRegistry(
+  ({ socket, wrapper }) => { socket.removeEventListener(socket, wrapper); }); // 6
```
marjakh, line 165: Should the first param to removeEventListener be "message" instead of socket?

mathiasbynens:
Now that Sathya mentions it, keeping the original example and adding the new example as a new section seems strictly better IMHO.

LeszekSwirski:
To be honest, I never much liked the previous example -- it served the purpose of explaining the feature, but I generally don't think we should encourage WeakRefs for caching. To me, this multi-parenting example presents a much more convincing need for a weak-ref feature.

src/features/weak-references.md
```diff
@@ -54,104 +57,138 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
+    // ...
+  }
 }
 ```
 
-But, there’s a problem here. `Map`s hold on to their keys and values strongly, and so the image names and data can never be garbage-collected. This steadily increases memory and causes a **memory leak**!
+We know that keeping all the server messages inside an instance `MovingAvg` uses a lot of memory, so we take care to null out `this.movingAvg` when monitoring is stopped to let the garbage collector reclaim memory.
 
-`WeakRef`s make it possible to solve the memory leak by creating a _weak reference_ to the image and storing _that_ in the cache (instead of the image itself). This way, the garbage collector can clean up the images that do not have a strong reference.
+However, after checking in the memory panel in DevTools, we found out that memory was not being reclaimed at all! The seasoned web developer may have already spotted the bug: event listeners are strong references and must be explicitly removed. Thus, the listener in `MovingAvg` instances, by referencing `this`, keeps the whole instance alive as long as the event listener isn't removed.
```
LeszekSwirski, line 104: This explanation might be better with explicit diagrams, e.g. https://drive.google.com/file/d/1Xca-f4PoieduOjoDMiOB2or16cUbI3u0/view?usp=sharing

src/features/weak-references.md
```diff
@@ -54,104 +57,138 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
+    this.listener = (ev) => { this.events.push(ev); };
+    addWeakListener(socket, this.listener);
   }
-});
+}
 ```
 
 :::note
-**Note:** The `ref !== undefined && ref.deref() === undefined` is required because we could’ve added a new `WeakRef` with the same `name` between the old `WeakRef` enqueueing the finalization callback and actually running the finalization callback.
+**Note:** `WeakRef`s to functions must be treated with caution. JavaScript functions are [closures](https://en.wikipedia.org/wiki/Closure_(computer_programming)) and strongly reference the outer environments which contain the values of free variables referenced inside the functions. These outer environments may contain variables that _other_ closures reference as well. That is, when dealing with closures, their memory is often strongly referenced by other closures in subtle ways. Keep this in mind when writing code.
```
LeszekSwirski, line 146: Say explicitly that this is why we have a separate `addWeakListener` function, and that it's explicitly not local to the MovingAvg constructor to not be part of its lexical scope.

src/features/weak-references.md
```diff
@@ -54,104 +57,138 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
+:::note
+**Note:** The finalization callback does not run immediately after garbage-collecting the event listener. It either runs at some point in the future, or not at all — the spec doesn’t guarantee that it runs! Keep this in mind when writing code.
+:::
 
-function getImageCached(name) {
-  const ref = cache.get(name); // 1
-  if (ref !== undefined) { // 2
-    const deref = ref.deref();
-    if (deref !== undefined) return deref;
+We can register a callback with a `FinalizationRegistry` to remove `listenerWrapper` from the socket when the inner event listener is garbage-collected. Our final implementation looks like this:
```
LeszekSwirski, line 161: nit: wrapper

mathiasbynens:
> I generally don't think we should encourage WeakRefs for caching

Could you elaborate on this?

src/features/weak-references.md
```diff
@@ -54,104 +57,138 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
+
+```js
+const gListenersRegistry = new FinalizationRegistry(
```
LeszekSwirski, line 164: Is it worth mentioning that gListenersRegistry here is global, because the lifetime of the registry is independent of the lifetime of the weakrefs?

src/features/weak-references.md
```diff
@@ -54,104 +57,138 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
 
-const finalizationRegistry = new FinalizationRegistry((name) => {
-  const ref = cache.get(name);
-  if (ref !== undefined && ref.deref() === undefined) {
-    cache.delete(name);
-  }
-});
+Since the event listener is wrapped in a `WeakRef`, the _only_ strong reference to it is the `listener` property on the `MovingAvg` instance. That is, we've successfully tied the lifetime of the event listener to the lifetime of the `MovingAvg` instance.
+
+But there's still a problem here: we've added a level of indirection to `listener` by wrapping it a `WeakRef`, but the wrapper in `addWeakListener` is still leaking for the same reason that `listener` was leaking originally. The solution to this is the companion feature to `WeakRef`, `FinalizationRegistry`. With the new `FinalizationRegistry` API, we can register a callback to run when the garbage collector zaps a register object. Such callbacks are known as _finalizers_.
```
LeszekSwirski, line 155: Is it worth noting explicitly that the leak isn't as bad as before, because we're only leaking the one closure, not the whole MovingAvg class?

src/features/weak-references.md
```diff
@@ -54,104 +57,138 @@ A JavaScript `WeakMap` is not really _weak_: it actually refers _strongly_ to it
...
-
-:::note
-**Note:** The finalization callback does not run immediately after garbage-collecting the image object. It either runs at some point in the future, or not at all — the spec doesn’t guarantee that it runs! Keep this in mind when writing code.
-:::
+The downside to this approach is that it is manual memory management. `MovingAvgComponent`, and all other users of the `MovingAvg` class, must remember to call `dispose` or suffer memory leaks. The application behavior doesn't depend on the event listener of this diagnostic class, and the listener is expensive in terms of memory use but not in computation. What we really want is for the listener's lifetime to be logically tied to the `MovingAvg` instance, so that `MovingAvg` could be used like any other JavaScript object whose memory is automatically reclaimed by the garbage collector.
```
LeszekSwirski, line 125: Another downside I just thought of -- if MovingAvgComponent is collected before stopping, then there's no one that could call dispose.

LeszekSwirski:
> > I generally don't think we should encourage WeakRefs for caching
> Could you elaborate on this?

Sure. A good cache has multiple properties, and "not holding objects strongly" is only one of them. In particular, WeakRefs are cleared after _every_ (major) GC, and have no measure of recency or usefulness of that cache, nor is there a concept of clearing them only under memory pressure (like Java SoftReferences) rather than on "normal" app behaviour. Even worse, for caching "large things", I'd expect more application GCs and so more clearing of otherwise useful caches. I'd rather encourage devs to use something smarter, like an LRU cache, manual cache budgeting, "frecency" measures, that sort of stuff. Ideally we'd even provide some sort of a keyed cache API, with some sort of memory budget (relative or absolute) and a callback to decide which entry to delete, but WeakRefs are a very incomplete solution to this problem.

We do tie some internal caches and cache-like behaviours in V8 to GCs (although e.g. bytecode is only flushed after 5 GCs, not immediately after the first one), but it's problematic due to GC non-determinism, we'd like to get rid of it, and we can only really get away with it in the first place because we can at least make _some_ assumptions about GC behaviour (e.g. not killing weak refs on scavenges, despite them being "a GC").

gsathya:
> In particular, WeakRefs are cleared after every (major) GC

They don't have to be. Nothing in the spec forces this. The engine can be smarter about everything you've mentioned here.

LeszekSwirski:
> > In particular, WeakRefs are cleared after every (major) GC
> They don't have to be. Nothing in the spec forces this. The engine can be smarter about everything you've mentioned here.

Sure, but they currently are, and nothing in the spec prevents this either. The spec doesn't really technically require them to be collected at all. That's a weakness (pun not indended) for using them for caching, not a strength -- they're simply a different tool more applicable to a different task.

LeszekSwirski:
Moreover, if people start using them more heavily for caching, then we'd likely want to start optimising their behaviour to be more cache-like, which would mean a) making them behave differently than in-engine weak refs, and b) making them _less_ suitable for tasks like the one in this PR, leaving devs with once again no good way of resolving multi-parent leak issues.

gsathya:
> That's a weakness (pun not indended) for using them for caching, not a strength

I mean, we use weakrefs extensively for inline caches in v8, clearly they provide value for this use case. Of course they come with shortcomings -- we need to age maps differently if they are in embedded code objects but trade offs are true for any design. I suppose our example can (and probably should) have a description of the shortcomings but it's not wrong to use a weakref based cache. 

> making them less suitable for tasks

Why?

LeszekSwirski:
Inline caches, at least Load/StoreICs, are actually an example of the multi-parent semantics, not "cached result" semantics; we don't want feedback vectors to keep maps alive if those maps are otherwise dead. A better example for your argument is the weak links in transition trees, and those being weak actually causes problems for exactly the reasons I gave; plus, those links predominantly rely on there being _other_ objects around with a strong ref to those maps, unlike the image cache in the current article. Another example is, as I mentioned, bytecode flushing, where we explicitly having GC-epoch aging rather than a weak ref because a weak ref isn't a sufficiently good cache; and even there it still has problems.

Why would making weak refs more "cache-y" make them less suitable for the sort of tasks like in this PR? Because we want `MovingAvg` to be collected pretty much ASAP once it's disconnected from the reachability graph, and a "cache-y" weak ref would keep it alive longer than necessary. It would still work, semantically, but it wouldn't be _as_ suitable as it is at the moment, and there wouldn't exist a more suitable alternative. Not to mention that an application relying on a weak reference to have cache-y behaviour is _very_ implementation specific, even more so than relying on e.g. FinalizationRegistry behaviour.

I'm not saying that the cache example is wrong, it definitely works in the sense that it won't break -- what I'm saying is that it won't be a particularly good cache for weak refs as they exist in current implementations, and it shouldn't be something we present as a canonical example of weak ref use.

gsathya:
> Inline caches, at least Load/StoreICs, are actually an example of the multi-parent semantics, not "cached result" semantics; we don't want feedback vectors to keep maps alive if those maps are otherwise dead. A better example for your argument is the weak links in transition trees

Eh, I think it depends on your mental model -- it's the way around for me. You do want to keep maps alive even if "parent" object is dead because you don't want to deopt. 

>  being other objects around with a strong ref to those maps, unlike the image cache in the current article.

Because this happens to be a very basic example attempting to provide an overview of how to use this feature. It's not hard to imagine a use case where there would be other links to the image.

> It would still work, semantically, but it wouldn't be as suitable as it is at the moment, and there wouldn't exist a more suitable alternative

It would just get collected a bit later which seems fine to me. 

> Not to mention that an application relying on a weak reference to have cache-y behaviour is very implementation specific, even more so than relying on e.g. FinalizationRegistry behaviour.

This is true for every use of weak ref because of how vague the spec is.

> what I'm saying is that it won't be a particularly good cache for weak refs as they exist in current implementations, and it shouldn't be something we present as a canonical example of weak ref use.

I think it's a fine example. In any case, the right way to educate users would be to explain the use case along with the tradeoffs. Given that the first iteration of the use case in this PR actually leaked the listener, I'm not sure there is any use of weak ref that doesn't have trade offs or shortcomings.

Anyways, I understand (but don't agree with) the motivation for this change -- so I'm fine with landing this.

LeszekSwirski:
Taking a step back, I think we simply have two interpretations of what a weak ref should be:

1. A reference that would prefer to be able to access an object, but is ok if that object goes away
2. A reference that wants to be able to access an object while it exists, but no longer than that

The difference is subtle, but the caching example used them in the former way, while the new example uses them in the latter way.

The spec is vague enough that both are valid interpretations, and I guess the conflict is because to me they're "obviously" (but I guess not actually obviously) only the latter, while as far as I can tell you're happy to use them as the former too.

Certainly V8 implements them as the latter; to me that's the "natural" implementation, and I'd expect other engines to do the same, but as you say it's not necessarily the case, and an engine could implement them with cache-behaviour specific optimisations.

The problem that I have with a caching example is that it's simply not a good cache if the implementation doesn't make it a good cache. A good caching WeakRef implementation would track recent accesses to weak ref, likely track the frequency of those accesses, preferably would even consider the retaining graph size of those references, and, most importantly, would pick and choose which references to clear based on that tracking. If there isn't logic to choose which weak refs to clear, I'd think it's reasonable to assume _all_ otherwise dead weak refs are cleared on each GC -- and if they're all cleared, then most of your cache dies on the first GC, which I think makes for an ineffective cache. It's an implementation assumption (even the fact that there's a GC at all is an implementation assumption), but I consider it a simpler assumption than assuming WeakRefs to be good at caching.

Anyway, since you're ok with landing this CL, I don't want to argue much more about the benefits of either interpretation, I just wanted you to know that I appreciate what point of view you're coming from, and try to explain where I stand relative to it.

> Given that the first iteration of the use case in this PR actually leaked the listener, I'm not sure there is any use of weak ref that doesn't have trade offs or shortcomings.
I wanted to call this out as an unfair comment -- differences in opinion of what WeakRefs are for aside, there's a difference between bugs in an implementation and inherent tradeoffs/shortcomings of an approach.

gsathya:
> Taking a step back, I think we simply have two interpretations of what a weak ref should be:
> 
> 1. A reference that would prefer to be able to access an object, but is ok if that object goes away
> 2. A reference that wants to be able to access an object while it exists, but no longer than that
> 

I think this captures the differences clearly. Thank you -- this has helped me understand and appreciate your point of view much better as well. 

> I wanted to call this out as an unfair comment -- differences in opinion of what WeakRefs are for aside, there's a difference between bugs in an implementation and inherent tradeoffs/shortcomings of an approach.

The implementation complexity of an approach (leading to bugs) is an inherent trade off/shortcoming -- that's what I was alluding to. I think I explained that poorly.

syg:
I'd like to say the exchange between @LeszekSwirski and @gsathya is one of the most well-reasoned and informative exchanges I've seen in recent memory. Bravo on such a high level of discourse.

RReverser:
> I'd like to say the exchange between @LeszekSwirski and @gsathya is one of the most well-reasoned and informative exchanges I've seen in recent memory. Bravo on such a high level of discourse.

It's tempting to encourage them to write up a separate blog post, in form of a dialogue, about WeakRefs, its use-cases and implementation details :) I'm sure lots of others would find it illuminating.

src/features/weak-references.md
syg: Ah oops, nice catch. Will fix.

src/features/weak-references.md
syg: Oh wow, thanks for the diagrams!

src/features/weak-references.md
syg: Will add.

src/features/weak-references.md
syg: Yeah, good call. Seems worth a note.

src/features/weak-references.md
syg: Will clarify.

src/features/weak-references.md
syg: I'll add this sentence: What's worse, manual memory management is cascading: users of `MovingAvgComponent` must remember to call `stop` or suffer memory leaks, so on and so forth.

syg:
This should be good to go. I've modified and added the excellent diagrams from @LeszekSwirski.

mathiasbynens:
🎉  Super happy to see this long discussion resolved. What a ride! +1 to @RReverser's gentle hint to maybe turn this discussion into a blog post of its own.

Before merging, let's run the SVGs through https://jakearchibald.github.io/svgomg/ if possible.

LeszekSwirski:
I applied a bit of svgomg, and a bit of that Leszek hand-tuning SVG magic to the SVGs.

RReverser:
Just to be clear - @gsathya and @mathiasbynens - are you ok with overriding your previous example?

I agree that it's better to have just have one example to avoid confusing readers, but since this is supposed to be an update to a previous blog post, I'd want to make sure that original authors are ok with their work being overridden.

syg:
Thanks for optimizing my SVGs, @LeszekSwirski!

RReverser:
> Just to be clear [...]

@gsathya and @mathiasbynens - ping - is this good to go in the current form?

mathiasbynens:
@RReverser I'd already LGTM'd the latest version (modulo the SVG changes I requested), so still LGTM from my side. Sathya, WDYT?

gsathya:
As I said before I don't like this but I don't care enough to block it. LGTM

RReverser:
Ok, I'll go ahead and merge it...

PR #427, opened 2020-06-23
Add note to add blink CQ bot when shipping wasm features

backes:
With the last two features we shipped, we accidentally broke the blink web tests, causing a delay in the v8 roll.
This adds a note to the shipping document to hopefully avoid this in the future.

Drive-by: Update the link to the `wasm-feature-flags.h` file (it will redirect there anyway).

### DIFF
--- a/src/docs/wasm-shipping-checklist.md
+++ b/src/docs/wasm-shipping-checklist.md
@@ -43,6 +43,7 @@ See the [staging of type reflection](https://crrev.com/c/1771791) as a reference
 
 ## How to ship a WebAssembly Feature
 
-- In [`src/wasm/wasm-feature-flags.h`](https://cs.chromium.org/chromium/src/v8/src/wasm/wasm-feature-flags.h), move the feature flag from the `FOREACH_WASM_STAGING_FEATURE_FLAG` macro list to the `FOREACH_WASM_SHIPPED_FEATURE_FLAG` macro list.
+- In [`src/wasm/wasm-feature-flags.h`](https://source.chromium.org/chromium/chromium/src/+/master:v8/src/wasm/wasm-feature-flags.h), move the feature flag from the `FOREACH_WASM_STAGING_FEATURE_FLAG` macro list to the `FOREACH_WASM_SHIPPED_FEATURE_FLAG` macro list.
+  - Ensure to add a blink CQ bot on the CL to check for [blink web test](https://v8.dev/docs/blink-layout-tests) failures caused by enabling the feature.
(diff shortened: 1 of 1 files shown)


backes:
@gahaas 
Andreas, can you take a look please? This might save us from some reverts :)

src/docs/wasm-shipping-checklist.md
```diff
@@ -43,6 +43,7 @@ See the [staging of type reflection](https://crrev.com/c/1771791) as a reference
 
 ## How to ship a WebAssembly Feature
 
-- In [`src/wasm/wasm-feature-flags.h`](https://cs.chromium.org/chromium/src/v8/src/wasm/wasm-feature-flags.h), move the feature flag from the `FOREACH_WASM_STAGING_FEATURE_FLAG` macro list to the `FOREACH_WASM_SHIPPED_FEATURE_FLAG` macro list.
+- In [`src/wasm/wasm-feature-flags.h`](https://source.chromium.org/chromium/chromium/src/+/master:v8/src/wasm/wasm-feature-flags.h), move the feature flag from the `FOREACH_WASM_STAGING_FEATURE_FLAG` macro list to the `FOREACH_WASM_SHIPPED_FEATURE_FLAG` macro list.
+    - Ensure to add a blink CQ bot on the CL to check for [blink web test](https://v8.dev/docs/blink-layout-tests) failures caused by enabling the feature (add this line to the footer of the CL description: `Cq-Include-Trybots: luci.v8.try:v8_linux_blink_rel`).
```
RReverser, line 47: Indentation looks odd, is this meant to be a sublist with a single item?

backes: Yes. First I only added two dots, but the linter complained and demands four spaces.
We have the same thing in line 14-16 above.

The previous looks good. Do you see a problem with this?

RReverser: Not really, just wanted to make sure this is intentional, because from text itself it doesn't sound like a sub-item of the parent.

backes: Oh, I see. Would it help to rewrite the first item as "Create a CL to move ..."?
@gahaas WDYT?

gahaas:
Should we add a description of what to do if the blink bots fail?

backes:
> Should we add a description of what to do if the blink bots fail?

There is already a link to https://v8.dev/docs/blink-layout-tests in the same line. I would have said that's enough? (That page explains what to do.)

PR #428, opened 2020-06-24
Add redirect rule

camillobruni:
Avoiding the colon by omitting `http:` might help with the regexp redirect pattern issues.


RReverser:
I doubt this is it, we have `https:` in all other redirect destinations too. The error specifically complains about source or regex pattern part, not the destination.

PR #437, opened 2020-07-03
Clarify expectations around finalizer callbacks

jakearchibald:
When I read this article I couldn't make sense of the certainty of:

> With all this, our original implementation of MovingAvgComponent neither leaks memory nor requires any manual disposal.

vs the vagueness of:

> It either runs at some point in the future, or not at all — the spec doesn’t guarantee that it runs!

It didn't feel that both could be true. Either the solution "might" work, or we can assume that finalizer callbacks will run.

The "progressive enhancement" comparison didn't help, as typically PE is used when something has a reasonable chance of not-working.

I've tried to clarify that, but I'm not 100% sure I understand the feature


src/features/weak-references.md
```diff
@@ -217,8 +217,6 @@ After hearing about these new capabilities, it might be tempting to `WeakRef` Al
 
 In general, avoid writing code that depends on the garbage collector cleaning up a `WeakRef` or calling a finalizer at any predictable time — [it can’t be done](https://github.com/tc39/proposal-weakrefs#a-note-of-caution)! Moreover, whether an object is garbage-collectible at all may depend on implementation details, such as the representation of closures, that are both subtle and may differ across JavaScript engines and even between different versions of the same engine.
 
-For example, don’t place important logic in the code path of a finalizer. There’s no way to predict _when_, or even _if_, a given finalizer gets called. It’s best to think of `WeakRef`s and finalizers as **progressive enhancement**: it’s nice if your custom finalizer code runs, but your program should still work without it.
```
LeszekSwirski, line 220: I think it's good to keep something like this, but say something closer to "it's not guaranteed to run if e.g. you close your browser window", and that it should be used only for cleanup and not for business logic (e.g. a dependable signal POSTed back to the server that some object was collected, as you allude to with "metrics" above).

I can't quite think of a way to phrase it, but basically the dev has to make sure that the they don't rely on the finalizer more than on the GC.

src/features/weak-references.md
```diff
@@ -177,7 +177,7 @@ Eventually, after a garbage collection occurs, the `MovingAvg` instance and the
...
 
 :::note
-**Note:** The finalization callback does not run immediately after garbage-collecting the event listener. It either runs at some point in the future, or not at all — the spec doesn’t guarantee that it runs! Keep this in mind when writing code.
+**Note:** The finalization callback does not run immediately after garbage-collecting the event listener, so don't use it for important logic or metrics. The timing of garbage collection and finalization callbacks is unspecifed. In fact, an engine that never garbage collected would be fully compliant. However, it's safe to assume that engines will garbage collect, and finalization callbacks will be called at _some point_. Keep this uncertainty in mind when writing code.
```
LeszekSwirski, line 180: "and finalization callbacks will be called at _some point_" is a dangerous suggestion to make, since they might not run if there's no need to worry about memory, e.g. the tab is closed.

jakearchibald:
I've taken another swing at it

src/features/weak-references.md
```diff
@@ -217,7 +217,7 @@ After hearing about these new capabilities, it might be tempting to `WeakRef` Al
 
 In general, avoid writing code that depends on the garbage collector cleaning up a `WeakRef` or calling a finalizer at any predictable time — [it can’t be done](https://github.com/tc39/proposal-weakrefs#a-note-of-caution)! Moreover, whether an object is garbage-collectible at all may depend on implementation details, such as the representation of closures, that are both subtle and may differ across JavaScript engines and even between different versions of the same engine.
 
-For example, don’t place important logic in the code path of a finalizer. There’s no way to predict _when_, or even _if_, a given finalizer gets called. It’s best to think of `WeakRef`s and finalizers as **progressive enhancement**: it’s nice if your custom finalizer code runs, but your program should still work without it.
+For example, don’t place important logic in the code path of a finalizer. Finalizer callbacks may not happen immediately after garbage collection, and may not happen in the same order as actual garbage collection. They're useful to perform clean-up in response to garbage-collection, but you can't reliably use them to, say, record meaningful metrics about memory usage.
```
mathiasbynens, line 220: If we’re gonna phrase it this way, let’s point to https://web.dev/monitor-total-page-memory-usage/ for this last bit?

src/features/weak-references.md
```diff
@@ -177,7 +177,7 @@ Eventually, after a garbage collection occurs, the `MovingAvg` instance and the
...
 
 :::note
-**Note:** The finalization callback does not run immediately after garbage-collecting the event listener. It either runs at some point in the future, or not at all — the spec doesn’t guarantee that it runs! Keep this in mind when writing code.
+**Note:** The finalization callback does not run immediately after garbage-collecting the event listener, so don't use it for important logic or metrics. The timing of garbage collection and finalization callbacks is unspecifed. In fact, an engine that never garbage collected would be fully compliant. However, it's safe to assume that engines _will_ garbage collect, and finalization callbacks will be called at some later time, unless the environment is discarded (such as the tab closing, or the worker terminating). Keep this uncertainty in mind when writing code.
```
syg, line 180:
```suggestion
**Note:** The finalization callback does not run immediately after garbage-collecting the event listener, so don't use it for important logic or metrics. The timing of garbage collection and finalization callbacks is unspecified. In fact, an engine that never garbage collected would be fully compliant. However, it's safe to assume that engines _will_ garbage collect, and finalization callbacks will be called at some later time, unless the environment is discarded (such as the tab closing, or the worker terminating). Keep this uncertainty in mind when writing code.
```

syg: GH isn't highlighting the word because I think the line is too long. The typo is "unspecifed" -> "unspecified".

src/features/weak-references.md
jakearchibald:
```suggestion
For example, don’t place important logic in the code path of a finalizer. Finalizer callbacks may not happen immediately after garbage collection, and may not happen in the same order as actual garbage collection. They're useful to perform clean-up in response to garbage-collection, but you can't reliably use them to, say, record meaningful metrics about memory usage. For that use case, see [`measureMemory`](https://web.dev/monitor-total-page-memory-usage/).
```

src/features/weak-references.md
jakearchibald: I've tried to address this

src/features/weak-references.md
jakearchibald: Done

src/features/weak-references.md
```diff
@@ -177,7 +177,7 @@ Eventually, after a garbage collection occurs, the `MovingAvg` instance and the
...
 
 :::note
-**Note:** The finalization callback does not run immediately after garbage-collecting the event listener. It either runs at some point in the future, or not at all — the spec doesn’t guarantee that it runs! Keep this in mind when writing code.
+**Note:** The finalization callback does not run immediately after garbage-collecting the event listener, so don't use it for important logic or metrics. The timing of garbage collection and finalization callbacks is unspecified. In fact, an engine that never garbage collected would be fully compliant. However, it's safe to assume that engines _will_ garbage collect, and finalization callbacks will be called at some later time, unless the environment is discarded (such as the tab closing, or the worker terminating). Keep this uncertainty in mind when writing code.
```
RReverser, line 180: "an engine that never garbage collected" / "garbage collect" sounds a bit odd IMO, wouldn't "collects garbage" be technically more correct?

```suggestion
**Note:** The finalization callback does not run immediately after garbage-collecting the event listener, so don't use it for important logic or metrics. The timing of garbage collection and finalization callbacks is unspecified. In fact, an engine that never collects garbage would be fully compliant. However, it's safe to assume that engines _will_ collect garbage, and finalization callbacks will be called at some later time, unless the environment is discarded (such as the tab closing, or the worker terminating). Keep this uncertainty in mind when writing code.
```

RReverser: Or, I guess, the dashed version "garbage-collects" used in other places in the post is also fine.

jakearchibald: I'm not confident in the conventions here, so I'm happy with whatever you're happy with.

RReverser: Not sure if there's a convention, and not a native speaker, it's just that "garbage collect" sounds odd as a  verb in terms of a grammatical construction.

jakearchibald:
```suggestion
**Note:** The finalization callback does not run immediately after garbage-collecting the event listener, so don't use it for important logic or metrics. The timing of garbage collection and finalization callbacks is unspecified. In fact, an engine that never garbage-collects would be fully compliant. However, it's safe to assume that engines _will_ garbage collect, and finalization callbacks will be called at some later time, unless the environment is discarded (such as the tab closing, or the worker terminating). Keep this uncertainty in mind when writing code.
```

jakearchibald: I've gone with "garbage-collects"

src/features/weak-references.md
RReverser: Looks like this is resolved, marking as such.

RReverser:
As far as I can tell, looks good. Thanks @jakearchibald!

PR #442, opened 2020-07-20
Add AggregateError explanation

mathiasbynens:


src/features/promise-combinators.md
```diff
@@ -136,8 +136,16 @@ try {
   // → e.g. 'b'
 } catch (error) {
   // All of the promises were rejected.
-  console.log(error);
+  console.assert(error instanceof AggregateError);
+  // Log the rejection values:
+  console.log(error.errors);
```
RReverser, line 141: General nitpick: I wonder if we could use a better example here. It's *very* rare for `fetch` to reject (it has to be either an invalid URL, or a general issue like network error, or something like that), so it's perhaps not the best use-case. And then, we don't even use the return response in a meaningful way.

I think an artificial example with couple of `Promise.resolve` / `Promise.reject` could work better here, especially because we could expand the comment with expected rejection values.

src/features/promise-combinators.md
```diff
@@ -136,8 +136,16 @@ try {
...
 }
 ```
 
 This code example checks which endpoint responds the fastest, and then logs it. Only if _all_ of the requests fail do we end up in the `catch` block, where we can then handle the errors.
+
+`Promise.any` rejections can represent multiple errors at once. To support this at the language-level, a new error type called `AggregateError` is introduced. In addition to its basic usage in the above example, `AggregateError` objects can also be programmatically constructed, just like the other error types:
+
+```js
+const aggregateError = new AggregateError(errors, 'Stuff went wrong!');
```
RReverser, line 150: I'd like to make it clearer that it accepts an array. Maybe like this?

```suggestion
const aggregateError = new AggregateError([errorA, errorB, errorC], 'Stuff went wrong!');
```

src/features/promise-combinators.md
mathiasbynens: For context, I actively tried to avoid artificial examples in an attempt to make the examples in this article as real-world as possible (while still being easy to understand).

I agree it's rare for `fetch` to reject, but I think that's the common use case of `Promise.any`: you may not expect any of the promises to reject, but you can (and probably should) still handle the all-promises-reject case.

RReverser: > I actively tried to avoid artificial examples in an attempt to make the examples in this article as real-world as possible

Oh yeah, I understand and agree with the goal, it's just that this particular example still feels a bit artificial despite using a real-world API...

But my bigger issue is that, from comment, it's not very clear what the output of `console.log` is, and in most cases I prefer it to be part of the docs so that user knows what to expect. Can you think of such example output for `fetch` use-case?

mathiasbynens: Hmm, we could add something like this to the log:

```
// → [
//     <TypeError: Failed to fetch /endpoint-a>,
//     <TypeError: Failed to fetch /endpoint-b>,
//     <TypeError: Failed to fetch /endpoint-c>
//   ]
```

mathiasbynens: Done. WDYT?

RReverser: I think this works, thanks!

PR #443, opened 2020-07-21
Add note on special replacement patterns w.r.t. replaceAll

mathiasbynens:


src/features/string-replaceall.md
```diff
@@ -100,6 +100,22 @@ For consistency with the pre-existing APIs in the language, `String.prototype.re
 
 The important piece of new functionality lies in that first item. `String.prototype.replaceAll` enriches JavaScript with first-class support for global substring replacement, without the need for regular expressions or other workarounds.
 
+## A note on special replacement patterns { #special-patterns }
+
+Worth calling out: both `replace` and `replaceAll` support [special replacement patterns](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter). Although these are most useful in combination with regular expressions, some of them (`$$`, `$&`, ``$` ``, and `$'`) also take effect when performing simple string replacement, which can be surprising:
```
RReverser, line 105: Woah, TIL!

mathiasbynens: FTR, I argued against inheriting this behavior for `replaceAll`, but the committee decided that it'd be confusing for a method named `replaceAll` not to match `replace` in this way, and also that choosing a different name like `substitute` would be confusing. 🤷‍♂️

LeszekSwirski: So if you want to replace arbitrary `foo` with arbitrary `bar`, you need to do `string.replaceAll(foo, bar.replaceAll("$","$$$$"))`?

RReverser: @LeszekSwirski As far as I understand from the snippet, you can do `string.replaceAll(foo, () => bar)` instead.

RReverser: @mathiasbynens I didn't even know that `replace` had that behaviour on strings TBH.

mathiasbynens: You're not wrong, Leszek 😭

PR #453, opened 2020-07-23
Add more redirects

camillobruni:
- versioned redirects for API
- default redirect for v8.dev/tool
- default redirect for v8.dev/api

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -64,7 +64,7 @@
         "type": 301
       },
       {
-        "source": "/tools)",
+        "source": "/tools",
         "destination": "https://v8.github.io/tools/head",
         "type": 301
       },
--- a/firebase.json
+++ b/firebase.json
@@ -64,8 +64,13 @@
         "type": 301
       },
       {
-        "regex": "/tools(?P<tool>/.*)",
-        "destination": "https://v8.github.io/tools/head:tool",
+        "source": "/tools)",
+        "destination": "https://v8.github.io/tools/",
+        "type": 301
+      },
+      {
+        "source": "/tools/:tool*)",
+        "destination": "https://v8.github.io/tools/head/:tool",
         "type": 301
       }
     ],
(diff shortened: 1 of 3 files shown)


gsathya:
maybe we should add a navigation toggle called "Tools"

camillobruni:
I'll add a link in the docs section.

gsathya:
what if we make HEAD as the default and the different versions behind `v8.dev/tools/versions`? HEAD is the one I care about 99% of the time, let's make it easier to access that behind `v8.dev/tools`

camillobruni:
v8.dev/tools/system-analyzer already does that, right.
v8.dev/tools => v8.dev/tools/head seems more consistent

firebase.json
```diff
@@ -59,8 +59,18 @@
...
       },
       {
-        "regex": "/tools/(?P<version>v[0-9]+.*)",
-        "destination": "https://v8.github.io/tools/:version",
+        "regex": "/tools/(?P<version_tool>v[0-9]+.*)",
+        "destination": "https://v8.github.io/tools/:version_tool",
+        "type": 301
+      },
+      {
+        "source": "/tools)",
```
RReverser, line 67: @camillobruni Those parentheses don't look right?

camillobruni: ups.

firebase.json
```diff
@@ -59,8 +59,18 @@
...
+        "destination": "https://v8.github.io/tools/:version_tool",
+        "type": 301
+      },
+      {
+        "source": "/tools",
+        "destination": "https://v8.github.io/tools/head",
+        "type": 301
+      },
+      {
+        "source": "/tools/versions)",
```
RReverser, line 72: @camillobruni Same here.

RReverser:
@camillobruni Please verify if redirects are working correctly once CI finishes deployment.

PR #455, opened 2020-07-29
Preload dark mode toggle code

tomayac:
The code for the `<dark-mode-toggle>` is discovered too late (see last row in screenshot below), causing flash of white:

Screen Shot 2020-07-29 at 10 59 51

This PR preloads the code. Fixes https://github.com/v8/v8.dev/issues/335 (at least improves the situation).

### DIFF
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -6,6 +6,7 @@
     <meta name="viewport" content="width=device-width,initial-scale=1">
     <meta name="color-scheme" content="dark light">
     <link rel="stylesheet" href="/_css/main.css">
+    <link rel="modulepreload" href="/_js/dark-mode-toggle.mjs">
 {% if '/features' in page.url %}
     <link rel="stylesheet" href="/_css/feature-support.css">
 {% endif %}
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -6,7 +6,6 @@
     <meta name="viewport" content="width=device-width,initial-scale=1">
     <meta name="color-scheme" content="dark light">
     <link rel="stylesheet" href="/_css/main.css">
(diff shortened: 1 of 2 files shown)


RReverser:
Oh, interesting. Perhaps we should just split it out into a separate `<script type="module"` in the same place altogether?

tomayac:
> Oh, interesting. Perhaps we should just split it out into a separate `<script type="module"` in the same place altogether?

We can also do this. Though the logic would still be in `main.js` (which is loaded early). I think preloading would do the job already.

RReverser:
> I think preloading would do the job already.

Yeah, it's just that in this case we seem to unnecessarily duplicate script in two places and depend on browser support for `modulepreload` (which e.g. Firefox doesn't have).

With `script type="module"` it would both be slightly simpler and work everywhere.

tomayac:
> > I think preloading would do the job already.
> 
> Yeah, it's just that in this case we seem to unnecessarily duplicate script in two places and depend on browser support for `modulepreload` (which e.g. Firefox doesn't have).
> 
> With `script type="module"` it would both be slightly simpler and work everywhere.

Convinced. https://github.com/v8/v8.dev/pull/455/commits/91cb79c5579276a68499825d2ea2c409905b5618

RReverser:
@tomayac Just to double-check - you don't observe the issue even though the script is in the bottom? I thought you'd need to put it inside the `<head>` like you did with preload tag.

tomayac:
> @tomayac Just to double-check - you don't observe the issue even though the script is in the bottom? I thought you'd need to put it inside the `<head>` like you did with preload tag.

I decided to put it at the end to have the scripts together. Since the logic is loaded in `main.js`. It's loaded early on now for sure:

Screen Shot 2020-07-29 at 16 55 11

As discussed earlier, this makes the issue less bad. We can't fix it completely since we're fighting the browser.

PR #457, opened 2020-08-04
Update information on using perf

backes:
The standard installed kernel and perf tool are often good enough, so mark the
respective section optional. Also, use gn directly instead of the v8gen.py tool.


backes:
Can someone review please?
@gahaas @ngzhian

ngzhian:
This LGM, I don't have perms to approve this, sorry Clemens :(

PR #458, opened 2020-08-07
Add tags to the Understanding ECMAScript posts

marjakh:
Also unify the posts a bit (add "Summary" to the ones where it was missing etc).

### DIFF
--- a/src/blog/understanding-ecmascript-part-1.md
+++ b/src/blog/understanding-ecmascript-part-1.md
@@ -6,9 +6,13 @@ avatars:
 date: 2020-02-03 13:33:37
 tags:
   - ECMAScript
+  - Understanding ECMAScript
 description: 'Tutorial on reading the ECMAScript specification'
 tweet: '1224363301146189824'
 ---
+
+[All episodes](/blog/tags/understanding-ecmascript)
+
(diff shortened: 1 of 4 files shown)


src/blog/understanding-ecmascript-part-1.md
```diff
@@ -6,9 +6,13 @@ avatars:
 date: 2020-02-03 13:33:37
 tags:
   - ECMAScript
+  - Understanding ECMAScript
```
RReverser, line 9: Huh, do tags with spaces actually work and render fine?

marjakh: They seem to "work" as in they will get transformed to the space-free version. I can also use that explicitly if you'd prefer.

marjakh: Ok this comment was inaccurate. So, they render as UnderstandingECMAScript and the link is /blog/tags/understanding-ecmascript

So maybe it's fine? I'd like the link to be understanding-ecmascript and not understandingecmascript.

RReverser: Yeah the link sounds good. The rendering without spaces is a bit weird, but if you're comfortable with it, fine by me (I suppose we can fix it later if it's a problem).

PR #459, opened 2020-08-11
Minor fixes in perf docs

backes:
Fix an oversight from #457 ("out.gn" -> "out").
Also, make sure that the "gn gen" command does not overflow.


backes:
Thanks, Ingvar!

PR #462, opened 2020-08-27
Add AVIF avatars

mathiasbynens:
Across all avatars, this saves 136.6 kB:

    $ perl -le 'map { $sum += -s } @ARGV; print $sum' -- src/_img/avatars/*.jpg
    447248

    $ perl -le 'map { $sum += -s } @ARGV; print $sum' -- src/_img/avatars/*.avif
    310680

@RReverser @surma @jakearchibald Thanks for adding AVIF support to Squoosh!


surma:
Why not add WebP at the same time?

mathiasbynens:
> Why not add WebP at the same time?

It didn’t seem worth the trouble — in a future where all browsers support AVIF, there’s no point in also serving WebP. Since Firefox and Chrome will already support AVIF soon, it’d only benefit Safari, and only until they add AVIF support.

sayjeyhi:
just out of curiosity, did you generate avatars with `Squoosh` or one of [libavif](https://github.com/AOMediaCodec/libavif) or avifenc on homebrew?

mathiasbynens:
@sayjeyhi I used Squoosh for a few avatars to easily (visually) find the “optimal” settings, and then used `avifenc` (installed through [`brew install joedrago/repo/avifenc`](https://github.com/joedrago/homebrew-repo)) to automate the conversion with those settings for the remaining images. I hear the Squoosh team is considering batch support, which will make this even easier in the future!

surma:
> Firefox and Chrome will already support AVIF soon

“soon” seems optimistic esp. with regards to FF.

mathiasbynens:
> > Firefox and Chrome will already support AVIF soon
> 
> “soon” seems optimistic esp. with regards to FF.

Oh? You probably have more background than I do here. I was assuming it wouldn’t take too long anymore since [Firefox has had AVIF support behind a flag for a while now](https://bugzilla.mozilla.org/show_bug.cgi?id=1625363) but my assumption may be wrong. 

Either way I’m not opposed to also adding WebP in a separate PR 👍  I was just explaining my thinking.

mathiasbynens:
Gonna go ahead and merge this since it seems uncontroversial.

Issue #464, opened 2020-09-02
Webpage on v8.dev blog not rendering HTML


indiescripter:
At this blog URL https://v8.dev/blog/web-tooling-benchmark ,

> Announcing the Web Tooling Benchmark
Published 06 November 2017 · Tagged with benchmarks Node.js

Clicking on *Node.js* item to the right of *benchmarks* takes the browser to URL

https://v8.dev/blog/tags/node.js

which is HTML text but is being send to the browser by the origin server with `text/javascript` content-type.

Looks like the backend is being tricked into sending `text/javascript` rather than `text/html` simply because the URL ends in `.js`

Suggest workaround is to drop the `.js` suffix & change that blog tag to simply "Node" as per adoption in this book:

https://subscription.packtpub.com/book/web_development/9781849515146/1/ch01lvl1sec10/spelling-node-node-js-or-node-js

MyIsaak:
Hi @indiescripter, I reproduced your issue and provided a PR. Please check it out, thanks!

indiescripter:
Hey @MyIsaak, thank you plus looks like you went to all the trouble to sign the Google CLA.

I'm cloning your patch branch right now; will build, test & let you know. Is it possible/useful to add my handle to the Assignees field in your PR (or maybe that's only for v8.dev admin to do on their side; I don't know).

Back in a while after some morning coffees. It's 5.30 am in my neck of the woods!

indiescripter:
Hi @MyIsaak, please see my thorough test report in PR #465 comments:

https://github.com/v8/v8.dev/pull/465#issuecomment-686776204

PR #465, opened 2020-09-03
Remove unneeded routing rule (#464)

MyIsaak:
This pull request fixes the issue proved at #464 where a URL ending with `.js` (in the case for `/tags/Node.js`) wouldn't render a html but a javascript page. 

I've removed the rule in `firebase.json` that forces the header of all files under the `**/*.@(js|mjs)` pattern. Removing the rule delegates the routing to the defaults defined by `superstatic`.

I've tested this fix quickly by navigating around and checking the headers from the browser inspector which seem to be correct. Please let me know if it works.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -88,13 +88,6 @@
         }]
       },
       {
-        "source": "**/*.@(js|mjs)",
-        "headers": [{
-          "key": "Content-Type",
-          "value": "text/javascript;charset=utf-8"
-        }]
-      },
-      {
         "source": "/*.atom",
         "headers": [{
           "key": "Content-Type",


MyIsaak:
@googlebot I signed it!

indiescripter:
Hey @MyIsaak,

cc: @mathiasbynens

**TL;DR PR fixes issue #464**


Thanks for your patch Isaak. LGTM 100% & your approach to fix the issue at the HTTP network serving level was better than my suggestion to drop the `.js` suffix on the URL (unfortunately I have zero experience with Firebase configuration).

I've tested your patch thoroughly using both curl at the CLI level and with a browser.

Reporting my findings of both tests methods as follows:

**A. Via curl**

```shell
# Find all .js files in the built dist folder that are likely affected by the patch.
v8.dev/dist$ find . -name "*.js"
./blog/tags/node.js
./sw.js
./features/tags/node.js
./_js/legacy.js

# This one, the subject of the issue, was previously returning text/javascript content-type.
# Pass: Now correctly returns text/html
v8.dev/dist$ curl -I http://localhost:9001/blog/tags/node.js
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Content-Length: 2375
ETag: f961ff5b5947420c6b16f4ddf8e12fba
Last-Modified: Thu, 03 Sep 2020 20:10:42 GMT
Cache-Control: max-age=31557600
Date: Thu, 03 Sep 2020 20:42:09 GMT
Connection: keep-alive
Keep-Alive: timeout=5

# This one previously returned application/javascript content-type.
# Pass: Still returns application/javascript (unchanged by patch).
v8.dev/dist$ curl -I http://localhost:9001/sw.js
HTTP/1.1 200 OK
Content-Type: application/javascript; charset=utf-8
Content-Length: 660
ETag: bada9a7f6520d81425a45cc3bbc5754b
Last-Modified: Thu, 03 Sep 2020 20:10:34 GMT
Cache-Control: max-age=31557600
Date: Thu, 03 Sep 2020 20:42:24 GMT
Connection: keep-alive
Keep-Alive: timeout=5

# Did not check what this one previously returned, nevertheless
# Pass: Returns text/html.
v8.dev/dist$ curl -I http://localhost:9001/features/tags/node.js
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Content-Length: 2045
ETag: 0ebd2e2c6f345174273bb1bfe33d52cb
Last-Modified: Thu, 03 Sep 2020 20:10:42 GMT
Cache-Control: max-age=31557600
Date: Thu, 03 Sep 2020 20:42:44 GMT
Connection: keep-alive
Keep-Alive: timeout=5

# Did not check what this one previously returned, nevertheless
# Pass: Returns application/javascript.
v8.dev/dist$ curl -I http://localhost:9001/_js/legacy.js
HTTP/1.1 200 OK
Content-Type: application/javascript; charset=utf-8
Content-Length: 989
ETag: 2b05602a8c299c812263e1066157a39f
Last-Modified: Thu, 03 Sep 2020 20:10:26 GMT
Cache-Control: max-age=31557600
Date: Thu, 03 Sep 2020 20:42:58 GMT
Connection: keep-alive
Keep-Alive: timeout=5

v8.dev/dist$ 
```




**Via browser**

*Blog page with link to both `benchmarks` and `Node.js` tags*

> Published 06 November 2017 · Tagged with benchmarks Node.js

Screenshot at 2020-09-04 06-03-36

*Blog index for `benchmarks` tag renders correctly*

> Blog posts tagged “benchmarks”

Screenshot at 2020-09-04 06-04-04

*Blog index for `Node.js` tag currently renders incorrectly (on v8.dev origin server)*

Screenshot at 2020-09-04 06-54-59

*Blog index for `Node.js` tag now renders correctly (on localhost:9001 origin test server)*

Screenshot at 2020-09-04 06-04-28

MyIsaak:
LGTM. Thanks for the thorough testing over the patch!

mathiasbynens:
Thanks for chiming in here and for testing so thoroughly. For background, this rule was added back when Firebase didn’t append `;charset=utf-8` to JavaScript files. I’ve worked with them and this has been fixed since.

Unfortunately, I don't think the `localhost` server matches the online Firebase server exactly. For example, you noted that `https://v8.dev/sw.js` was already served as `application/javascript` instead of `text/javascript` before the patch, but that’s not the case:

```
$ curl -sI https://v8.dev/sw.js | grep content-type
content-type: text/javascript;charset=utf-8
```

Ideally, we’d serve all `*.js` and `*.mjs` files with that header instead of `application/javascript`.

Then again, the current approach isn’t great, since it also serves e.g. `https://v8.dev/404-this-doesnt-exist.mjs` as `text/*`. So I’ll merge this patch. Thanks!

mathiasbynens (approved):

mathiasbynens:
Followed up with a different approach in 47cf099339c66f6c5edee75ed8a9f770d97e9c07: it still gives us control over the `Content-Type` header, but does so only for known JS files.

MyIsaak:
Agreed creating header controlling rules that lock by folder also works. Simple and effective!

PR #472, opened 2020-09-20
Slack Tracking

ripsawridge:
This is an exploration of slack tracking: why we do it, what it is, how it expresses at runtime.


src/docs/slack-tracking.md
```diff
@@ -0,0 +1,399 @@
...
+A tree grows from the root map, with a branch for each property added from that point. The current structure looks like this:
+
+<figure>
+  <img src="/_img/docs/slack-tracking/root-map-1.svg" loading="lazy">
+  <figcaption>
+  <i>(X, Y, Z) means (instance size, number of in-object properties, number of unused properties)</i>
+  </figcaption>
+</figure>
+
+These transitions based on property names are how the "blind mole" of JavaScript builds it's hidden classes behind you. This root map is also stored in the function 'Peak', so when it's used as a constructor, that map can be used to set up the `this` object.
```
MayaLekova, line 126: nit: mode

src/docs/slack-tracking.md
```diff
@@ -0,0 +1,399 @@
...
+|three|pointer to string "Switzerland"|
+|four|undefined|
+|five|undefined|
+|six|undefined|
+
+We have those extra `undefined` values there in case you decide to add more properties.  We kind of think you might, based on your behavior so far!
+
+## Optional properties
+
+It may happen that you add properties in some cases only. Suppose if height is 4000 meters or more, you want to keep track of two additional properties, `prominance` and `isClimbed`:
```
MayaLekova, line 268: nit: prominence

src/docs/slack-tracking.md
```diff
@@ -0,0 +1,399 @@
...
+
+### But wait, it's not that easy
+
+We can compile optimized code on the main thread, in which case we can get away with prematurely ending slack tracking with some calls to change the root map. However, we do as much compilation as possible on a background thread. From this thread it would be dangerous to touch the initial map because it might be changing on the main thread. So our technique is like this:
+
+ 1. **Guess** that the instance size will be what it would be if you did stop slack tracking right now. Remember this size.
+ 2. When the compilation is almost done, we return to the main thread where we can safely force completion of slack tracking if it wasn't already done.
+ 3. Check: is the instance size what we predicted? If so, **we are good!** If not, throw away the code object and head to the bar to drown our sorrows.
```
MayaLekova, line 333: To keep it informative, you can write something like:
"~~head to the bar to drown our sorrows~~ reoptimize the function"

src/docs/slack-tracking.md
ripsawridge: Where would that be? I want the image of a mole, the essentially blind creature that digs in the dirt, finding it's way carefully forward. "Blind mode" would then suggest that there is some other mode of JavaScript, but that would probably be wasm.

src/docs/slack-tracking.md
ripsawridge: thanks -- I *always* misspelled that! 👍

src/docs/slack-tracking.md
ripsawridge: Thanks, I changed it to "try again later."

ripsawridge:
Thanks, just awaiting any comments from Igor too.

ripsawridge:
Added additional diagram per offline convo. w/Igor -- thanks!

ripsawridge:
Comments from Igor:

You can put a link to a blog post with all the details about hidden classes, descriptor arrays and transition trees: https://v8.dev/blog/fast-properties
3. You can also mention that V8 uses heuristics for predicting the amount of required in-object properties storage area by "looking" at the constructor function, or in case of JS classes - at the whole inheritance chain.
4. In the object layout table you might consider counting words from 0 ))
5. Please introduce the "Map" term the first (or second) time you use the "hidden classes" term (currently they are first met together in the heap layout picture and later in the middle of the text where you describe how GC iterates heap).
6. An introduction of the "initial map" might require a bit more details - something around the words "it's a shape/map the newly-created object (via given constructor) gets before executing the constructor code".
7. Same for back pointers. Maybe you can mention what they are and their relation to the transition trees, including what transition trees are.
8. Please make the word/value tables uniform by always using either numbers of words in the first column.
9. Re --no-lazy-feedback-allocation, I think if you say %PrepareFunctionForOptimization(foo); before the first call to foo then it will do all the necessary things to make the latter %OptimizeFunctionOnNextCall() call successful.
10. How about giving this kind of link to InitialMapInstanceSizePredictionDependency just in case the file will change in the future: https://source.chromium.org/chromium/chromium/src/+/master:v8/src/compiler/compilation-dependencies.cc?q=InitialMapInstanceSizePredictionDependency&ss=chromium%2Fchromium%2Fsrc

ripsawridge:
Updated to blog post.

mathiasbynens:
@RReverser Can haz @v8js tweet for this?

PR #475, opened 2020-09-21
Remove section on debugging SIMD in Liftoff

ngzhian:
Debugging SIMD module is not ready yet.


backes (approved):
LGTM. Should we add a note to say that debugging support will follow in v8.7?

ngzhian:
I don't think we need to, maybe we can have a short section in 8.7 release notes to mention debugging support.

backes:
Ack.

RReverser (approved):

PR #494, opened 2020-10-23
Update dead link to CPU profiler docs

MayaLekova:
The old link seems to be leading to the [main DevTools page](https://developers.google.com/web/tools/chrome-devtools) instead of specific docs about the CPU profiler. This sub-page seems more appropriate.


RReverser (approved):
Thanks!

PR #497, opened 2020-10-31
update: cq link in source-code.md

RaisinTen:
I updated the link because the old link leads to a page that says that the page has moved to another link.


juzhiyuan (approved):
Good catch!

RaisinTen:
Thank you! :)

RReverser (approved):
Thanks!

PR #502, opened 2020-11-12
Feature explainer for WebAssembly-BigInt integration

kripken:
PTAL @RReverser 

Thanks!

### DIFF
--- a/src/features/wasm-bigint.md
+++ b/src/features/wasm-bigint.md
@@ -8,10 +8,9 @@ tags:
   - WebAssembly
   - ECMAScript
 description: 'BigInts make it easy to pass 64-bit integers between JavaScript and WebAssembly. This post explains what that means and why it’s useful, which includes making things simpler for developers, letting code run more quickly, and also speeding up build times.'
-tweet: 'TODO'
+tweet: ''
 ---
-The
-[JS-BigInt-Integration](https://github.com/WebAssembly/JS-BigInt-integration) feature makes it easy to pass 64-bit integers between JavaScript and WebAssembly. This post explains what that means and why it’s useful, which includes making things simpler for developers, letting code run more quickly, and also speeding up build times.
+The [JS-BigInt-Integration](https://github.com/WebAssembly/JS-BigInt-integration) feature makes it easy to pass 64-bit integers between JavaScript and WebAssembly. This post explains what that means and why it’s useful, which includes making things simpler for developers, letting code run more quickly, and also speeding up build times.
 
 ## 64-bit integers
 
@@ -46,7 +45,7 @@ Legalization would turn that into this:
     ..))
 ```
 
-Legalization is normally done on the tools side, before it reaches the VM that runs it. For example, the [Binaryen](https://github.com/WebAssembly/binaryen) toolchain library has a pass called [LegalizeJSInterface](https://github.com/WebAssembly/binaryen/blob/fd7e53fe0ae99bd27179cb35d537e4ce5ec1fe11/src/passes/LegalizeJSInterface.cpp) that does that transformation, which is run automatically in [Emscripten](https://emscripten.org/) when it is needed.
+Legalization is done on the tools side, before it reaches the VM that runs it. For example, the [Binaryen](https://github.com/WebAssembly/binaryen) toolchain library has a pass called [LegalizeJSInterface](https://github.com/WebAssembly/binaryen/blob/fd7e53fe0ae99bd27179cb35d537e4ce5ec1fe11/src/passes/LegalizeJSInterface.cpp) that does that transformation, which is run automatically in [Emscripten](https://emscripten.org/) when it is needed.
 
 ## Downsides of legalization
 
--- a/src/features/wasm-bigint.md
+++ b/src/features/wasm-bigint.md
@@ -128,7 +128,7 @@ Perfect, exactly what we wanted!
 
 And not only is this simpler, but it’s faster. As mentioned earlier, in practice it’s rare that `i64` conversions happen on a hot path, but when it does the slowdown can be noticeable. If we turn the above example into a benchmark, running many calls of `send_i64_to_js`, then the BigInt version is 18% faster.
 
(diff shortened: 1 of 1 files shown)


src/features/wasm-bigint.md
```diff
@@ -0,0 +1,136 @@
...
+title: 'WebAssembly Integration with JavaScript BigInt'
+author: 'Alon Zakai'
+avatars:
+  - 'alon-zakai'
+date: 2020-11-12
+tags:
+  - WebAssembly
+  - ECMAScript
+description: 'BigInts make it easy to pass 64-bit integers between JavaScript and WebAssembly. This post explains what that means and why it’s useful, which includes making things simpler for developers, letting code run more quickly, and also speeding up build times.'
+tweet: 'TODO'
```
RReverser, line 11:
```suggestion
tweet: ''
```

to prevent rendering

src/features/wasm-bigint.md
```diff
@@ -0,0 +1,136 @@
...
+---
+The
+[JS-BigInt-Integration](https://github.com/WebAssembly/JS-BigInt-integration) feature makes it easy to pass 64-bit integers between JavaScript and WebAssembly. This post explains what that means and why it’s useful, which includes making things simpler for developers, letting code run more quickly, and also speeding up build times.
```
RReverser, line 14: Newlines are preserved, so this renders as a line break (can see at https://v8-dev--pr502-bigint-5waus1ta.web.app/features/wasm-bigint).

```suggestion
The [JS-BigInt-Integration](https://github.com/WebAssembly/JS-BigInt-integration) feature makes it easy to pass 64-bit integers between JavaScript and WebAssembly. This post explains what that means and why it’s useful, which includes making things simpler for developers, letting code run more quickly, and also speeding up build times.
```

src/features/wasm-bigint.md
```diff
@@ -0,0 +1,136 @@
...
+
+## 64-bit integers
+
+JavaScript Numbers are doubles, that is, 64-bit floating-point values. Such a value can contain any 32-bit integer with full precision, but not all 64-bit ones. WebAssembly, on the other hand, has full support for 64-bit integers, the `i64` type. A problem occurs when connecting the two: If a Wasm function returns an i64, for example, then the VM will throw an exception if you call it from JavaScript, something like this:
```
RReverser, line 18: I'm not sure, but doesn't the exception get thrown upon instantiation?

src/features/wasm-bigint.md
```diff
@@ -0,0 +1,136 @@
...
+
+```wasm
+(module
+  (func $send_i64 (param $x_low i32) (param $x_high i32)
+    (local $x i64) ;; the real value the rest of the code will use
+    ;; code to combine $x_low and $x_high into $x
+    ..))
+```
+
+Legalization is normally done on the tools side, before it reaches the VM that runs it. For example, the [Binaryen](https://github.com/WebAssembly/binaryen) toolchain library has a pass called [LegalizeJSInterface](https://github.com/WebAssembly/binaryen/blob/fd7e53fe0ae99bd27179cb35d537e4ce5ec1fe11/src/passes/LegalizeJSInterface.cpp) that does that transformation, which is run automatically in [Emscripten](https://emscripten.org/) when it is needed.
```
RReverser, line 49: Do we need "normally" here - is it ever done on the engines side?

src/features/wasm-bigint.md
```diff
@@ -0,0 +1,136 @@
...
+```
+node --experimental-wasm-bigint a.out.js
+JS received: 0xabcd12345678
+```
+
+Perfect, exactly what we wanted!
+
+And not only is this simpler, but it’s faster. As mentioned earlier, in practice it’s rare that `i64` conversions happen on a hot path, but when it does the slowdown can be noticeable. If we turn the above example into a benchmark, running many calls of `send_i64_to_js`, then the BigInt version is 18% faster.
+
+Another benefit is that when using BigInt integration the toolchain can avoid performing legalization on the Wasm, which makes building faster. In Emscripten, legalization happens after the LLVM linker (`wasm-ld`) runs, and so if we can avoid legalization we may be able to avoid making any changes to the Wasm file at all. You can get that speedup if you build with `-s WASM_BIGINT`, and do not provide any other flags that require changes to be made. For example, `-O0 -s WASM_BIGINT`, or `-O1 -s WASM_BIGINT`, will work (but at higher optimization levels [the Binaryen optimizer will run](https://emscripten.org/docs/optimizing/Optimizing-Code.html#link-times)).
```
RReverser, line 132: I might be wrong (not a native speaker either), but

```suggestion
Another benefit is that, when using the BigInt integration, the toolchain can avoid performing legalization on the Wasm, which makes building faster. In Emscripten, legalization happens after the LLVM linker (`wasm-ld`) runs, and so, if we can avoid legalization, we may be able to avoid making any changes to the Wasm file at all. You can get that speedup if you build with `-s WASM_BIGINT`, and do not provide any other flags that require changes to be made. For example, `-O0 -s WASM_BIGINT`, or `-O1 -s WASM_BIGINT`, will work (but at higher optimization levels [the Binaryen optimizer will run](https://emscripten.org/docs/optimizing/Optimizing-Code.html#link-times)).
```

kripken: Good question, I'm not sure... the original feels better to me, but that may be wrong... maybe someone else will have an opinion.

src/features/wasm-bigint.md
RReverser: Huh that seems surprising, but okay.

src/features/wasm-bigint.md
RReverser: General rule I follow (which AFAIK applies to English just like to other languages) is that commas need to be used around parts of the sentence that are not essential to its meaning (that is, if they're removed, the sentence remains valid). In this regard, e.g. `when using the BigInt integration` or `if we can avoid legalization` give extra context for clarity but otherwise aren't essential to the meaning of the sentence, and it would still be understandable without them.

kripken: That's good advice, I agree. It does make the sentence have a great deal of commas though, which I've been told is not great either. I did a full rewrite of the paragraph, trying to avoid both, let me know what you think.

RReverser (approved):
Looks good to me, thanks!

PR #504, opened 2020-11-19
Configure AVIF MIME type for Firebase hosting

mathiasbynens:
This makes it so that AVIF images still work when requested directly, e.g.

    https://v8.dev/_img/avatars/ingvar-stepanyan.avif

…currently has `Content-Type: text/html`. Luckily this hasn’t been a problem
for displaying these images within HTML documents.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -133,7 +133,16 @@
         ]
       },
       {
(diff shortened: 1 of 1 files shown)


RReverser (approved):
Thanks!

PR #507, opened 2020-11-26
Link to our BigInt explainer

mathiasbynens:


src/features/wasm-bigint.md
```diff
@@ -14,10 +14,10 @@ The [JS-BigInt-Integration](https://github.com/WebAssembly/JS-BigInt-integration
 
 ## 64-bit integers
 
-JavaScript Numbers are doubles, that is, 64-bit floating-point values. Such a value can contain any 32-bit integer with full precision, but not all 64-bit ones. WebAssembly, on the other hand, has full support for 64-bit integers, the `i64` type. A problem occurs when connecting the two: If a Wasm function returns an i64, for example, then the VM will throw an exception if you call it from JavaScript, something like this:
+JavaScript Numbers are doubles, that is, 64-bit floating-point values. Such a value can contain any 32-bit integer with full precision, but not all 64-bit ones. WebAssembly, on the other hand, has full support for 64-bit integers, the `i64` type. A problem occurs when connecting the two: If a Wasm function returns an i64, for example, then the VM throws an exception if you call it from JavaScript, something like this:
 
 ```
-TypeError: wasm function signature contains illegal type
+TypeError: Wasm function signature contains illegal type
```
RReverser, line 20: Isn't this a quote of how actual error looks like?

RReverser (approved):

src/features/wasm-bigint.md
mathiasbynens: Previous sentence says "like" so there's some freedom to correct the casing here, IMHO. We should correct it upstream too! @backes @gahaas Are you familiar with this error message?

RReverser: > so there's some freedom to correct the casing here

I think it was meant to be exact quote (and code formatting suggests that too).

backes: Looks like the actual error message is "type incompatibility when transforming from/to JS":
https://source.chromium.org/chromium/chromium/src/+/master:v8/src/common/message-template.h;l=564;drc=40ad911657e160af18dfc0ebe36c0ea3078fbf25

backes: And it's a `TypeError` indeed: https://source.chromium.org/chromium/chromium/src/+/master:v8/src/runtime/runtime-wasm.cc;l=149;drc=14c5b0ae67db0cb41a61929008c7af0d3d75b953

RReverser: Odd. @kripken where did you take the error message above from - I think on the original PR you said you checked it with V8?

RReverser: Ah, blame shows that it was a recent change: https://chromium-review.googlesource.com/c/v8/v8/+/2488688

kripken: Yes, seems like it changed since the version of v8 I checked with when I wrote the post...

mathiasbynens: Does this test need updating? https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/web_tests/http/tests/wasm/resources/wasm-constants.js;l=390;drc=77578ccb4082ae20a9326d9e673225f1189ebb63

backes: This is an outdated copy of our in-tree `wasm-module-builder.js`. The `wasm-constants.js` file was merged into it in the meantime. I guess that no one uses this constant in the tests, otherwise it would fail.
So I would say it does not need updating as long as the tests still pass. It's just used for constructing a wasm module, and that still works.

PR #522, opened 2021-02-03
Make heading links copyable

tomayac:
If you click a heading, a link to it will be placed on your clipboard. 

```md
<!-- article.md -->
## Some heading

Lorem ipsum…

## Some other heading

Lorem ipsum…
```

Clicking **Some heading** will now place a deeplink on your clipboard pointing to `/article#some-heading`.

### DIFF
--- a/src/_js/main.mjs
+++ b/src/_js/main.mjs
@@ -68,6 +68,38 @@ navToggle.addEventListener('click', (event) => {
   navToggle.classList.add('hide');
 });
 
+// Make headings copyable by clicking.
+const copyLinkToClipboard = () => {
+  if (!('clipboard' in navigator)) {
+    return;
+  }
+  document
+    .querySelector('main')
+    // ToDo: Use `:is(h2, h3, h4, h5, h6)[id]` once support is better.
+    .querySelectorAll('h2[id], h3[id], h4[id], h5[id], h6[id]')
+    .forEach((heading) => {
+      heading.addEventListener('click', (ev) => {
+        // Don't jump when the '#' is clicked.
+        if (ev.target.nodeName === 'A') {
+          ev.preventDefault();
+        }
+        try {
+          navigator.clipboard.writeText(
+            `${location.origin}${location.pathname}#${heading.id}`,
+          );
+          const temp = heading.innerHTML;
+          heading.innerHTML += '&nbsp;<small>(📋 Copied.)</small>';
+          setTimeout(() => {
+            heading.innerHTML = temp;
+          }, 2000);
+        } catch (err) {
+          console.warn(err.name, err.message);
+        }
+      });
+    });
+};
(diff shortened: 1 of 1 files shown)


src/_js/main.mjs
```diff
@@ -68,6 +68,38 @@ navToggle.addEventListener('click', (event) => {
...
+        }
+        try {
+          navigator.clipboard.writeText(
+            `${location.origin}${location.pathname}#${heading.id}`,
+          );
+          const temp = heading.innerHTML;
+          heading.innerHTML += '&nbsp;<small>(📋 Copied.)</small>';
+          setTimeout(() => {
+            heading.innerHTML = temp;
+          }, 2000);
```
tomayac, line 94:
```suggestion
          }, 2_000);
```

mathiasbynens:
This is a nice gimmick, but I'm not sure why we need it. Could you clarify? Note that we already have a way to let users copy links to individual sections:

tomayac:
Try it on mobile :-) It's a real pain copying a deeplink to a heading on a mobile device. I have deployed this on my blog ([example link](https://blog.tomayac.com/2021/01/24/ruby-html-footnotes/)).

RReverser:
@tomayac I tried on mobile, seems to work quite similarly to desktop.

Click on heading - `#` appears, long tap - copy link.

I think there is an argument for making `#` more discoverable on mobile, since it's not very obvious one needs to tap on heading.

I'd be happy if we improved discoverability by e.g. making visible but semi-transparent by default, but copy-on-tap doesn't seem like solving the underlying issue in that it's still not discoverable unless you are already aware of the feature.

tomayac:
Long-pressing on mobile and then choosing between… 
- Open in new tab group
- Open in incognito tab
- **Copy link address** (This one)
- Copy link text (Not this one)
- Download link
- Share link
 or right-clicking 
…is quite some mental overload. Right-clicking on desktop (or going to the URL bar after actually navigating) is no better.

My suggestion would be to make the `#` always visible, paired with the copy code from this PR.

mathiasbynens:
> Long-pressing on mobile and then choosing between…
> 
> * Open in new tab group
> * Open in incognito tab
> * **Copy link address** (This one)
> * Copy link text (Not this one)
> * Download link
> * Share link
> 
> …is quite some mental overload. Right-clicking on desktop (or going to the URL bar after actually navigating) is no better.

Copying links to headers is not a primary use case for the website though. As you explained, it’s already possible without this patch, and IMHO it’s not quite common or painful enough to warrant adding more JavaScript. Let’s keep things simple.

LeszekSwirski:
Strong -1 to messing with my clipboard when I'm just clicking.

tomayac:
Personally I think it'd be a nice gesture to the user, but I hear you. Closing the PR.

PR #524, opened 2021-02-09
Note that match indices depend on the `d` flag

mathiasbynens:
Also mark the feature as shipping in Chrome 90.

### DIFF
--- a/src/features/regexp-match-indices.md
+++ b/src/features/regexp-match-indices.md
@@ -102,6 +102,7 @@ As mentioned above, [the new JavaScript feature](https://github.com/tc39/proposa
 
 ```js
 function getVariablePosition(source) {
+  // Notice the `d` flag, which enables `match.indices`
   const re = /(let|const|var)\s+([a-zA-Z_$][0-9a-zA-Z_$]*)/d;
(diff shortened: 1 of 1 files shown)


src/features/regexp-match-indices.md
```diff
@@ -102,7 +102,7 @@ As mentioned above, [the new JavaScript feature](https://github.com/tc39/proposa
 
 ```js
 function getVariablePosition(source) {
-  const re = /(let|const|var)\s+([a-zA-Z_$][0-9a-zA-Z_$]*)/;
+  const re = /(let|const|var)\s+([a-zA-Z_$][0-9a-zA-Z_$]*)/d;
```
jridgewell, line 105:
```suggestion
  // Notice the `d` flag, which enables `match.indices`
  const re = /(let|const|var)\s+([a-zA-Z_$][0-9a-zA-Z_$]*)/d;
```

jridgewell (commented):
Nit: I would highlight the `d` flag to draw attention to it.

syg (approved):
lgtm, thanks for doing the update, @mathiasbynens!

PR #530, opened 2021-02-23
Change -std=c++0x to -std=c++14 for hello_world.cc

bobbydigitales:
hello_world.cc fails to compile with -std=c++0x as various standard library features that it depends on are not available. Changing to -std=c++14 fixes this.


bobbydigitales:
@googlebot I signed it!

RReverser (approved):

PR #534, opened 2021-03-18
Add class static blocks explainer

syg:


src/features/class-static-initializer-blocks.md
```diff
@@ -0,0 +1,109 @@
...
+      seed = MyPRNG.entropy_pool.pop();
+    }
+    this.seed = seed;
+  }
+
+  getRandom() { ... }
+
+  static entropy_pool;
+  static {
+    this.entropy_pool = [];
```
jridgewell, line 30: Nit: If we're going to demonstrate a static field, we should just initialize with it.

```suggestion
  static entropy_pool = [];
  static {
    this.entropy_pool;
```

jridgewell (approved):

src/features/class-static-initializer-blocks.md
syg: Sure.

src/features/class-static-initializer-blocks.md
```diff
@@ -0,0 +1,108 @@
...
+date: 2021-03-19
+tags:
+  - ECMAScript
+description: 'JavaScript classes get dedicated syntax for static initialization.'
+---
+The new class static initializer block syntax lets developers gather code that should run once for a given class definition and put them in a single place. Consider the following example where a pseudo random number generator uses a static block to initialize an entropy pool once, when the `class MyPRNG` definition is evaluated.
+
+```js
+class MyPRNG {
+  seed;
```
RReverser, line 15: This seems unnecessary?

syg: The declaration? Eh, I think some folks like declaring the fields they're going to use up front.

jridgewell: I think this  is common, especially with TS requiring you to forward declare.

src/features/class-static-initializer-blocks.md
```diff
@@ -0,0 +1,108 @@
...
+  constructor(seed) {
+    if (seed == undefined) {
+      if (MyPRNG.entropy_pool.length === 0) {
+        throw new Error("Entropy pool exhausted");
+      }
+      seed = MyPRNG.entropy_pool.pop();
+    }
+    this.seed = seed;
+  }
```
RReverser, line 24: I'm tempted to say this should be static function `getSeed`, and then the class body gets reduced to something like:

```suggestion
  constructor(seed = MyPRNG.getSeed()) {
    this.seed = seed;
  }
```

or

```suggestion
  constructor(seed) {
    this.seed = seed ?? MyPRNG.getSeed();
  }
```

The last one seems clearer than the current approach, but maybe you think it would be too distracting?

syg: Sure, why not. I didn't write the example with a lot of thought. :)

syg: Actually I take it back. Factoring that out into a static method opens up public/private issues that I don't want to get into. There's no reason for an external user to get a seed, so `#getSeed` should be private. I don't want to be gratuitous with the number of new features being used here, and it's such a small method anyway.

src/features/class-static-initializer-blocks.md
```diff
@@ -0,0 +1,108 @@
+---
+title: 'Class static initializer blocks'
+author: 'Shu-yu Guo ([@_shu](https://twitter.com/_shu))'
+avatars:
+  - 'shu-yu-guo'
+date: 2021-03-19
```
RReverser, line 6:
```suggestion
date: 2021-03-30
```

src/features/class-static-initializer-blocks.md
```diff
@@ -0,0 +1,108 @@
...
+  - ECMAScript
+description: 'JavaScript classes get dedicated syntax for static initialization.'
+---
+The new class static initializer block syntax lets developers gather code that should run once for a given class definition and put them in a single place. Consider the following example where a pseudo random number generator uses a static block to initialize an entropy pool once, when the `class MyPRNG` definition is evaluated.
+
+```js
+class MyPRNG {
+  seed;
+  constructor(seed) {
+    if (seed == undefined) {
```
RReverser, line 17:
```suggestion
    if (seed === undefined) {
```

src/features/class-static-initializer-blocks.md
RReverser: @syg I think I'd still remove this - it's rather unusual to see in pure JS for regular properties, and you did say below you want to minimize number of new features to avoid distractions.

RReverser (approved):

mathiasbynens:
Btw I loved [the tweet announcing this](https://twitter.com/v8js/status/1376925666780798989) as well! Added it in f391a1d464d327845966ad2efd3a2b215e7b707e

PR #536, opened 2021-03-20
Add a small NodeJS Beginner Guide into the docs.

mtgsquad:
This is a small guide i've written for beginners to nodejs, i added it to here since its kind of good for it to be here, since many new nodejs developers come here for help, and nodejs uses chromes v8 engine, so i just wanted to write a first-timer guide. hope you like it :)


mtgsquad:
@googlebot I signed it!

Nytelife26:
This documentation is for V8, not Node. Node's documentation should be hosted with Node, V8's documentation should be hosted with V8. I think this pull request is somewhat out of scope, but I'll leave that to the maintainers I suppose.

That and I disagree with making the beginner guide opinionated on libraries and the like. It would be better to include reference links and a small walkthrough of basic concepts.

But, once again, I truly think that would be better placed in a Node related repository.

RReverser:
I agree with @Nytelife26. @mtgsquad thanks for your contribution, but Node.js docs do look more appropriate for the Node.js website, not V8 itself.

PR #537, opened 2021-03-27
chore: make inclusive language truly inclusive

Nytelife26:
While "blocklist" / "denylist" and "allowlist" / "grantlist" or other alternatives are quite commonly proposed, they are neglecting something - English is not the only language spoken on our Earth. In many languages, these fabricated alternatives have no direct translations.

For instance, Spanish. None of those alternatives have translations that make sense. Not to mention that "whitelist" and "blacklist" are already inaccessible in those languages due to their rarity. The most commonly used phrasing is, for instance, "list of blocked users" or "blocked user list".

We need to be mindful of this when creating guidelines for inclusivity. We need to be TRULY inclusive, and considerate of those that come from foreign backgrounds, otherwise, you are actively making the terminology less accessible to more people than you are helping.

This is not to create a flame or attack anyone on the V8 development team. I just truly believe that this hasn't been considered enough, and I do not wish to see the second wave of realization when people wake up to the fact that many people from other linguistic backgrounds will struggle to understand these alternatives. Software engineering should be accessible to all, no matter the abstract features of their background.

### DIFF
--- a/src/docs/respectful-code.md
+++ b/src/docs/respectful-code.md
@@ -33,16 +33,16 @@ Apply the principles above. If you have any questions, you can reach out to `v8-
 This list is NOT meant to be comprehensive. It contains a few examples that people have run into frequently.
 
 :::table-wrapper
-| Term      | Suggested alternatives                                        |
-| --------- | ------------------------------------------------------------- |
-| master    | primary, controller, leader, host                             |
-| slave     | replica, subordinate, secondary, follower, device, peripheral |
-| whitelist | allowlist, exception list, inclusion list                     |
-| blacklist | denylist, blocklist, exclusion list                           |
-| insane    | unexpected, catastrophic, incoherent                          |
-| sane      | expected, appropriate, sensible, valid                        |
-| crazy     | unexpected, catastrophic, incoherent                          |
-| redline   | priority line, limit, soft limit                              |
+| Term      | Suggested alternatives                                             |
+| --------- | ------------------------------------------------------------------ |
+| master    | primary, controller, leader, host                                  |
+| slave     | replica, subordinate, secondary, follower, device, peripheral      |
+| whitelist | exception list, inclusion list, list of allowed X / allowed X list |
+| blacklist | exclusion list, list of blocked X / blocked X list                 |
+| insane    | unexpected, catastrophic, incoherent                               |
+| sane      | expected, appropriate, sensible, valid                             |
+| crazy     | unexpected, catastrophic, incoherent                               |
+| redline   | priority line, limit, soft limit                                   |
 :::
 
 ## What if I am interfacing with something that violates this policy? { #violations }


Nytelife26:
@googlebot I signed it!

Nytelife26:
cc: @mathiasbynens (not sure who the relevant people would be but you seem to be an active maintainer so apologies)

mathiasbynens:
> cc: @mathiasbynens (not sure who the relevant people would be but you seem to be an active maintainer so apologies)

@RReverser and @surma are maintaining v8.dev. Could you PTAL?

surma:
I like your additions, @Nytelife26. I would ask that you not remove “allowlist” et al from the list, tho. For English resources, they are completely valid and have somewhat become an industry-standard as replacements for whitelist/blacklist. Happy to merge afterwards :)

Nytelife26:
> I would ask that you not remove “allowlist” et al from the list, tho. For English resources, they are completely valid and have somewhat become an industry-standard as replacements for whitelist/blacklist

For English resources *for native speakers that understand the language fluently*, yes. The entire point here is accessibility. They should not become the industry standard because they are unsuitable replacements.

See [a statement I authored](https://github.com/kludge-cs/transparency/blob/master/open-statements/2021-04-04--TERMINOLOGY.md) for more information on that, and come back with your verdict. Do not let popularity be a reason, though - the amount of people doing something does not change its value.

surma:
The industry is increasingly adopting “allowlist” and “blocklist”, despite the shortcomings you correctly point out. I worry that if we remove any appearance from “allowlist” and “blocklist” from the table, that people who are used to these terms are now confused because they don’t appear at all. How do you envision that being handled? To that extent, we need to at least consider the relative popularity of these terms.

I see your point and I like your suggestions, but I don’t decide what is and is not popular/commonly used in the industry, which is why I am trying to get your point reflected without increasing confusion.

Nytelife26:
> which is why I am trying to get your point reflected without increasing confusion.

It may be wise to link to an external resource on why we have not included them on viable suggestions, perhaps such as the links I have provided before and maybe a brief explanation then.

Nytelife26:
Following [discussion with others from the inclusive naming initiative](https://github.com/inclusivenaming/website/pull/45#discussion_r614125724), it seems that `exclusion list` and `inclusion list` are likely the best solutions.

It:
1. Makes actual linguistic sense in English
2. Does not rely on native-speaker inference and can be translated properly
3. Lines up with software use cases nicely

On point 3:
```yml
include: [a, b, c]
exclude: [a, b, c]
```
That's quite a common configuration pattern, and it lines up nicely with inclusion / exclusion list.

Thoughts on this, anyone?

RReverser:
I don't really see the value in excluding something from the _list of suggested alternatives_, especially if such alternatives are indeed already used in the wild. If any of the proposed alternatives don't work for a particular project, they can always choose another one.

Nytelife26:
> I don't really see the value in excluding something from the _list of suggested alternatives_, especially if such alternatives are indeed already used in the wild. If any of the proposed alternatives don't work for a particular project, they can always choose another one.

Because we shouldn't be suggesting those "alternatives", if they can even be called that since they are not viable. They are more harmful than beneficial, as we have stated, for many reasons - and as such, there is no reason to use them, especially when there are better alternatives.

RReverser:
> for many reasons

As far as I understand, the main reason is usage in other languages. As non-native English speaker, I appreciate that concern but terminology (or words in general) by design don't map 1:1 between different languages, and I assume that other languages would have their own tables.

This table is in English and is describing alternatives used in English language, and since those words are already used in the wild, removing them is worse and would cause more confusion than keeping.

Nytelife26:
> As far as I understand, the main reason is usage in other languages. As non-native English speaker, I appreciate that concern but terminology (or words in general) by design don't map 1:1 between different languages, and I assume that other languages would have their own tables.

Terminology doesn't map 1:1, sure. But there are no translations for such words. Not to mention they don't even make sense in their language of origin - blocklist, denylist, et cetera, are not valid in English.

They are less clear, they are invalid, they do not translate well, and others. I would implore you to read the statement I linked to in this pull request if you need any further reasoning.

> This table is in English and is describing alternatives used in English language, and since those words are already used in the wild, removing them is worse and would cause more confusion than keeping.

That's why I proposed adding a note below the table as to why certain examples were omitted, so we can explain the decision or link to other resources. Objectively, the present alternatives are not sufficient, and if we have the opportunity to use better, as engineers, we should steer for the better choice.

surma:
I understand your point and I want to add your _additions_. However, we can’t just remove terms, as we are also trying to align with the [Inclusive Chromium Code guidelines](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/inclusive_code.md), for example.

Please add the removed terms back to the table so we can merge this.

Nytelife26:
That defeats the entire purpose of the pull request. The point is these terms do more harm than good. If I must make a pull request there too prior to this so things are consistent, I will, but I'm not doing half a job here and leaving this to continue happening.

If I have to explain why these terms are not present in the document, that's fine - but I will continue to drive this point until I am either proven wrong or people begin to realize the situation.

Thank you for understanding, and I appreciate that you want to work with me here rather than against me, but I simply cannot drop the movement behind this just because "standards". I have discussed with members of other projects, including employees of IBM, and further discussion is in progress. If you wish to participate, you may do so [here](https://github.com/inclusivenaming/website/pull/45#discussion_r614125724). If not, I'll keep going until this is all resolved :)

Once again, thank you. I appreciate your time and consideration.

Nytelife26:
Status, anyone?

Nytelife26:
> I can not really change anything from my last statement. I don’t
> control the Chromium Code guidelines and don’t know who has the power
> to change them. I am assuming there is a whole lot of process involved.

There most likely is. And that's what I'll do. It seems to me that they
don't accept pull requests from GitHub, much like v8 itself, so I'll
have to do some more investigation on their Google-hosted source and
make the relevant communications in any mailing lists.

Nytelife26:
Found it. I am unable to post on the issues tracker for the Chromium project, [including for the respective issues](https://bugs.chromium.org/p/chromium/issues/detail?id=842296), so if someone could forward the following message via issue or contact on my behalf (the file is located at `HEAD/styleguide/inclusive_code.md` on the googlesource page so associated committers will be useful mentions). 

```
I am submitting this as I cannot comment on the current issue, #842296. As a result of a discussion in v8 (https://github.com/v8/v8.dev/pull/537) and the Inclusive Naming Initiative with employees of Google and IBM respectively (https://github.com/inclusivenaming/website/pull/45), which can be seen for context, "blocklist" and "allowlist" are not suitable replacements in any circumstances and should be removed from the inclusive code guidelines (I am raising this issue here to bring broader attention to it and to enable this change to take place elswhere within Chromium's organizational structure, v8 included). Instead, things like "inclusion list" or "exclusion list" (just examples) would be better. The primary issue with current proposed alternatives is translatability and the fact that they do not work within the English language and so have no direct translations and are very difficult to pick up from a perspective with lack of knowledge of English given that they can only be derived and interpreted correctly with nuance, and also fail when ran through translators, including Google Translate.

Thank you for your time and attention. This is not an attack on the people who proposed those originally, but there are better ways to go about this, given that those alternatives pose more harm than good overall. As engineers, we should strive to do our best, and I do not wish to witness a second wave of realization when we have to correct all of our corrections if they are to go further.

For more information, see the linked PRs GitHub-side.
```

Thank you in advance.

surma:
I’d recommend opening a new issue instead and reference that one

Nytelife26:
> I’d recommend opening a new issue instead and reference that one

I tried, wouldn't let me as aforementioned unfortunately, which is why
I'm asking if a Googler can assist me with making this happen :)

Nytelife26:
Turns out it allowed me to today. If anyone wants to participate or follow the discussion, [here](https://bugs.chromium.org/p/chromium/issues/detail?id=1208335)

Nytelife26:
The lack of engagement with the Chromium issue is, to say the least, disappointing. I'd like to hope that an organization like Google wouldn't just `sed -i` with linguistics to avoid accusations of racism without caring about any of the actual implications, but it's starting to seem that way.

It has become evident that these changes were made for virtue signalling and nothing more. I'll be working closely with the INI, now, until this gains some actual internal attention from a Chromium team member - because plenty of them are interested in avoiding exclusion, but none of them are interested in engineering a real solution.

PR #538, opened 2021-03-30
Add compile-arm64.md

planglois:
This documents a few tips and tricks that can help building V8 natively
on non-x86 machines.


planglois:
Hi there!

Lately we've been doing more and more of our V8 work natively on Arm64 workstations and so thought it might be a good idea to document a few tips and tricks, even though it's not officially supported.

WDYT?

RReverser:
How does this intersect with https://github.com/v8/v8.dev/blob/main/src/docs/cross-compile-arm.md? Would be good to avoid duplication if possible, but I don't have enough understanding of differences to review this. /cc @backes

planglois:
Hi @RReverser! Sorry for the late reply!

Yeah the content doesn't overlap with cross-compile-arm.md really so I thought it'd be better to do a separate page. The other reason to put it in another page was that compiling V8 natively on an Arm64 machine isn't officially supported, as opposed to cross-compiling. Although, even if it's not officially supported, these few tips and tricks have been useful to us at Arm, so we figured others may find them useful too!

WDYT?

backes:
The content looks useful, thanks. We should just be careful to emphasize that we still don't officially support building on arm64, even though we have some tips and tricks now.

src/docs/compile-arm64.md
```diff
@@ -0,0 +1,102 @@
+---
+title: 'Compiling on Arm64'
```
RReverser, line 2: Nit: should capitalization here and elsewhere be "ARM64"?

planglois: These days we no longer capitalize and use "Arm" everywhere, including when referring to the architecture. That being said, it's true that "Arm64" isn't the real name, it should technically be AArch64, but historically we've used Arm64 throughout V8 so I thought we'd stick with that in the documentation as well.

planglois:
I'm sorry it took me over a month to get back to this!

I've:
  - Put emphasis on "__not officially supported__" in the introduction.
  - Changed the wording in a couple of places to say "may run into issues" rather than "will run into issues", given things may just work sometimes :-)
  - Removed the mention of building clang on Apple Silicon, since nowadays `gclient sync` downloads a working clang by default.

I realized I did a force push instead of adding an extra commit, appologies, I'm not yet super familiar with the Github review process. Here's what actually changed:

```diff
diff --git a/src/docs/compile-arm64.md b/src/docs/compile-arm64.md
index 805000b..a8720dc 100644
--- a/src/docs/compile-arm64.md
+++ b/src/docs/compile-arm64.md
@@ -2,7 +2,7 @@
 title: 'Compiling on Arm64'
 description: 'Tips and tricks to build V8 natively on Arm64'
 ---
-If you've gone through instructions on how to [check out](/docs/source-code) and [build](/docs/build-gn) V8 on a machine that is not x86, you'll have ran into a bit of trouble, due to the build system downloading native binaries and then not being able to run them. However, even though using an Arm64 machine to work on V8 is not officially supported, overcoming those hurdles is pretty straightforward.
+If you've gone through instructions on how to [check out](/docs/source-code) and [build](/docs/build-gn) V8 on a machine that is not x86, you may have ran into a bit of trouble, due to the build system downloading native binaries and then not being able to run them. However, even though using an Arm64 machine to work on V8 is __not officially supported__, overcoming those hurdles is pretty straightforward.
 
 ## Bypassing `vpython`
 
@@ -39,7 +39,7 @@ export PATH=/path/to/gn/out:$PATH
 
 ## Compiling clang
 
-By default, V8 will want to use its own build of clang, shipped for x86. You could tweak GN arguments in order to [use the system's clang or GCC](#system_clang_gcc), however, you may want to use the same clang as upstream, as it will be the most supported version.
+By default, V8 will want to use its own build of clang that may not run on your machine. You could tweak GN arguments in order to [use the system's clang or GCC](#system_clang_gcc), however, you may want to use the same clang as upstream, as it will be the most supported version.
 
 You can build it locally directly from the V8 checkout:
 
@@ -49,8 +49,6 @@ You can build it locally directly from the V8 checkout:
                                --disable-asserts
\ ```
 
-If you're building on an Apple Silicon Mac, the `--gcc-toolchain` option isn't necessary.
-
 ## Setting up GN arguments manually
 
 Convenience scripts may not work by default, instead you'll have to set GN arguments manually following the [manual](/docs/build-gn#gn) workflow. You can get the usual "release", "optdebug" and "debug" configurations with the following arguments:

```

PR #542, opened 2021-04-12
Add private brand check explainer

marjakh:


src/features/private-brand-checks.md
```diff
@@ -0,0 +1,120 @@
...
+A.test({}); // false
+
+class B {
+  #foo = 0;
+}
+
+A.test(new B()); // false; it's not the same #foo
+```
+
+Since private names are only available inside the class which defines them, the test must also occur inside the class, for example in a method like`static test` above.
```
jridgewell, line 44: Missed whitespace.
```suggestion
Since private names are only available inside the class which defines them, the test must also occur inside the class, for example in a method like `static test` above.
```

src/features/private-brand-checks.md
```diff
@@ -0,0 +1,120 @@
...
+
+Subclass instances inherit private fields from the parent class:
```
jridgewell, line 46: Nit: Clarify own-ness:
```suggestion
Subclass instances receive private fields from the parent class as own-properties:
```

src/features/private-brand-checks.md
```diff
@@ -0,0 +1,120 @@
...
+
+```javascript
+class SubA extends A {};
+A.test(new SubA()); // true
+```
+
+But creating an object with a prototype with `Object.create` or setting the prototype later (via the `__proto__` setter or `Object.setPrototypeOf`) doesn't make private fields accessible:
```
jridgewell, line 53: Nit: Clarify inheritance:
```suggestion
But objects created with with `Object.create` (or that have the prototype set later via the `__proto__` setter or `Object.setPrototypeOf`) doesn't receive the private fields as own-properties. Because private field lookup only works on own-properties, the `in` operator does not find these inherited fields:
```

src/features/private-brand-checks.md
```diff
@@ -0,0 +1,120 @@
...
+
+```javascript
+const a = new A();
+A.test(Object.create(a)); // false
+let o = {};
+Object.setProrotypeOf(o, a);
+A.test(o); // false
```
jridgewell, line 60: Maybe add checks on the `__proto__` object, to clarify the field still exists?

```suggestion
const a = new A();
const o = Object.create(a);
A.test(o); // false, private field is inherited and not owned
A.test(o.__proto__); // true

const o2 = {};
Object.setProrotypeOf(o2, a);
A.test(o2); // false
A.test(o.__proto__); // true
```

jridgewell (approved):

RReverser (approved):

PR #549, opened 2021-05-05
Update iOS cross-compile instructions

schuay:
Pointer compression is unsupported and must be disabled.

### DIFF
--- a/src/docs/cross-compile-ios.md
+++ b/src/docs/cross-compile-ios.md
@@ -42,6 +42,7 @@ use_xcode_clang = true
 v8_enable_i18n_support = false        # Produces a smaller binary.
 v8_monolithic = true                  # Enable the v8_monolith target.
 v8_use_external_startup_data = false  # The snaphot is included in the binary.
+v8_enable_pointer_compression = false # Unsupported on iOS.
 ```
 
 Now build:


backes:
Is this just because of the compiler error with V8_ASSUME_ALIGNED, or do we expect more problems?
If it's only the compile error, we could just avoid using __builtin_assume_aligned on iOS.

schuay:
> Is this just because of the compiler error with V8_ASSUME_ALIGNED, or do we expect more problems?
> If it's only the compile error, we could just avoid using __builtin_assume_aligned on iOS.

The V8_ASSUME_ALIGNED error was the motivation yes, and I vaguely recall ptr-compression problems on iOS builds in the past. Do you mean that the root cause is not ptrcomp but the V8_ASSUME_ALIGNED macro?

backes:
> > Is this just because of the compiler error with V8_ASSUME_ALIGNED, or do we expect more problems?
> > If it's only the compile error, we could just avoid using __builtin_assume_aligned on iOS.
> 
> The V8_ASSUME_ALIGNED error was the motivation yes, and I vaguely recall ptr-compression problems on iOS builds in the past. Do you mean that the root cause is not ptrcomp but the V8_ASSUME_ALIGNED macro?

I don't know about further issues - but as discussed offline it seems like there were also other issues. In that case, let's just recommend to disable it on iOS.

backes (approved):

schuay:
> > > Is this just because of the compiler error with V8_ASSUME_ALIGNED, or do we expect more problems?
> > > If it's only the compile error, we could just avoid using __builtin_assume_aligned on iOS.
> > 
> > 
> > The V8_ASSUME_ALIGNED error was the motivation yes, and I vaguely recall ptr-compression problems on iOS builds in the past. Do you mean that the root cause is not ptrcomp but the V8_ASSUME_ALIGNED macro?
> 
> I don't know about further issues - but as discussed offline it seems like there were also other issues. In that case, let's just recommend to disable it on iOS.

Thanks! For posterity, the issue back then was that iOS doesn't support 4GB reservations needed for ptrcomp.

PR #551, opened 2021-05-06
Add blog post on short builtin calls

verwaest:


src/_img/short-builtin-calls/v8-browsing.svg
```diff
@@ -0,0 +1 @@
....2469Q536.2156 347.84064 535.4187 348.71564Q534.6218 349.59064 533.5281 349.59064Q532.4187 349.59064 531.7937 348.66876L531.7937 349.45ZM531.7781 346.29376Q531.7781 347.40314 532.0906 347.90314Q532.5906 348.71564 533.4343 348.71564Q534.1218 348.71564 534.6218 348.1219Q535.13745 347.52814 535.13745 346.325Q535.13745 345.10626 534.6531 344.52814Q534.1687 343.95 533.4812 343.95Q532.7937 343.95 532.2781 344.5594Q531.7781 345.15314 531.7781 346.29376ZM542.07495 347.45L543.1687 347.575Q542.9187 348.52814 542.2156 349.0594Q541.51245 349.59064 540.4343 349.59064Q539.07495 349.59064 538.26245 348.7469Q537.4656 347.90314 537.4656 346.3875Q537.4656 344.825 538.2781 343.96564Q539.0906 343.09064 540.3718 343.09064Q541.6218 343.09064 542.4031 343.9344Q543.19995 344.77814 543.19995 346.325Q543.19995 346.41876 543.19995 346.60626L538.5593 346.60626Q538.6218 347.6375 539.13745 348.1844Q539.6531 348.71564 540.4343 348.71564Q541.01245 348.71564 541.4187 348.41876Q541.8406 348.10626 542.07495 347.45ZM538.6218 345.7469L542.0906 345.7469Q542.0281 344.95 541.69995 344.5594Q541.1843 343.95 540.38745 343.95Q539.6531 343.95 539.1531 344.4344Q538.6687 344.91876 538.6218 345.7469Z" fill-rule="nonzero"/></svg>
```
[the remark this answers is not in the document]
verwaest: done, thanks

src/blog/short-builtin-calls.md
```diff
@@ -0,0 +1,45 @@
+---
+title: 'Short Builtin Calls'
+author: '[Toon Verwaest](https://twitter.com/tverwaes), The Big Short'
+avatars:
+  - toon-verwaest
+date: 2021-05-06
+tags:
+  - JavaScript
+description: 'Faster builtin calls in V8 v9.1'
```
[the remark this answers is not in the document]
verwaest: done!

src/blog/short-builtin-calls.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Direct vs Indirect Calls
+
+On 64 bit architectures, the Chrome binary, which includes these builtins, lies arbitrarily far away from JIT code. With the [x86-64](https://en.wikipedia.org/wiki/X86-64) instruction set, this means we can’t use direct calls: they take a 32-bit signed immediate that’s used as an offset to the address of the call, and the target may be more than 2 GiB away. Instead, we need to rely on indirect calls through a register or memory operand. Such calls rely more heavily on prediction since it’s not immediately apparent from the call instruction itself what the target of the call is. On [ARM64](https://en.wikipedia.org/wiki/AArch64) we can’t use direct calls at all since the range is limited to 128 MiB. This means that in both cases we rely on the accuracy of the CPU's indirect branch predictor.
```
mathiasbynens, line 23:
```suggestion
On 64-bit architectures, the Chrome binary, which includes these builtins, lies arbitrarily far away from JIT code. With the [x86-64](https://en.wikipedia.org/wiki/X86-64) instruction set, this means we can’t use direct calls: they take a 32-bit signed immediate that’s used as an offset to the address of the call, and the target may be more than 2 GiB away. Instead, we need to rely on indirect calls through a register or memory operand. Such calls rely more heavily on prediction since it’s not immediately apparent from the call instruction itself what the target of the call is. On [ARM64](https://en.wikipedia.org/wiki/AArch64) we can’t use direct calls at all since the range is limited to 128 MiB. This means that in both cases we rely on the accuracy of the CPU's indirect branch predictor.
```

src/blog/short-builtin-calls.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Indirect Branch Prediction Limitations
```
mathiasbynens, line 25:
```suggestion
## Indirect branch prediction limitations
```

src/blog/short-builtin-calls.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Temporary Solution: Copy the Builtins
+
+To avoid the cost of frequent mispredictions, and to avoid unnecessarily relying on branch prediction where possible on x86-64, we’ve decided to temporarily copy the builtins into V8's pointer compression cage on desktop machines with enough memory. This puts the copied builtin code close to dynamically generated code. The performance results heavily depend on the device configuration, but here are some results from our performance bots:
+
+![Browsing benchmarks recorded from live pages](/_img/short-builtin-calls/v8-browsing.svg)
+
+![Benchmark score improvement](/_img/short-builtin-calls/benchmarks.svg)
+
+Unembedding builtins does increase memory usage on affected devices by 1.2 to 1.4MiB per V8 instance. As a better long-term solution we’re looking into allocating JIT code closer to the Chrome binary. That way we can re-embed the builtins to regain the memory benefits, while additionally improving the performance of calls from V8-generated code to C++ code.
```
mathiasbynens, line 45:
```suggestion
Unembedding builtins does increase memory usage on affected devices by 1.2 to 1.4 MiB per V8 instance. As a better long-term solution we’re looking into allocating JIT code closer to the Chrome binary. That way we can re-embed the builtins to regain the memory benefits, while additionally improving the performance of calls from V8-generated code to C++ code.
```

RReverser (approved):

PR #552, opened 2021-05-26
Update SIMD again

RReverser:
Mostly removing instructions for Canary & OT, as well as for Node.js (which, unfortunately, only has version of V8 before the latest SIMD opcode update, and won't work with up-to-date toolchains).

### DIFF
--- a/src/features/simd.md
+++ b/src/features/simd.md
@@ -2,7 +2,7 @@
 title: 'Fast, parallel applications with WebAssembly SIMD'
 author: 'Deepti Gandluri ([@dptig](https://twitter.com/dptig)), Thomas Lively ([@tlively52](https://twitter.com/tlively52)), Ingvar Stepanyan ([@RReverser](https://twitter.com/RReverser))'
 date: 2020-01-30
-updated: 2021-04-19
+updated: 2021-05-26
 tags:
   - WebAssembly
 description: 'Bringing vector operations to WebAssembly'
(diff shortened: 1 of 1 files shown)


RReverser:
Can't add as reviewers, so just /cc @ngzhian @tlively.

tlively (approved):

src/features/simd.md
```diff
@@ -50,31 +50,19 @@ import { simd } from 'wasm-feature-detect';
...
-
-To try this out, read the [origin trial developer guide](https://github.com/GoogleChrome/OriginTrials/blob/gh-pages/developer-guide.md), and [register for an origin trial token](https://developers.chrome.com/origintrials/#/view_trial/-4708513410415853567). More information about origin trials can be found in the [FAQ](https://github.com/GoogleChrome/OriginTrials/blob/gh-pages/developer-guide.md#faq). Please file a [bug](https://bugs.chromium.org/p/v8/issues/entry) if something isn't working as you expect. The origin trial is compatible with Emscripten versions 2.0.17 onwards.
+WebAssembly SIMD support is available by default from Chrome 91. Make sure to use the latest version of the toolchain as well as detailed below, as well as latest wasm-feature-detect to detect engines that support the final version of the spec. If something doesn’t look right, please [file a bug](https://crbug.com/v8).
```
dtig, line 55:
```suggestion
WebAssembly SIMD support is available by default from Chrome 91. Make sure to use the latest version of the toolchain as detailed below, as well as latest wasm-feature-detect to detect engines that support the final version of the spec. If something doesn’t look right, please [file a bug](https://crbug.com/v8).
```

src/features/simd.md
```diff
@@ -50,31 +50,19 @@ import { simd } from 'wasm-feature-detect';
...
 WebAssembly SIMD is available behind a flag in Firefox. Currently it's supported only on x86 and x86-64 architectures. To try out the SIMD support in Firefox, go to `about:config` and enable `javascript.options.wasm_simd`. Note that this feature is still experimental and being worked on.
 
-### Enabling experimental SIMD support in Node.js
-
-In Node.js WebAssembly SIMD can be enabled via `--experimental-wasm-simd` flag:
-
-```bash
-node --experimental-wasm-simd main.js
-```
-
```
dtig, line 72: Is there a reason this was removed? IIUC, node users would still need the flag until V8 9.1 is picked up by an official node version.

ngzhian:
LGTM! thanks :)

src/features/simd.md
RReverser: It's in the PR description - I checked and Node has version of SIMD before latest opcode renumbering, so won't work with final version of SIMD produced by up-to-date toolchains. I want to add it back once it updates the V8, but for now prefer not to mention it as supported.

src/features/simd.md
RReverser: Oops, thanks

src/features/simd.md
dtig: Sorry, missed the description - makes sense.

dtig (approved):

PR #559, opened 2021-06-11
Add a feature entry for Import Assertions / JSON modules

dandclark:
Add a feature entry for [Import Assertions](https://github.com/tc39/proposal-import-assertions) / [JSON modules](https://github.com/tc39/proposal-json-modules).


src/features/import-assertions.md
```diff
@@ -0,0 +1,76 @@
...
+# Dynamic `import()`
+
+Import assertions can also be passed to dynamic `import()` with a new second parameter:
+
+```javascript
+// foo.json
+{ "answer": 42 }
+
+// main.mjs
+const jsonModule = await import("foo.json", { assert: { type: "json" } });
```
jridgewell, line 58:
```suggestion
const jsonModule = await import("./foo.json", { assert: { type: "json" } });
```

src/features/import-assertions.md
```diff
@@ -0,0 +1,76 @@
...
+console.log(jsonModule.default.answer); // 42
+```
+
+The data is referenced through the `default` property because the JSON content is the default export of the module.
```
jridgewell, line 62: Nit: Explain that `default` isn't specific to JSON here, this is just how the any default export is accessed with dynamic import. Eg, even a `const obj = {answer: 42}; export default obj;` in regular JS would also be accessed as `namespace.default.answer`

jridgewell (approved):

src/features/import-assertions.md
RReverser: I don't think we want to deviate from main point of the content much here. At most, let's link to an external resource.

src/features/import-assertions.md
```diff
@@ -0,0 +1,76 @@
...
+---
+
+The new [import assertions](https://github.com/tc39/proposal-import-assertions) feature allows module import statements to include additional information alongside the module specifier. An initial use for the feature is to enable JSON documents to be imported as [JSON modules](https://github.com/tc39/proposal-json-modules):
+
+```javascript
+// foo.json
+{ "answer": 42 }
+
+// main.mjs
+import json from "./foo.json" assert { type: "json" };
```
RReverser, line 20: Nit: we use single-quoted in our codestyle.

(we really should automate this...)

src/features/import-assertions.md
dandclark: I added an [MDN link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#importing_defaults) about importing default exports, does that sound reasonable?

src/features/import-assertions.md
```diff
@@ -0,0 +1,76 @@
...
+
+A natural question to ask is why a JSON module couldn't simply be imported like this:
+
+```javascript
+import json from "./foo.json";
+```
+
+The web platform checks the MIME type of a module resource for validity prior to executing it, and in theory this MIME type could also be used to determine whether to treat the resource as a JSON or as a JavaScript module.
+
+But, there's an [escalation-of-privilege security issue](https://github.com/w3c/webcomponents/issues/839) with relying on the MIME type alone. Modules can be imported cross-origin, and a developer might import a JSON module from a third-party source. The developer might consider an import of JSON even from an untrusted  third-party to be basically safe as long as the JSON is properly sanitized, since importing JSON won't execute script. But, the third-party server could unexpectedly reply with a JavaScript MIME type and a malicious JavaScript payload, executing code in the importer's domain.
```
RReverser, line 34: I'd split this sentence up into smaller paragraphs for visual separation.

src/features/import-assertions.md
```diff
@@ -0,0 +1,76 @@
...
+
+```javascript
+// Executes JS if sketchy-third-party.com responds with
+// application/javascript MIME type!
+import data from "https://sketchy-third-party.com/data.json";
+```
+
+File extensions can't be used to make a module type determination because they [aren't a reliable indicator of content type on the web](https://github.com/tc39/proposal-import-assertions/blob/master/content-type-vs-file-extension.md). So instead, we use import assertions to indicate the expected module type and prevent this escalation-of-privilege pitfall. When a developer wants to import a JSON module, they must use an import assertion to specify that it's supposed to be JSON. The import will fail if the MIME type received from the network doesn't match the expected type.
```
RReverser, line 42: Same here, e.g.

```suggestion
File extensions can't be used to make a module type determination because they [aren't a reliable indicator of content type on the web](https://github.com/tc39/proposal-import-assertions/blob/master/content-type-vs-file-extension.md). So instead, we use import assertions to indicate the expected module type and prevent this escalation-of-privilege pitfall.

When a developer wants to import a JSON module, they must use an import assertion to specify that it's supposed to be JSON. The import will fail if the MIME type received from the network doesn't match the expected type:
```

src/features/import-assertions.md
```diff
@@ -0,0 +1,76 @@
...
+console.log(jsonModule.default.answer); // 42
+```
+
+The data is referenced through the `default` property because the JSON content is the [default export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#importing_defaults) of the module.
+
+## Conclusion
+
+Currently the only specified use of import assertions is for specifying module type. However, the feature was designed to allow arbitrary key/value assertion pairs, so additional uses may be added in the future if it becomes useful to restrain module imports in other ways.
+
+Meanwhile, JSON modules with the new import assertions syntax are available by default in Chromium 91. [CSS module scripts](https://chromestatus.com/feature/5948572598009856#) are also coming soon, using the same module-type-assertion syntax.
```
RReverser, line 68:
```suggestion
Meanwhile, JSON modules with the new import assertions syntax are available by default in Chromium 91. [CSS module scripts](https://chromestatus.com/feature/5948572598009856) are also coming soon, using the same module-type-assertion syntax.
```

src/features/import-assertions.md
```diff
@@ -0,0 +1,76 @@
+---
+title: 'Import Assertions'
```
RReverser, line 2:
```suggestion
title: 'Import assertions'
```

src/features/import-assertions.md
```diff
@@ -0,0 +1,76 @@
...
+author: 'Dan Clark ([@dandclark1](https://twitter.com/dandclark1))'
+avatars:
+  - 'dan-clark'
+date: 2021-06-10
```
RReverser, line 6: (TODO: update before merging as this should be publishing date)

RReverser:
This looks great, thanks! I left a few stylistic nits but otherwise we can publish Monday or Tuesday.

src/features/import-assertions.md
RReverser: Looks reasonable, yeah. I think though we could link to our own explainer - https://v8.dev/features/dynamic-import#static seems like a good reference.

dandclark: I switched the link to https://v8.dev/features/dynamic-import#dynamic, moved it to the first sentence in the section where it seems a bit more natural, and changed the wording a bit.

src/features/import-assertions.md
```diff
@@ -14,10 +14,10 @@ The new [import assertions](https://github.com/tc39/proposal-import-assertions)
 
 ```javascript
 // foo.json
-{ "answer": 42 }
+{ 'answer': 42 }
```
RReverser, line 17: Er, actually this one is JSON so probably should keep double quotes :)

src/features/import-assertions.md
```diff
@@ -71,8 +75,8 @@ Meanwhile, JSON modules with the new import assertions syntax are available by d
 
 ## Import Assertions support { #support }
 
-<feature-support chrome="91 https://chromestatus.com/feature/5765269513306112"
-                 firefox="no"
-                 safari="no"
-                 nodejs="no"
-                 babel="yes https://github.com/babel/babel/pull/12139"></feature-support>
+<feature-support chrome='91 https://chromestatus.com/feature/5765269513306112'
```
RReverser, line 78: Kinda same here - I think we keep double quotes for HTML-ish tags.

src/features/import-assertions.md
RReverser: Looks good!

src/features/import-assertions.md
dandclark: Oops, I switched it back here and in the other example, thanks :)

src/features/import-assertions.md
```diff
@@ -0,0 +1,82 @@
...
+---
+
+The new [import assertions](https://github.com/tc39/proposal-import-assertions) feature allows module import statements to include additional information alongside the module specifier. An initial use for the feature is to enable JSON documents to be imported as [JSON modules](https://github.com/tc39/proposal-json-modules):
+
+```javascript
+// foo.json
+{ "answer": 42 }
+
+// main.mjs
+import json from './foo.json' assert { type: 'json' };
```
RReverser, line 20: Perhaps worth splitting up?

````suggestion
```json
// foo.json
{ "answer": 42 }
```

```javascript
// main.mjs
import json from './foo.json' assert { type: 'json' };
````

File: src/features/import-assertions.md [hunk 12]
```diff
@@ -0,0 +1,82 @@
...
+## Dynamic `import()`
+
+Import assertions can also be passed to [dynamic `import()`](https://v8.dev/features/dynamic-import#dynamic) with a new second parameter:
+
+```javascript
+// foo.json
+{ "answer": 42 }
+
+// main.mjs
+const jsonModule = await import('./foo.json', { assert: { type: 'json' } });
```
RReverser, line 64: Just some visual separation.

```suggestion
const jsonModule = await import('./foo.json', {
  assert: { type: 'json' }
});
```

src/features/import-assertions.md
```diff
@@ -0,0 +1,82 @@
...
+The web platform checks the MIME type of a module resource for validity prior to executing it, and in theory this MIME type could also be used to determine whether to treat the resource as a JSON or as a JavaScript module.
+
+But, there's a [security issue](https://github.com/w3c/webcomponents/issues/839) with relying on the MIME type alone.
+
+Modules can be imported cross-origin, and a developer might import a JSON module from a third-party source. They might consider this to be basically safe even from an untrusted third-party as long as the JSON is properly sanitized, since importing JSON won't execute script.
+
+However, third-party script can actually execute in this scenario because the third-party server could unexpectedly reply with a JavaScript MIME type and a malicious JavaScript payload, running code in the importer's domain.
+
+```javascript
+// Executes JS if sketchy-third-party.com responds with
```
RReverser, line 41: I've noticed this causes scrollbars to appear on most devices for import statement. How about we use `evil.com` instead? (it's commonly used in CSP examples, so seems appropriate here too)

src/features/import-assertions.md
```diff
@@ -0,0 +1,82 @@
+---
+title: 'Import assertions'
+author: 'Dan Clark ([@dandclark1](https://twitter.com/dandclark1))'
```
RReverser, line 3: Ah, and one final suggestion: do you want to add a funny (on-topic or not) tagline here? You can see other blog posts for examples.

src/features/import-assertions.md
RReverser: @danelphick Ping - what do you think? I guess we'll publish this tomorrow then, but let's flesh out those final details before merging.

dandclark: I agree with this. I just pushed a change switching it to evil.com and it looks like this gets rid of the scrollbars, at least on my device.

RReverser: Oh, huh, Github didn't load the new commits and the conversation was unresolved too so I assumed these were still open. In terms of timing, I still think it's best if we publish tomorrow now though, but this looks good, thanks.

RReverser (approved):

src/features/import-assertions.md
dandclark: I bumped this to the 15, on the assumption that this will be published tomorrow.

RReverser:
Thank you, it's out! https://twitter.com/v8js/status/1404801114810236932

PR #560, opened 2021-06-15
Tweak MIME type example

mathiasbynens:
The HTML Standard recommends text/javascript and considers application/javascript a legacy alias: https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages:javascript-mime-type


Atom1985 (commented):
src/features/import-assertions.md

PR #561, opened 2021-06-15
Add tools tab

camillobruni:
Put http://v8.dev/tools in a more prominent spot on the main page.


surma (approved):

mathiasbynens:
Hmm, this is a top-level navigation menu item that points to a separate website, with a different design, hosted on a different origin — which is a little unusual. Could we instead make a `/tools` page with a list of links, similar to `/docs`? That way it would still feel like part of v8.dev.

This change also broke the design on narrow viewports:



This could be addressed by tweaking the media queries in the CSS.

@surma, WDYT?

surma:
> This change also broke the design on narrow viewports:

Ah good catch. Fixed via #565.

> Hmm, this is a top-level navigation menu item that points to a separate website, with a different design, hosted on a different origin

I don’t think this is an issue, tbqh. But yeah, if someone wants to add a subpage, I’m happy to review.

PR #571, opened 2021-07-21
Update @doNotGenerateCppClass docs

sethbrenith:
I just changed the default so we now specify when not to generate a CPP
class, rather than when to generate. This small change updates the
documentation to match (without attempting to patch all of the other
holes in the documentation).

### DIFF
--- a/src/docs/torque.md
+++ b/src/docs/torque.md
@@ -209,7 +209,7 @@ ClassDeclaration :
   }
 
 ClassAnnotation :
-  @generateCppClass
+  @doNotGenerateCppClass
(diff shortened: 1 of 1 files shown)


sethbrenith:
@nico-hartmann PTAL

nico-hartmann:
Change looks good to me, thanks!

nico-hartmann (approved):

sethbrenith:
@mathiasbynens would you please review? It seems that Nico doesn't have write access, since Github still says a review is required. Thanks!

mathiasbynens (approved):

PR #566, opened 2021-07-09
Fixed Firefox support link for support of cause options on Error Constructor

HolgerJeromin:
The linked Bugzilla entry is something else.
~~I did not found something about it in Bugzilla.~~

PR #576, opened 2021-08-09
Error cause: added start version for firefox

HolgerJeromin:
Follow up from #566 to avoid confusion that this is not supported in older versions.


RReverser (approved):
Thanks!

PR #567, opened 2021-07-12
Actualize core-js data

zloirock:
https://github.com/v8/v8.dev/pull/564#issuecomment-875916452

PR #597, opened 2021-10-19
Add feature explainer for Array#find and friends

syg:


zloirock:
@syg you missed core-js data [again](https://github.com/v8/v8.dev/pull/567). Could you check it next time before posts publishing?

PR #606, opened 2021-11-08
Add Oilpan library blog post

mlippautz:


src/blog/oilpan-library.md
```diff
@@ -0,0 +1,45 @@
...
+avatars:
+  - anton-bikineev
+  - omer-katz
+  - michael-lippautz
+date: 2021-11-09
+tags:
+  - internals
+  - memory
+  - cppgc
+description: 'In V8 v9.4 we are shipping Oilpan, a garbage collection library for C++.'
```
RReverser, line 13: I think it's more correct to say "we've shipped" here as it's been some time ago now? Alternatively, perhaps at this point we shouldn't mention V8 version in the meta description and shift the focus to the library itself.

mlippautz: Good point, I rephrased this a little removing the V8 version.

src/blog/oilpan-library.md
```diff
@@ -0,0 +1,45 @@
...
...troducing Oilpan in Blink was to aid integration into other garbage collected systems such as V8 which eventually materialized in implementing the [unified JavaScript and C++ heap](https://v8.dev/blog/tracing-js-dom) where Oilpan takes care of processing C++ objects. With more and more object hierarchies being managed and better integration with V8, Oilpan became increasingly complex over time and the team realized that they were reinventing the same concepts as in V8’s garbage collector and solving the same problems. On top of that, the integration in Blink required building around 30k targets to actually run a hello world garbage collection test. Early 2020, we started a journey in carving out Oilpan from Blink and encapsulating it into a library. We decided to host the code in V8, reuse abstractions where possible, and do some spring cleaning on the garbage collection interface. In addition to fixing all of the aforementioned issues, [a library](https://docs.google.com/document/d/1ylZ25WF82emOwmi_Pg-uU6BI1A-mIbX_MG9V87OFRD8/) would also enable other projects to make use of garbage-collected C++. We launched the library in V8 v9.4 and enabled it in Blink starting in Chromium M94.
```
RReverser, line 20:
```suggestion
So, what’s the deal with providing it as a library through V8? Blink, being forked from WebKit, originally used reference counting, a [well-known paradigm for C++ code](https://en.cppreference.com/w/cpp/memory/shared_ptr), for managing its on-heap memory. Reference counting is supposed to solve memory management issues but is known to be prone to memory leaks due to cycles. On top of this inherent problem, Blink also suffered from [use-after-free issues](https://en.wikipedia.org/wiki/Dangling_pointer) as sometimes reference counting would be omitted for performance reasons. Oilpan was initially developed specifically for Blink to simplify the programming model and get rid of memory leaks and use-after-free issues. We believe Oilpan succeeded in simplifying the model and also in making the code more secure. Another maybe less pronounced reason for introducing Oilpan in Blink was to aid integration into other garbage collected systems such as V8 which eventually materialized in implementing the [unified JavaScript and C++ heap](https://v8.dev/blog/tracing-js-dom) where Oilpan takes care of processing C++ objects. With more and more object hierarchies being managed and better integration with V8, Oilpan became increasingly complex over time and the team realized that they were reinventing the same concepts as in V8’s garbage collector and solving the same problems. On top of that, the integration in Blink required building around 30k targets to actually run a hello world garbage collection test. Early 2020, we started a journey in carving out Oilpan from Blink and encapsulating it into a library. We decided to host the code in V8, reuse abstractions where possible, and do some spring cleaning on the garbage collection interface. In addition to fixing all of the aforementioned issues, [a library](https://docs.google.com/document/d/1ylZ25WF82emOwmi_Pg-uU6BI1A-mIbX_MG9V87OFRD8/) would also enable other projects to make use of garbage-collected C++. We launched the library in V8 v9.4 and enabled it in Blink starting in Chromium M94.
```

RReverser:
Can you please break up text into smaller paragraphs (couple of sentences each) where makes sense to insert a break? It's usually hard for readers to comprehend long paragraphs like this with no splits:

image

(screenshot from preview link above)

src/blog/oilpan-library.md
mlippautz: Done

mlippautz:
> Can you please break up text into smaller paragraphs (couple of sentences each) where makes sense to insert a break? It's usually hard for readers to comprehend long paragraphs like this with no splits:
> 
> (screenshot from preview link above)

I split the paragraph. In general the paragraphs already lay out a story so I'd rather not clutter it further at this point.

RReverser:
> In general the paragraphs already lay out a story so I'd rather not clutter it further at this point.

Maybe, but for reader it's useful to have some visual separation to stop an eye on or return to if temporarily distracted from screen. It's harder to do that when you're in the middle of a long paragraph - in those cases, a more granular structure is better.

mlippautz:
> > In general the paragraphs already lay out a story so I'd rather not clutter it further at this point.
> 
> Maybe, but for reader it's useful to have some visual separation to stop an eye on or return to if temporarily distracted from screen. It's harder to do that when you're in the middle of a long paragraph - in those cases, a more granular structure is better.

In any case, the longer paragraph is split now. Does that work for you? (It initially captured the idea of "why a library?"; now that part is split into the different reasons as of why a library may be useful)

src/blog/oilpan-library.md
```diff
@@ -0,0 +1,51 @@
...
...o allows for reusing an existing V8 platform to create a heap for managed C++ objects. Garbage collections can be configured to run automatically, integrating in the task infrastructure or can be triggered explicitly considering the native stack as well. The idea is to allow embedders that just want managed C++ objects to avoid dealing with V8 as a whole, see this [hello world program](https://chromium.googlesource.com/v8/v8.git/+/main/samples/cppgc/hello-world.cc) as an example. Conveniently, tests for the core of Oilpan use this setup which means that it’s a matter of seconds to build and run a specific garbage collection test. As of today, [>400 of such unit tests](https://source.chromium.org/chromium/chromium/src/+/main:v8/test/unittests/heap/cppgc/) for the core of Oilpan exist. The setup also serves as a playground to experiment and try out new things and may be used to validate assumptions around raw performance. An embedder of this configuration is PDFium which uses Oilpan’s stand-alone version for [securing XFA](https://groups.google.com/a/chromium.org/g/chromium-dev/c/RAqBXZWsADo/m/9NH0uGqCAAAJ?utm_medium=email&utm_source=footer) which allows for more dynamic PDF content.
```
RReverser, line 32:
```suggestion
The core of Oilpan is available as a stand-alone C++ garbage collector in the `cppgc` namespace. The setup also allows for reusing an existing V8 platform to create a heap for managed C++ objects. Garbage collections can be configured to run automatically, integrating in the task infrastructure or can be triggered explicitly considering the native stack as well. The idea is to allow embedders that just want managed C++ objects to avoid dealing with V8 as a whole, see this [hello world program](https://chromium.googlesource.com/v8/v8.git/+/main/samples/cppgc/hello-world.cc) as an example. Conveniently, tests for the core of Oilpan use this setup which means that it’s a matter of seconds to build and run a specific garbage collection test. As of today, [>400 of such unit tests](https://source.chromium.org/chromium/chromium/src/+/main:v8/test/unittests/heap/cppgc/) for the core of Oilpan exist. The setup also serves as a playground to experiment and try out new things and may be used to validate assumptions around raw performance. An embedder of this configuration is PDFium which uses Oilpan’s stand-alone version for [securing XFA](https://groups.google.com/a/chromium.org/g/chromium-dev/c/RAqBXZWsADo/m/9NH0uGqCAAAJ?utm_medium=email&utm_source=footer) which allows for more dynamic PDF content.
```

This one caught my eye, but if there are other things that relate to code, it's best to wrap them into backticks too.

RReverser: I'd also split this paragraph into smaller pieces as well, e.g. the test one can be neatly split out into its own part:

```suggestion
The core of Oilpan is available as a stand-alone C++ garbage collector in the `cppgc` namespace. The setup also allows for reusing an existing V8 platform to create a heap for managed C++ objects. Garbage collections can be configured to run automatically, integrating in the task infrastructure or can be triggered explicitly considering the native stack as well. The idea is to allow embedders that just want managed C++ objects to avoid dealing with V8 as a whole, see this [hello world program](https://chromium.googlesource.com/v8/v8.git/+/main/samples/cppgc/hello-world.cc) as an example.

Conveniently, tests for the core of Oilpan use this setup which means that it’s a matter of seconds to build and run a specific garbage collection test. As of today, [>400 of such unit tests](https://source.chromium.org/chromium/chromium/src/+/main:v8/test/unittests/heap/cppgc/) for the core of Oilpan exist.

The setup also serves as a playground to experiment and try out new things and may be used to validate assumptions around raw performance. An embedder of this configuration is PDFium which uses Oilpan’s stand-alone version for [securing XFA](https://groups.google.com/a/chromium.org/g/chromium-dev/c/RAqBXZWsADo/m/9NH0uGqCAAAJ?utm_medium=email&utm_source=footer) which allows for more dynamic PDF content.
```

RReverser: Re: code references - I see a few more namespaces mentioned throughout the post; it's best to add backticks in those so that they remain links but also clearly are marked as code.

RReverser: Looks like this was accidentally marked as resolved - the other namespaces still need to be wrapped too as per last comment.

mlippautz: Yeah, indeed.

- I now added more ticks around C++ names.
- I split the paragraph but did so slightly differently.

RReverser: Thanks, looks great!

RReverser (approved):

RReverser:
Too late for this one, but please don't merge posts - I usually do this at the optimal publishing time myself.

PR #607, opened 2021-11-10
No longer include PWAcompat

tomayac:
🎉 https://bugs.webkit.org/show_bug.cgi?id=231339 Seems like the proper fix is coming.


tomayac:
This addresses https://github.com/v8/v8.dev/pull/310#issuecomment-965096774.

mathiasbynens (approved):

mathiasbynens:
Awesome that this is finally getting fixed at the WebKit level!

PR #622, opened 2022-02-01
Update merge-patch.md

lutzvahl:
Added details how to resolve conflicts during the merge

@verwaest PTAL as well

### DIFF
--- a/src/docs/merge-patch.md
+++ b/src/docs/merge-patch.md
@@ -54,11 +54,10 @@ Note that this option only works if the patch applies cleanly on the release bra
 1. Open the CL you want to back-merge.
 1. Select "Cherry pick" from the extended menu (three vertical dots in the upper right corner).
 1. Enter "refs/branch-heads/*X.X*" as destination branch (replace *X.X* by the proper branch).
-1. In case of merge conflict, please also go ahead and create the CL
 1. Modify the commit message:
    1. Prefix the title with "Merged: ".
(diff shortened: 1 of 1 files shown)


src/docs/merge-patch.md
```diff
@@ -54,10 +54,11 @@ Note that this option only works if the patch applies cleanly on the release bra
 1. Open the CL you want to back-merge.
 1. Select "Cherry pick" from the extended menu (three vertical dots in the upper right corner).
 1. Enter "refs/branch-heads/*X.X*" as destination branch (replace *X.X* by the proper branch).
+1. In case of merge conflict, please also go ahead and create the CL
```
backes, line 57: I think this step should be moved down to *after* modifying the commit message (which still happens before actually creating the cherry-pick, thus before the merge conflicts are detected).

As there is already an item below about resolving conflicts, maybe we can also just drop this item here completely.

lutzvahl: Good catch - just tried the flow and you're right. Adjusted the ordering to reflect the flow

lutzvahl:
Adjusted the ordering in the gerrit UI flow

backes:
Thanks for double-checking, and adjusting!

Can you also fix the punctuation such that all items end in a period?

backes (approved):
Thanks, LGTM.

verwaest:
lgtm, thanks!

PR #633, opened 2022-04-14
Add class features improvements blog post

joyeecheung:
cc @syg who suggested me to write a guest blog post about the class features work we've been doing in V8.

I am hoping this can be published before next Tuesday when the Node.js 18 release (which includes the work mentioned in this post) goes out, so I can link to this post in the Node.js's release announcement which lists the V8 updates the new release has, thanks :)


src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+}
+```
+
+To implement the specified semantics before the proposal finalized, V8 used calls to runtime functions since they are more flexible. As shown in the bytecode above, the initialization of public fields was implemented with `%CreateDataProperty()` runtime calls, while the initialization of private fields was implemented with `%AddPrivateField()`. Since calling into the runtime incurs a significant overhead, the initialization of class fields was significantly slower compared to the assignment of ordinary object properties.
+
+In most use cases, however, the semantic differences are insignificant. It would be nice to have the performance of the optimized assignments of properties when the pattern of the class field initialization is predictable enough - so a more optimal implementation was created after the proposal finalized.
+
+### Optimizing private class fields and computed public class fields
+
+To speed up initialization of private class fields and computed public class fields, the implementation introduced a new machinery to plug into the [inline cache (IC) system](https://mathiasbynens.be/notes/shapes-ics) when handling these operations, including:
```
syg, line 121:
```suggestion
To speed up initialization of private class fields and computed public class fields, the implementation introduced a new machinery to plug into the [inline cache (IC) system](https://mathiasbynens.be/notes/shapes-ics) when handling these operations. This new machinery comes in three cooperating pieces:
```

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+
+- A new bytecode `DefineKeyedOwnProperty`. This gets emitted when generating code for the `ClassLiteral::Property` AST nodes representing class field initializers.
```
syg, line 123:
```suggestion
- In the interpreter, a new bytecode `DefineKeyedOwnProperty`. This gets emitted when generating code for the `ClassLiteral::Property` AST nodes representing class field initializers.
```

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+- A corresponding Turbofan IR opcode `JSDefineKeyedOwnProperty`, which can be compiled from the new bytecode.
```
syg, line 124:
```suggestion
- In the TurboFan JIT, a corresponding IR opcode `JSDefineKeyedOwnProperty`, which can be compiled from the new bytecode.
```

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+- A new `DefineKeyedOwnIC` which is used in the interpreter handler of the new bytecode as well as the code compiled from the new IR opcode. To simplify the implementation, the new IC reuses some of the code in `KeyedStoreIC` which was intended for ordinary property stores.
```
syg, line 125:
```suggestion
- In the IC system, a new `DefineKeyedOwnIC` that is used in the interpreter handler of the new bytecode as well as the code compiled from the new IR opcode. To simplify the implementation, the new IC reuses some of the code in `KeyedStoreIC` which was intended for ordinary property stores.
```

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+// the property keyed by the private name symbol `#a` in the instance,
+// that is, `#a = 0`.
+LdaZero
+DefineKeyedOwnProperty <this>, r0, [0]
+```
+
+When the initializer is executed for enough times, V8 allocates one [feedback vector slot](https://ponyfoo.com/articles/an-introduction-to-speculative-optimization-in-v8) for each field being initialized. The slot contains the key of the field being added (in the case of the private field, the private name symbol) and a pair of [hidden classes](https://v8.dev/blog/fast-properties) between which the instance has been transitioning as the result of field initialization. In subsequent initializations, the IC uses the feedback to see if the fields are initialized in the same order on instances with the same hidden classes. If the initialization matches the pattern that V8 has seen before (which is usually the case), V8 takes the fast path and performs the initialization with pre-generated code instead of calling into the runtime, thus speeding up the operation. If the initialization does not match a pattern that V8 has seen before, it falls back to a runtime call to deal with the slow cases.
```
syg, line 149:
```suggestion
When the initializer is executed enough times, V8 allocates one [feedback vector slot](https://ponyfoo.com/articles/an-introduction-to-speculative-optimization-in-v8) for each field being initialized. The slot contains the key of the field being added (in the case of the private field, the private name symbol) and a pair of [hidden classes](https://v8.dev/docs/hidden-classes) between which the instance has been transitioning as the result of field initialization. In subsequent initializations, the IC uses the feedback to see if the fields are initialized in the same order on instances with the same hidden classes. If the initialization matches the pattern that V8 has seen before (which is usually the case), V8 takes the fast path and performs the initialization with pre-generated code instead of calling into the runtime, thus speeding up the operation. If the initialization does not match a pattern that V8 has seen before, it falls back to a runtime call to deal with the slow cases.
```

Camillo just wrote a new article. :)

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+
+### Optimizing named public class fields
+
+To speed up initialization of named public class fields, we reused the existing `DefineNamedOwnProperty` bytecode which calls into `DefineNamedOwnIC` either in the interpreter or through the code compiled from the `JSDefineNamedOwnProperty` IR opcode. These were intended for object literal initializations and essentially implement a specialized case of the `[[CreateDataProperty]]` semantics which is also what the named public class fields are specified to use.
```
syg, line 153:
```suggestion
To speed up initialization of named public class fields, we reused the existing `DefineNamedOwnProperty` bytecode which calls into `DefineNamedOwnIC` either in the interpreter or through the code compiled from the `JSDefineNamedOwnProperty` IR opcode.
```

I'd get move the historical note below where it's needed.

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+// Load the value of the property keyed by `#a` from the instance into r2
+// Note: LdaKeyedProperty is renamed to GetKeyedProperty in the refactoring
+GetKeyedProperty <this>, [2]
+
+// Use the DefineKeyedOwnProperty bytecode to store the property keyed
+// by `#a` as the value of the property keyed by `b`, that is, `b = this.#a;`
+DefineNamedOwnProperty <this>, [0], [4]
+```
+
+The original `DefineNamedOwnIC` machinery could not be simply plugged into the handling of the named public class fields, as it needed a few patches to deal with an edge case of the class fields. Previously it expected the target being initialized to be an object created by V8, which was always true for object literals, but the class fields can be initialized on user-defined objects when the class extends a base class whose constructor overrides the target:
```
syg, line 179:
```suggestion
The original `DefineNamedOwnIC` machinery could not be simply plugged into the handling of the named public class fields, since it was originally intended only for object literal initialization. Previously it expected the target being initialized to be an object created by V8, which was always true for object literals, but the class fields can be initialized on user-defined objects when the class extends a base class whose constructor overrides the target:
```

I don't follow the distinction between "object created by V8" and "user-defined objects". I think you mean an ordinary object with no own properties or something?

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+
+To deal with these targets, we patched the IC to fall back to the runtime when it sees that the object being initialized is a proxy, if the field being defined already exists on the object, or if the object just has a hidden class that the IC has not seen before. It is still possible to optimize the edge cases if they become common enough, but so far it seems better to trade the performance of them for simplicity of the implementation.
+
+## Optimizing private methods
+
+### The implementation of private methods
+
+In [the specification](https://tc39.es/ecma262/#sec-privatemethodoraccessoradd), the private methods are described as if they are installed on the instances but not on the class. In order to save memory, however, V8's implementation stores the private methods along with a private brand symbol in a context associated with the class. When the constructor is invoked, V8 only stores a reference to that context in the instance, with the private brand symbol as the key.
+
+![Evaluation and instantiation of classes with private methods](/_img/better-class-features/class-evaluation-and-instantiation.svg)
```
syg, line 216: Great graphic! I wonder if adding another instance would make it even clearer that the context is shared by multiple instances.

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+LdaImmutableCurrentContextSlot [3]
+Star0
+
+// Use the DefineKeyedOwnProperty bytecode to store the
+// context in the instance with the private brand as key
+Ldar <context>
+DefineKeyedOwnProperty <this>, r0, [0]
+```
+
+![Performance of initialization of instances with private methods](/_img/better-class-features/private-methods-performance.svg)
```
syg, line 265: Ditto on these graphs to label the OS + arch + general performance profile info.

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+class B extends A {
+  #method() {}
+  callMethod() { return this.#method(); }
+  constructor(o) {
+    super();
+  }
+};
+```
+
+As described before, when initializing the brand, V8 also stores a reference to the class context in the instance. This reference isn't used in brand checks, it's intended for the debugger to retrieve a list of private methods from the instance without knowing which class it is constructed from. When `super()` is invoked directly in the constructor, V8 can simply load the context from the context register (which is what `Mov <context>, r2` or `Ldar <context>` in the bytecodes above does) to perform the initialization, but `super()` can also be invoked from a nested arrow function, which in turn can be invoked from a different context. In this case, V8 falls back to a runtime function (still named `%AddPrivateBrand()`) to look for the class context in the context chain instead of relying on the context register. For example, for the `callSuper` function below:
```
syg, line 286:
```suggestion
As described before, when initializing the brand, V8 also stores a reference to the class context in the instance. This reference isn't used in brand checks, but is instead intended for the debugger to retrieve a list of private methods from the instance without knowing which class it is constructed from. When `super()` is invoked directly in the constructor, V8 can simply load the context from the context register (which is what `Mov <context>, r2` or `Ldar <context>` in the bytecodes above does) to perform the initialization, but `super()` can also be invoked from a nested arrow function, which in turn can be invoked from a different context. In this case, V8 falls back to a runtime function (still named `%AddPrivateBrand()`) to look for the class context in the context chain instead of relying on the context register. For example, for the `callSuper` function below:
```

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+
+## Final notes
+
+The work mentioned in this blog post is also included in the upcoming Node.js v18 release. Previously, Node.js switched to symbol properties in a few built-in classes that had been using private fields in order to include them into the embedded bootstrap snapshot as well as to improve the performance of the constructors (see [this blog post](https://www.nearform.com/blog/node-js-and-the-struggles-of-being-an-eventtarget/) for more context). With the improved support of class features in V8, Node.js [switched back to private class fields](https://github.com/nodejs/node/pull/42361) in these classes and Node.js's benchmarks showed that [these changes did not introduce any performance regressions](https://github.com/nodejs/node/pull/42361#issuecomment-1068961385).
+
+We'd like to thank Bloomberg for sponsoring Igalia to work on improvements of the class features and startup snapshots.
```
syg, line 332:
```suggestion
This work was contributed to V8 by Igalia under the sponsorship of Bloomberg.
```
@lutzvahl Any concerns about my suggested wording and calling out who did the work?

syg (commented):
Second pass. Looks great generally. Some phrasing suggestions for flow.

src/blog/better-class-features.md
joyeecheung: I am wondering if it's accurate to phrase it as an interpreter thing, as the bytecode is also compiled by the compilers. Perhaps `In the bytecode generator` would be more accurate?

src/blog/better-class-features.md
joyeecheung: I meant "an object that has not yet been touched by the user since its creation", that's probably clearer

joyeecheung (commented):
Thanks, will get to the remaining comments later

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+      enumerable: true,
+      value: this[_a]
+    });
+  }
+}
+```
+
+To implement the specified semantics before the proposal finalized, V8 used calls to runtime functions since they are more flexible. As shown in the bytecode above, the initialization of public fields was implemented with `%CreateDataProperty()` runtime calls, while the initialization of private fields was implemented with `%AddPrivateField()`. Since calling into the runtime incurs a significant overhead, the initialization of class fields was significantly slower compared to the assignment of ordinary object properties.
+
+In most use cases, however, the semantic differences are insignificant. It would be nice to have the performance of the optimized assignments of properties when the pattern of the class field initialization is predictable enough - so a more optimal implementation was created after the proposal finalized.
```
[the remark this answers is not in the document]
joyeecheung: Switched to `in these cases` to echo the `In most use cases` before.

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
...
+
+The first issue has been fixed in V8 v9.7 and the fix for the second issue has been released in V8 v10.0. This post covers how the first issue was fixed, for another read about how the fix of the snapshot issue, check out [this post](https://joyeecheung.github.io/blog/2022/04/14/fixing-snapshot-support-of-class-fields-in-v8/).
+
+## Optimizing class fields
+
+To get rid of the performance gap between the assignment of ordinary properties and the initialization of class fields, we updated the existing [inline cache (IC) system](https://mathiasbynens.be/notes/shapes-ics) to work with the latter. Before v9.7, V8 always used a costly runtime call for class field initializations. With v9.7, when V8 considers the pattern of the initialization to be predictable enough, it uses a new IC to speed up the operation just like what it does for assignments of ordinary properties.
+
+![Performance of initialization, optimized](/_img/better-class-features/class-fields-performance-optimized.svg)
```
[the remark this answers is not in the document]
joyeecheung: I included the OS and CPU model in the label, the latter should give away the performance level as well as the arch

src/blog/better-class-features.md
syg: Fair point. `In the bytecode generator` sgtm.

src/blog/better-class-features.md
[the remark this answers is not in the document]
syg: Looks good! There's a typo with "Restuls".

src/blog/better-class-features.md
syg: That is indeed clearer, let's go with that.

src/blog/better-class-features.md
joyeecheung: This looks like the only remaining unresolved comment with the blog post now - so pinging @lutzvahl, any comments about this?

syg: Lutz and I chatted internally. Let's go with the historically used phrasing "Thanks to Igalia and Bloomberg for contributing this implementation!"

joyeecheung: Done

joyeecheung:
I think I've resolved all the outstanding comments, also removed the `upcoming` in `..in the upcoming Node.js v18 release` since it's just released. Shu, can you take a look again? Thanks!

src/blog/better-class-features.md
```diff
@@ -0,0 +1,332 @@
+---
+title: 'Better support for class features'
```
syg, line 2:
```suggestion
title: 'Faster class field initialization'
```

I think the original title made more sense with the snapshot stuff also in scope, which we've since moved out to the Node article.

Please also rename the blog markdown file + asset folder name.

syg (approved):
lgtm! Will merge after title rename.

src/blog/better-class-features.md
joyeecheung: Since the post also includes how initialization of private brands was optimized maybe “Faster instance initialization with new class features”  (and “faster-class-features” for the file name) would be better?

joyeecheung: I took the liberty to rename the title to `Faster initialization of instances with new class features` and added a link to the Node.js release announcement before I log off - please let me know if you think that the original suggestion is better, thanks

syg: wfm, thanks

PR #639, opened 2022-05-24
Update compile-arm64.md instructions.

planglois:
This updates instructions for compiling V8 on an arm64 Linux host:

- Clarify that this page only applies to Arm64 Linux hosts.

- No longer recommend building GN from source, now that `gclient sync` downloads
  a arm64 binary for it.

- Use the --host-cc and --host-cxx flags to build clang.

### DIFF
--- a/src/docs/compile-arm64.md
+++ b/src/docs/compile-arm64.md
@@ -1,8 +1,8 @@
 ---
-title: 'Compiling on Arm64'
-description: 'Tips and tricks to build V8 natively on Arm64'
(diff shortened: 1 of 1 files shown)


planglois:
Hi, I realized I probably should try and assign a reviewer for this, however I'm not sure I'm able to.

@backes Could you take a look? I recently had to fix the scripts to allow bootstrapping clang from source, so I thought I'd reflect the changes here too.

Thanks!
Pierre

backes (approved):

planglois:
Thanks!

PR #642, opened 2022-06-14
Add RegExp `v` flag feature article

mathiasbynens:
cc @markusicu @macchiati


src/features/regexp-v-flag.md
```diff
@@ -0,0 +1,252 @@
...
+- [set notation + string literal syntax](/features/regexp-v-flag#set-notation)
+- [improved case-insensitive matching](/features/regexp-v-flag#ignoreCase)
+
+This article dives into each of these. But first things first — here’s how to use the new flag:
+
+```js
+const re = /…/v;
+```
+
+The `v` flag can be combined with existing regular expression flags, with one notable exception. The `v` flag enables all the good parts of the `u` flag, but with additional features and improvements. Crucially, `v` is a completely separate mode from `u` rather than a complementary one. For this reason, the `v` and `u` flags cannot be combined — trying to use both flags on the same regular expression results in an error. The only valid options are: either use `u`, or use `v`, or use neither `u` nor `v`. But since `v` is the most feature-complete option, that choice is easily made…
```
syg, line 27: The "The `v` flag enables all the good parts of the `u` flag, but with additional features and improvements." sentence makes it sound like `v` is a strict feature superset of `u`.

Maybe "The `v` flag enables all the good parts of the `u` flag, albeit with additional incompatible features and improvements."?

src/features/regexp-v-flag.md
```diff
@@ -0,0 +1,252 @@
...
+// Match an emoji that consists of just 1 code point:
+re.test('⚽'); // '\u26BD'
+// → true ✅
+
+// Match an emoji that consists of multiple code points:
+re.test('👨🏾‍⚕️'); // '\u{1F468}\u{1F3FE}\u200D\u2695\uFE0F'
+// → false ❌
+```
+
+In the above example, the regular expression doesn’t match the 👨🏾‍⚕️ emoji because it happens to consist of multiple code points, and `Emoji` is a Unicode _character_ property.
```
syg, line 58: Side comment on "character": is "Unicode character property" the what Unicode calls it officially? Between concepts of "code unit" and "code point", I thought calling stuff "characters" was discouraged.

src/features/regexp-v-flag.md
```diff
@@ -0,0 +1,252 @@
...
+
+*Note:* Although properties of strings are currently gated on the new `v` flag, [we plan to eventually make them available in `u` mode as well](https://github.com/tc39/proposal-regexp-v-flag/issues/49).
+
+## Set notation + string literal syntax { #set-notation }
+
+When working with `\p{…}` escapes (be it character properties or the new properties of strings) it can be useful to perform difference/subtraction or intersection. With the `v` flag, character classes can now be nested, and those set operations can now be performed within them.
+
+### Difference/subtraction with `--` { #difference }
+
+The syntax `A&&B` can be used to match strings _in `A` but not in `B`_, a.k.a. difference/subtraction. For example, what if you want to match all Greek symbols except for the letter `π`? Previously, there was no way to benefit from `\p{Script_Extensions=Greek}` in this case, since it already matches `π`. The only way to achieve this was to go back to hand-writing a lengthy character class containing the individual characters or ranges. And if Unicode added another Greek character, you would need to notice, and manually update your character class.
```
syg, line 96:
```suggestion
The syntax `A--B` can be used to match strings _in `A` but not in `B`_, a.k.a. difference/subtraction. For example, what if you want to match all Greek symbols except for the letter `π`? Previously, there was no way to benefit from `\p{Script_Extensions=Greek}` in this case, since it already matches `π`. The only way to achieve this was to go back to hand-writing a lengthy character class containing the individual characters or ranges. And if Unicode added another Greek character, you would need to notice, and manually update your character class.
```

syg (approved):
This article is great, thanks!

src/features/regexp-v-flag.md
mathiasbynens: Or

> The `v` flag enables all the good parts of the `u` flag, but with additional features and improvements — some of which are backwards-incompatible with the `u` flag.

src/features/regexp-v-flag.md
mathiasbynens: https://unicode.org/reports/tr18/#domain_of_properties

> Typically the important information is for the subset of the code points that are characters; therefore, those properties are often also called properties of characters.
> 
> In addition to properties of characters, there are also properties of strings (sequences of characters). A property of strings is more general than a property of characters. In other words, any property of characters is also a property of strings; its domain is, however, limited to strings consisting of a single character.

mathiasbynens: Perhaps @markusicu @macchiati have some guidance here?

src/features/regexp-v-flag.md
macchiati: Agree with @mathiasbynens

src/features/regexp-v-flag.md
macchiati: The quotation shows the reason.

src/features/regexp-v-flag.md
mathiasbynens:
```suggestion
The `v` flag can be combined with existing regular expression flags, with one notable exception. The `v` flag enables all the good parts of the `u` flag, but with additional features and improvements — some of which are backwards-incompatible with the `u` flag. Crucially, `v` is a completely separate mode from `u` rather than a complementary one. For this reason, the `v` and `u` flags cannot be combined — trying to use both flags on the same regular expression results in an error. The only valid options are: either use `u`, or use `v`, or use neither `u` nor `v`. But since `v` is the most feature-complete option, that choice is easily made…
```

mathiasbynens: Suggestion applied. Thanks for pointing this out, @syg!

src/features/regexp-v-flag.md
```diff
@@ -0,0 +1,252 @@
+---
+title: 'RegExp `v` flag with set notation and properties of strings'
+author: 'Mark Davis ([@mark_e_davis](https://twitter.com/mark_e_davis)), Markus Scherer, and Mathias Bynens ([@mathias](https://twitter.com/mathias))'
+avatars:
+  - 'mark-davis'
+  - 'markus-scherer'
+  - 'mathias-bynens'
+date: 2022-06-15
```
mathiasbynens, line 8:
```suggestion
date: 2022-06-27
```

mathiasbynens: I plan on publishing this tomorrow.

PR #650, opened 2022-06-27
Use SVG for assets in *Scan article

mathiasbynens:
Thanks for sharing the source docs!

I exported SVG versions, gave them a white (instead of transparent) background so they work well in both light/dark mode, cropped them using [svgcrop.com](https://svgcrop.com/), and finally optimized them using [SVGOMG](https://jakearchibald.github.io/svgomg/).


AntonBikineev (approved):
Thanks a lot!

PR #651, opened 2022-06-27
Link Babel repl in regexp-v-flag article

nicolo-ribaudo:
cc @mathiasbynens


mathiasbynens:
Thanks for kicking this off! I was thinking of mentioning this in the support section where we already mention Babel. I've gone ahead and made this change. Hopefully this is cool with you :) Thanks again for making this work in the REPL!

mathiasbynens (approved):

PR #667, opened 2022-09-18
add return type explicitly to Torque sample code

Syuparn:
Since implicit return type `void` is deprecated,  `void` is added explicitly to the Torque hello world sample code.

```
Torque Error: Default void return types are deprecated. Add `: void`.
```


tebbi (approved):
Thanks!

PR #686, opened 2023-03-01
Update embed.md

sykickDuck:
Add compiler option to enable sandbox so that you no longer get the error: # Embedder-vs-V8 build configuration mismatch. On embedder side sandbox is DISABLED while on V8 side it's ENABLED.


camillobruni:
This has been fixed in PR #783

PR #691, opened 2023-04-04
Wasm tail calls

thibaudmichaud:
PTAL, this adds a blog post entry about wasm tail calls (shipped in V8 11.2).


src/blog/wasm-tail-call.md
```diff
@@ -0,0 +1,164 @@
...
+With a regular call, this consumes O(n) stack space: each element of the list adds a new frame on the call stack. With a long enough list, this could very quickly overflow the stack. By replacing the call with a jump, tail-call optimization effectively turns this recursive function into a loop which uses
+O(1) stack space:
+
+```c
+int sum(List* list, int acc) {
+  while (list != nullptr) {
+    acc = acc + list->val;
+    list = list->next;
+  }
+}
```
backes, line 35: Missing: `return acc;`

src/blog/wasm-tail-call.md
```diff
@@ -0,0 +1,164 @@
...
+
+int main() {
+  printf("is_even(1000000): %d\n", is_even(1000000));
+}
+```
+
+Note that both of these examples are simple enough that if we compile with -O2, the compiler can precompute the answer and avoid exhausting the stack even without tail calls, but this wouldn't be the case with more complex code. In real-world code, the musttail attribute can be helpful for writing high-performance interpreter loops as described in this blog post:
+<https://blog.reverberate.org/2021/04/21/musttail-efficient-interpreters.html>.
+
+Besides the `musttail` attribute, C++ depends on tail calls for one other feature: C++20 coroutines. The relationship between tail calls and C++20 coroutines is covered in extreme depth in this blog post: https://lewissbaker.github.io/2020/05/11/understanding_symmetric_transfer, but to summarize, it is possible to use coroutines in a pattern that would subtly cause stack overflow even though the source code doesn't make it look like there is a problem. To fix this problem, the C++ committee added a requirement that compilers implement "symmetric transfer" to avoid the stack overflow, which in practice means using tail calls under the covers.
```
backes: The formatting of the link inside the sentence looks weird. Should we make "this blog post" the link text, and then change it to "a blog post by Lewis Baker"?

src/blog/wasm-tail-call.md
```diff
@@ -0,0 +1,165 @@
...
+}
+```
+
+This optimization is particularly important for functional languages. They rely heavily on recursive functions, and pure ones like Haskell don’t even provide loop control structures. Any kind of custom iteration typically uses recursion one way or another. Without tail-call optimization, this would very quickly run into a stack overflow for any non-trivial program.
+
+### The WebAssembly Tail Call Proposal
+
+There are two ways to call a function in Wasm MVP: `call` and `call_indirect`.  The Tail Call proposal adds their tail-call counterparts: `return_call` and `return_call_indirect`. This means that it is the responsibility of the toolchain to actually perform tail-call optimization and emit the appropriate call kind, which gives it more control over performance and stack space usage.
+
+Let’s look at a recursive fibonacci function. The Wasm bytecode is included here in the text format for completeness, but you can find it in C++ in the next section:
```
backes, line 44: There is a mix of `’` and `'` used in this file. We should clean that up (and always use `'`?).

src/blog/wasm-tail-call.md
```diff
@@ -0,0 +1,165 @@
...
+
+![Complex tail call in TurboFan](/_img/tail-call-complex.png)
+
+TurboFan handles these stack and register manipulations with the “gap resolver”, a component which takes a list of moves that should semantically be executed in parallel, and generates the appropriate sequence of moves to resolve potential interferences between the move’s sources and destinations. If the conflicts are acyclic, this is just a matter of reordering the moves such that all sources are read before they are overwritten. For cyclic conflicts (e.g. if we swap two stack parameters), this can involve moving one of the sources to a temporary register or a temporary stack slot to break the cycle.
+
+Tail calls are also supported in Liftoff, our baseline compiler. However they are not optimized in this tier: Liftoff will push the parameters, return address and frame pointer to complete the frame as if this was a regular call, and then shift everything downwards to discard the caller frame:
```
backes, line 159: Should we mention that we *have to* implement tail calls properly in Liftoff, to avoid running out of stack space?

backes (approved):
Nice! LGTM, left some comments.

src/blog/wasm-tail-call.md
thibaudmichaud: Good catch, thanks! Done.

src/blog/wasm-tail-call.md
thibaudmichaud: Done.

src/blog/wasm-tail-call.md
thibaudmichaud: Done. There is also the issue with tail-calls inside a try block, which wouldn't work as expected if we did regular calls in Liftoff, but I'm not sure we want to be so technical in this article...

backes: I don't think we need to be that specific. But saying that return calls must not be implemented via regular calls is an important point.

src/blog/wasm-tail-call.md
```diff
@@ -0,0 +1,165 @@
...
+
+This is a simple case where all parameters and return values fit into registers, and the callee has the same signature as the caller. In the general case, we might need to do complex stack manipulations:
+
+- Read outgoing parameters from the old frame
+- Move parameters into the new frame
+- Adjust the frame size by moving the return address up or down, depending on the number of stack parameters in the callee
+
+All these reads and writes can conflict with each other, because we are reusing the same stack space. This is a crucial difference with a non-tail call, which would simply push all the stack parameters and the return address on top of the stack.
+
+![Complex tail call in TurboFan](/_img/tail-call-complex.png)
```
mathiasbynens, line 155: Please export these images as SVG and then run them through https://jakearchibald.github.io/svgomg/ and use those instead of the PNGs.

src/blog/wasm-tail-call.md
```diff
@@ -0,0 +1,165 @@
+---
+title: 'WebAssembly Tail Calls'
+description: 'This document explains the WebAssembly Tail Calls proposal and
+demonstrates it with some examples'
+author: 'Thibaud Michaud, Thomas Lively'
+date: 2023-04-04
+tags:
+  - WebAssembly
+---
+We are shipping WebAssembly Tail Calls in V8 11.2! In this post we give a brief overview of this proposal, demonstrate an interesting use case for C++ coroutines with Emscripten, and show how V8 handles tail calls internally.
```
mathiasbynens, line 10: For consistency with how we do this in all other blog posts:

```suggestion
We are shipping WebAssembly Tail Calls in V8 v11.2! In this post we give a brief overview of this proposal, demonstrate an interesting use case for C++ coroutines with Emscripten, and show how V8 handles tail calls internally.
```

src/blog/wasm-tail-call.md
```diff
@@ -0,0 +1,165 @@
+---
+title: 'WebAssembly Tail Calls'
+description: 'This document explains the WebAssembly Tail Calls proposal and
+demonstrates it with some examples'
```
mathiasbynens, line 4: Semantic newlines only:

```suggestion
description: 'This document explains the WebAssembly Tail Calls proposal and demonstrates it with some examples'
```

src/blog/wasm-tail-call.md
```diff
@@ -0,0 +1,165 @@
...
+
+This optimization is particularly important for functional languages. They rely heavily on recursive functions, and pure ones like Haskell don't even provide loop control structures. Any kind of custom iteration typically uses recursion one way or another. Without tail-call optimization, this would very quickly run into a stack overflow for any non-trivial program.
+
+### The WebAssembly Tail Call Proposal
+
+There are two ways to call a function in Wasm MVP: `call` and `call_indirect`.  The Tail Call proposal adds their tail-call counterparts: `return_call` and `return_call_indirect`. This means that it is the responsibility of the toolchain to actually perform tail-call optimization and emit the appropriate call kind, which gives it more control over performance and stack space usage.
+
+Let's look at a recursive fibonacci function. The Wasm bytecode is included here in the text format for completeness, but you can find it in C++ in the next section:
+
+<pre>
```
mathiasbynens, line 47: Hmmm, can we not use triple backticks here? That way it gets processed and style exactly the same way as other code blocks.

If no specific supported syntax-highlighting language applies, you can use

```
```text
```

mathiasbynens: Alternatively, we could add a custom WAT/WAST highlighter to https://github.com/v8/v8.dev/blob/main/prism-languages.js

mathiasbynens: Hmm, it looks like Prism supports `wasm` nowadays: https://github.com/PrismJS/prism/tree/master/tests/languages/wasm Have you tried that?

thibaudmichaud: `wasm` looks like it does work! I also wanted to highlight the `return_call` instruction, which it looks like I can't do in a code block, but it's not super important.

src/blog/wasm-tail-call.md
thibaudmichaud: Done.

src/blog/wasm-tail-call.md
mathiasbynens: You can! Play around with this syntax using line offsets:

```
```wasm/2
or
```wasm/2,7
```

thibaudmichaud: Neat, thanks! Done

mathiasbynens (approved):

PR #693, opened 2023-04-06
Minor tweaks for consistency with other v8.dev content

mathiasbynens:


src/blog/wasm-tail-call.md
```diff
@@ -1,15 +1,14 @@
 ---
-title: 'WebAssembly Tail Calls'
-description: 'This document explains the WebAssembly Tail Calls proposal and demonstrates it with some examples'
+title: 'WebAssembly tail calls'
```
mathiasbynens, line 2: We generally use sentence case, not title case, for headings. Must “WebAssembly Tail Calls” be capitalized at all times? The current post uses a mix of both “Tail Calls” and “tail calls” so I’m not sure.

thibaudmichaud: Sgtm if it's more consistent with v8.dev, I don't think there is any official guidance about this for wasm proposals.

thibaudmichaud (approved):

PR #695, opened 2023-04-17
debug : footer-line break issue

wooseok123:
### changes :
`margin-left` of `dark-mode-toggle` : `1em` to `0.15em`
`font-size` of `nav` : `16px` to `15px`

 I've changed `margin-left` value of `dark-mode-toggle` to `0.15em` and `font-size` of anchor tag in `#footer nav` to `15px` so that the letter 'light' and 'dark' can be in the same line until `819px`.

The layout changes when the width is under `818px`.  So i think it would be better if there's no line break until `819px`.
So it works on my code!

### Before 
(over `860px`)
image
image

### After 
(over `860px`, it maintains the same line until `819px`)
image
image


Nitinkumar2905 (commented):
This change in font size really solved the line break on the theme change.

src/_css/main.css
```diff
@@ -224,6 +224,10 @@ hr {
   align-items: flex-end;
 }
 
+#footer nav {
+  font-size : 15px;
```
camillobruni, line 228: I'd prefer not adding a new font-size.
How about removing the dots `.` separators between the nav links and add `margin-right: 0.5em` for `footer nav a`?

wooseok123: I agree with your opinion.
I've just added the modified version. (removing dots and adding `margin-right` to them, and things i had to do to align them in the right position.)

The line-break still remains in the mobile version, but i think it's not a big things.

wooseok123:
@camillobruni If there aren’t any major issues, how about merging it?

src/_css/main.css
camillobruni: adding a .prettierrc sounds good to me

PR #696, opened 2023-04-21
Update embed.md to resolve samples crash

bbalaban:
On the compile line for hello_world.cc, you also need -DV8_ENABLE_SANDBOX, or else the sample will crash when run


camillobruni:
This has been fixed in PR #783

camillobruni:
This has been fixed in PR #783

PR #697, opened 2023-04-25
Add tags to feed entry `<id>`s

tomayac:
Over on web.dev, we plan to dynamically link to blog posts tagged with `WebAssembly` on your site. Therefore, I have annotated the feeds with the tags in the `<id>`, which is what the [RFC does](https://datatracker.ietf.org/doc/html/rfc4287#section-4.2.6:~:text=%3Cid%3Etag%3Aexample.org%2C2003%3A3%3C/id%3E). This will allow us to parse out the relevant posts  from your feed.

```xml
<entry>
    <title>Introducing the WebAssembly JavaScript Promise Integration API</title>
    <link href="https://v8.dev/blog/jspi"/>
    <updated>2023-01-19T00:00:00Z</updated>
    <!-- Here 👇 -->
    <id>webassembly:https://v8.dev/blog/jspi</id>
    <author>
      <name>Francis McCabe, Thibaud Michaud, Ilya Rezvov, Brendan Dahl</name>
    </author>
    <content type="html">[…]</content>
  </entry>
```

### DIFF
--- a/src/blog.atom.njk
+++ b/src/blog.atom.njk
@@ -19,7 +19,7 @@ excludeFromSitemap: true
     <title>{{ post.data.title }}</title>
     <link href="{{ absolutePostUrl }}"/>
     <updated>{{ (post.data.updated or post.date) | rssDate }}</updated>
-    <id>{{ absolutePostUrl }}</id>
+    <id>{% for tag in post.data.tags %}{% if not 'io' in tag and not 'Node.js' in tag %}{{ tag | slug }}:{% endif %}{% endfor %}{{ absolutePostUrl }}</id>
     <author>
       <name>{{ post.data.author | markdown | striptags }}</name>
     </author>
--- a/src/blog.atom.njk
+++ b/src/blog.atom.njk
@@ -19,7 +19,12 @@ excludeFromSitemap: true
     <title>{{ post.data.title }}</title>
     <link href="{{ absolutePostUrl }}"/>
     <updated>{{ (post.data.updated or post.date) | rssDate }}</updated>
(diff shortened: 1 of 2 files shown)


mathiasbynens:
We cannot change the `id`s without violating https://datatracker.ietf.org/doc/html/rfc4287#section-4.2.6:

> The "`atom:id`" element conveys a permanent, universally unique identifier for an entry or feed.

Is there another solution to this?

tomayac:
Yikes, you're right. Unfortunately there's no proper field for tags. Maybe if we only add the tags for future posts? I can do a manual dump once, and the automatism would kick in from there on? Or maybe you have another idea?

mathiasbynens:
> Is there another solution to this?

E.g. can we use `<category>`?

tomayac:
Oh, I must have overlooked `<category>`. This is the right thing. I was looking for "tag", and found nothing relevant. I'll rework this so it uses `<category>`.

tomayac:
PTAL.

src/blog.atom.njk
```diff
@@ -20,6 +20,11 @@ excludeFromSitemap: true
     <link href="{{ absolutePostUrl }}"/>
     <updated>{{ (post.data.updated or post.date) | rssDate }}</updated>
     <id>{{ absolutePostUrl }}</id>
+    {%- for tag in post.data.tags %}
+      {%- if not 'io' in tag and not 'Node.js' in tag %}
+        <category term="{{ tag | slug }}" />
```
mathiasbynens, line 25: WDYT about adding `label="…"` with the human-readable version of the tag? Here and in the other file.

tomayac: I like it. Done in https://github.com/v8/v8.dev/pull/697/commits/55f6a42d4744ca3e83816b4db1c221320b734bf0.

mathiasbynens (approved):

PR #698, opened 2023-04-26
Stack Trace API: Remove CallSite.IsPromiseAny which is not exposed

Liedtke:
This method is not registered on the JS `CallSite` object.


syg:
Did you mean for this to be marked as Draft?

Liedtke:
This would've been the alternative to ship it: https://chromium-review.googlesource.com/c/v8/v8/+/4479733
Given this doesn't seem to be missed by anyone, it was decided to instead remove it from the documentation.

PR #699, opened 2023-05-01
Update cross-compile-ios.md to remove bitcode arg

tiagovignatti:
According to [Xcode 14 documentation](https://developer.apple.com/documentation/Xcode-Release-Notes/xcode-14-release-notes#Deprecations): 

> Xcode no longer builds bitcode by default and generates a warning message if a project explicitly enables bitcode: “Building with bitcode is deprecated. Please update your project and/or target settings to disable bitcode.” The capability to build with bitcode will be removed in a future Xcode release.

This has been removed for other projects e.g. Chromium a while a go https://bugs.chromium.org/p/chromium/issues/detail?id=1335872

### DIFF
--- a/src/docs/cross-compile-ios.md
+++ b/src/docs/cross-compile-ios.md
@@ -31,7 +31,6 @@ This section shows how to build a monolithic V8 version for use on either a phys
 Set up GN build files by running `gn args out/release-ios` and inserting the following keys:
 
 ```python
-enable_ios_bitcode = true
 ios_deployment_target = 10
 is_component_build = false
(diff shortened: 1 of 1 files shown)


backes (approved):

tiagovignatti:
thanks again @backes. I don't have write access so I can't merge. Maybe you can take a look at that as well? :)

backes:
Yes, sure, was waiting for all checks to pass and then forgot. I would probably have found the tab later again :) 
(thanks for the ping!)

PR #700, opened 2023-05-05
Update cross-compile-ios.md to remove use_xcode_clang arg

tiagovignatti:
`use_xcode_clang` arg is not used anymore. Please refer to https://bugs.chromium.org/p/chromium/issues/detail?id=1266466. This is even generating a warning when setting up GN build files:

```
WARNING at build arg file (use "gn args <out_dir>" to edit):6:19: Build argument has no effect.
use_xcode_clang = true
                  ^---
The variable "use_xcode_clang" was set as a build argument
but never appeared in a declare_args() block in any buildfile.
```


tiagovignatti:
@backes could PTAL? Thanks again.

backes (approved):

PR #702, opened 2023-06-21
Speeding up v8 heap snapshots

jdapena:
* Add "Speeding up V8 heap snapshots" blog post.


robpalme:
My email address has now been registered as CLA compliant via my employer. How do we make the CLA bot reassess?

jdapena:
CLA is ready now.

I cannot approve running the workflow (nor assign anybody to review the patch).

robpalme:
@syg are you able to run the workflow?

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+After fixing the hashing, we re-profiled and found a further optimization opportunity that would take a significant part of the time. *An old friend*.
+
+For each allocation in the V8 heap, the heap snapshot tries to record the call stack responsible for the allocation. Therefore, for each stack frame, it needs to know the function name and its source location (filename, line number, column number). It was the fetching of this information that turned out to be super slow!
+
+What was happening? It was something similar to [what I fixed in the ETW stack walk](https://chromium-review.googlesource.com/c/v8/v8/+/3669698) and that I explained in [this post](https://blogs.igalia.com/dape/2022/12/21/native-call-stack-profiling-3-3-2022-work-in-v8/).
+
+To compute the source code lines of a function, V8 knows the linear position of the function in the file. But, to find the source line, it needs to traverse the whole file to identify where each newline occurs. This is expensive.
+
+When requesting line information, V8 already knows how to cache the source line positions per-script. But the heap snapshot implementation was not using a cache!
```
joyeecheung, line 128: IIUC this is talking about `Script::InitLineEnds`. V8 cache this information when it knows that the script is being compiled in an active profiling session, and most importantly when it's allowed to store the line ends in the script (when heap allocation is allowed). Since heap snapshot can be generated anytime without alerting V8 in any way in advance, V8 could likely need this information in the middle of the heap explorer run, when it cannot save this information in the scripts since heap allocation is not allowed at that point. As a result V8 needs to recompute this information over and over again but cannot save it due to the unfortunate timing. The fix is to compute and cache all the line ends in advance when the heap allocation is still allowed.

It would be nicer to explain why this wasn't done before, and why the timing matters.

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+[This source position caching fix](https://chromium-review.googlesource.com/c/v8/v8/+/4538766) has also landed in V8.
+
+## Did we make it fast?
+
+After enabling both fixes, we re-profiled. Both of our fixes only affect snapshot generation time, so, as expected, snapshot serialization times were unaffected.
+
+When operating on a JS program containing...
+
+- Unoptimized development JS, generation time is **50% faster** 👍
```
joyeecheung, line 140: Maybe `unprocessed`/`processed` would be less ambiguous here? I kept thinking about V8 optimizations when I read `optimized` in this context :)

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+The optimizations were validated on both Windows and Linux target environments.
+
+For the particularly challenging problem originally faced by the Bloomberg engineers, the total end-to-end time to capture a 100MB snapshot was reduced from a painful 10 minutes down to a very pleasant 6 seconds. That is **a 100x win!** 🔥
+
+The optimizations are generic wins that we expect to be widely applicable to anyone performing memory debugging on V8, Node, and Chromium. These wins shipped in V8 11.5.130, which means they are found in Chromium 115.0.5576.0. We look forward to Node gaining these optimizations once it adopts the newer V8 release.
+
+## What’s next?
+
+First, it would be useful for Node.js to accept the new `--profile-heap-snapshot` flag. Today, it filters V8 command line options, and only allows a known subset.
```
joyeecheung, line 153: This is still talking about `NODE_OPTIONS` right? I think it would make sense to explain how `NODE_OPTIONS` comes into the picture in this case:

```suggestion
First, it would be useful for Node.js to accept the new `--profile-heap-snapshot` flag in `NODE_OPTIONS`. In some use cases, users cannot control the the command line options passed to Node.js directly and have to configure them through the environment variable `NODE_OPTIONS`. Today, Node.js filters V8 command line options set in the environment variable, and only allows a known subset, which could make it harder to test new V8 flags in Node.js, as what happened in our case.
```

Though my personal comment is that something like https://github.com/nodejs/node/issues/46339 that might make a bit more sense than allowing another ad-hoc V8 flag in NODE_OPTIONS. e.g. having a `NODE_OPTIONS_V8_ALL`.

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+ - memory
+ - tools
+---
+*This blog post has been authored by José Dapena Paz (Igalia), with contributions from Jason Williams (Bloomberg), Ashley Claymore (Bloomberg), and Rob Palmer (Bloomberg).*
+
+In this post about V8 heap snapshots, I will talk about some performance problems found by Bloomberg engineers, and how we fixed them to make JavaScript memory analysis faster than ever.
+
+## The Problem
+
+Ashley Claymore was working on diagnosing a memory leak in a JavaScript application. It was failing with *Out-Of-Memory* errors. Despite the process having access to plenty of system memory, V8 places a hard limit on the amount of memory dedicated to the garbage-collected heap from which all JavaScript objects are allocated. This V8 heap limit (~1400MB) was being hit.
```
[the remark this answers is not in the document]
jdapena: I have been trying to find out where those 1400MB could be coming from... I guess that was because of some heuristic depending on available system memory in the target device that was crashing? I will check with Ashley.

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+The standard way to debug a routine memory leak scenario like this is to capture a heap snapshot and then inspect the various summaries and object attributes using DevTools "Memory" tab to find out what is consuming the most memory.  In DevTools, you click the round button marked _"Take heap snapshot"_ to perform the capture. For Node.js applications, you can [trigger the snapshot](https://nodejs.org/en/docs/guides/diagnostics/memory/using-heap-snapshot) programmatically using this API:
+
+```javascript
+require('v8').writeHeapSnapshot();
+```
+
+The desire was to capture several snapshots at different points in the application's life, so that DevTools Memory viewer could be used to show the difference between each.  The problem was that capturing a single full-size (500MB) snapshot was taking **over 30 minutes** alone!
```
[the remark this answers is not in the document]
jdapena: Going to rewrite that as "the difference between the heaps at different times". Thanks!

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+`--heapsnapshot-near-heap-limit=10`
+: This tells V8 to produce a maximum of 10 snapshots, each time it comes close to running out of memory.  This prevents thrashing where the memory-starved program spends a long time producing more snapshots than needed.
+
+`--enable-etw-stack-walking`
+: This allows tools such as ETW, WPA & xperf to see the JS stack which has been called in V8. (Node v20+)
+
+`--interpreted-frames-native-stack`
+: This flag is used in combination with tools like ETW, WPA & xperf to see the native stack when profiling. (Node v20+).
+
+After each snapshot, V8 tries to force garbage collection to reduce memory usage and avoid hitting the limit. In the test case, the memory usage increases, but, after several iterations, garbage collection ultimately could not free up enough space and so the application terminated with an *Out-Of-Memory* error.
```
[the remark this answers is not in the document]
jdapena: This is interesting: the near limit heap snapshot callback is invoked in several scenarios. Most of them are exactly happening on different garbage collection stages (the exception is when it has an ineffective MarkCompact, that is usually pretty bad.

So, I think we need to rewrite this paragraph to be more accurate. Something like...

> When V8 is close to the heap limit, it forces a garbage collection to reduce the memory usage and avoid hitting the limit. With the provided Node.js `--heapsnapshot-near-heap-limit` flag it will also dump a new heap snapshot. In the test case, the memory usage increases, but, after several iterations, garbage collection ultimately can not free up enough space and so the application is terminated with an *Out-Of-Memory* error.

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: In this case, I wanted to avoid going too much deep on the implementation details, and write more about the full process of iterating with profiling tools to solve the problem.

I agree on the interest of reviewing all this, to further explore improvements in the heap snapshot capture process. I.e. with the new changes, we actually allocate heap memory for line ends.

I think we should eventually move back to not allocating new line end information in heap snapshot capture (likely allocating a temporary line ends memory structure out of V8 heap that can be reused while heap snapshot is running, and is used only when line end information is not already available in the heap).

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: Yes, it is a bit confusing (though it is clarified in next paragraph after this list). Processed/unprocessed looks to me quite ambiguous, and not really improving the readability.

Maybe just removing unoptimized/optimized words could be OK, as we are already talking about development/production JS.

I think I will keep it as is (unless some other reviewer provides an alternate writing that provides a clearly less ambiguous writing?).

src/blog/speeding-up-v8-heap-snapshots.md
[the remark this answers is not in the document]
joyeecheung: LGTM, except

> With the provided Node.js --heapsnapshot-near-heap-limit flag it will also dump a new heap snapshot. 

could be tweaked to

> It also notifies the embedder that the heap is about to reach the memory limit. The `--heapsnapshot-near-heap-limit` flag in Node.js dumps a new heap snapshot upon notification.

Otherwise it reads as if V8 is the one initiating the heap snapshot generation.

(Also I think you probably meant `the memory usage decreases, but..` :)).

src/blog/speeding-up-v8-heap-snapshots.md
joyeecheung: "development/production JS" seems unambiguous enough to me 👍

src/blog/speeding-up-v8-heap-snapshots.md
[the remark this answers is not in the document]
jdapena: Using all these comments. Thanks!

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: I have been rereading this... and definitely there is a big difference between the detail for the first fix and the second fix. It even looks as I was in a hurry explaining the second one!

I will try to provide a more accurate description of what is fixed, mostly inspired in your first comment in this thread.

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: Yup, I will change writing to make that less ambiguous, providing a better explanation of how production source code is optimized.

src/blog/speeding-up-v8-heap-snapshots.md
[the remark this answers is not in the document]
jdapena: Rewritten the paragraph with information collected from Ashley and Joyee. Thanks!

jdapena:
@syg < Apparently I cannot add you or any other reviewer to the ticket. Who should review blog posts?

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+Some examples of problems with this hash function:
+
+- Once we had strings with a hash key value in lower positions, then the storing of new numbers would offset all the positions.
+- Or even worse: if we would introduce a string with a hash key value that would be a low number, and we already found several hundreds or thousands of consecutive numbers, then that value would be moved several hundreds or thousands of positions.
+
+What did I do to fix it? As the problem comes mostly from numbers represented as strings that would fall in consecutive positions, I modified the hash function so we would rotate the resulting hash value 2 positions to the left. So, for each number, we would introduce 3 free positions. Why 2? Empirical testing across several work-sets showed this number was the best choice to minimize collisions.
```
joyeecheung, line 114:
```suggestion
What did I do to fix it? As the problem comes mostly from numbers represented as strings that would fall in consecutive positions, I modified the hash function so we would rotate the resulting hash value 2 positions to the left. So each pair of consecutive number would be placed with 3 free positions in between. Why 2? Empirical testing across several work-sets showed this number was the best choice to minimize collisions.
```

(I took me a while to realize that it was talking about `((n + 1) << 2) - (n << 2) = 3`)

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+[This hashing fix](https://chromium-review.googlesource.com/c/v8/v8/+/4428811) has landed in V8.
+
+## Second Optimization: Caching source positions
+
+After fixing the hashing, we re-profiled and found a further optimization opportunity that would take a significant part of the time. *An old friend*.
```
joyeecheung, line 120:
```suggestion
After fixing the hashing, we re-profiled and found a further optimization opportunity that would reduce a significant part of the overhead.
```

It seems the "*An old friend*" part is too further away from the explanation about the ETW stack walk patch and leaves the reader wondering if there is some text or a link missing. I think it can be either removed because the subsequent mention was good enough, or the subsequent mention could be moved here to avoid leaving the readers wondering if there's something missing.

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+What was happening? It was something similar to [what I fixed in the ETW stack walk](https://chromium-review.googlesource.com/c/v8/v8/+/3669698) and that I explained in [this post](https://blogs.igalia.com/dape/2022/12/21/native-call-stack-profiling-3-3-2022-work-in-v8/).
+
+To compute the source code lines of a function, V8 knows the linear position of the function in the script. But, to find the source line, it needs to traverse the whole script to identify where each newline occurs. This is expensive.
+
+When requesting line information, V8 already implements caching the source line positions per-script in method `Script::InitLineEnds()`. That information is stored in each script using a new heap object. Unfortunately, that was a problem that prevented the heap snapshot from using the cache. This is because, when the snapshot implementation traverses the heap, it cannot modify it, so the newly calculated line information cannot be cached.
```
joyeecheung, line 128:
```suggestion
When requesting line information, V8 already implements caching of the source line positions per-script in method `Script::InitLineEnds()`. That information is stored in each script using a new heap object. Unfortunately, the snapshot implementation cannot modify the heap when traversing it, so the newly calculated line information cannot be cached.
```

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+The solution? Before generating the heap snapshot, we now iterate over all the scripts in the V8 context to compute and cache the new source line positions information. As this is done out of the heap traversal done while capturing the snapshot, it is still possible to modify the heap to store the cached source code lines information.
```
joyeecheung, line 130:
```suggestion
The solution? Before generating the heap snapshot, we now iterate over all the scripts in the V8 context to compute and cache the source line positions. As this is not done when we traverse the heap for heap snapshot generation, it is still possible to modify the heap and store the source line positions as a cache.
```

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+## Did we make it fast?
+
+After enabling both fixes, we re-profiled. Both of our fixes only affect snapshot generation time, so, as expected, snapshot serialization times were unaffected.
+
+When operating on a JS program containing...
+
+- Development JS, generation time is **50% faster** 👍
+- Production JS, generation time is **90% faster** 😮
+
+Why such a massive difference between production and development code? It is mostly because of the production source code optimizations. In production, code is optimized using bundling and minification, so there are fewer JS files and these files tend to be large, which means finding the newlines takes longer. And therefore, these larger files benefit most from the source position caching that only performs this work once.
```
joyeecheung, line 143: Some grammar and clarity fixes:

```suggestion
Why was there a massive difference between production and development code? The prodcution code is optimized using bundling and minification, so there are fewer JS files, and these files tend to be large. It takes longer to calculate source line positions for these large files, so they benefit the most when we can cache the source position and avoid repeated calculations.
```

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+The optimizations were validated on both Windows and Linux target environments.
+
+For the particularly challenging problem originally faced by the Bloomberg engineers, the total end-to-end time to capture a 100MB snapshot was reduced from a painful 10 minutes down to a very pleasant 6 seconds. That is **a 100x win!** 🔥
+
+The optimizations are generic wins that we expect to be widely applicable to anyone performing memory debugging on V8, Node, and Chromium. These wins shipped in V8 11.5.130, which means they are found in Chromium 115.0.5576.0. We look forward to Node gaining these optimizations once it adopts the newer V8 release.
```
joyeecheung, line 149:
```suggestion
The optimizations are generic wins that we expect to be widely applicable to anyone performing memory debugging on V8, Node.js, and Chromium. These wins were shipped in V8 11.5.130, which means they are found in Chromium 115.0.5576.0. We look forward to Node.js gaining these optimizations in the next semver-major release.
```

(Node.js usually does full V8 upgrades in semver-major releases, e.g. https://github.com/nodejs/node/pull/48710 for V8 11.6 is probably going out for Node.js v21)

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+require('v8').writeHeapSnapshot();
+```
+
+The desire was to capture several snapshots at different points in the application's life, so that DevTools Memory viewer could be used to show the difference between the heaps at different times. The problem was that capturing a single full-size (500MB) snapshot was taking **over 30 minutes** alone!
+
+It was this slowness in the memory analysis workflow that we needed to solve.
+
+## Narrowing down the Problem
+
+Jason Williams started investigating the issue using some V8 parameters. As described in the previous post, V8 has some nice command line parameters that can help with that. These options were used to curate the heap snapshots , simplify the reproduction, and improve observability:
```
syg, line 30:
```suggestion
Jason Williams started investigating the issue using some V8 parameters. These options were used to create the heap snapshots, simplify the reproduction, and improve observability:
```

syg: Other than --max-old-space-size, all the flags below are Node-specific. What do you recommend folks do that are trying to debug memory leaks not on Node?

Also what does "improve observability" mean?

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+`--max-old-space-size=100`
+: This limits the heap to 100 megabytes and helps to reproduce the issue much faster.
+
+`--heapsnapshot-near-heap-limit=10`
+: This is a Node.js specific command line parameter that tells V8 to produce a maximum of 10 snapshots, each time it comes close to running out of memory.  This prevents thrashing where the memory-starved program spends a long time producing more snapshots than needed.
+
+`--enable-etw-stack-walking`
+: This allows tools such as ETW, WPA & xperf to see the JS stack which has been called in V8. (Node v20+)
```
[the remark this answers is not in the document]
syg: Also Node-specific, yes?

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+`--interpreted-frames-native-stack`
+: This flag is used in combination with tools like ETW, WPA & xperf to see the native stack when profiling. (Node v20+).
```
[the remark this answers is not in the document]
syg: Also Node-specific, yes?

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+When V8 is close to the heap limit, it forces a garbage collection to reduce the memory usage and avoid hitting the limit. It also notifies the embedder that the heap is about to reach the memory limit. The `--heapsnapshot-near-heap-limit` flag in Node.js dumps a new heap snapshot upon notification. In the test case, the memory usage decreases, but, after several iterations, garbage collection ultimately can not free up enough space and so the application is terminated with an *Out-Of-Memory* error.
+
+Jason took recordings using Windows Performance Analyzer (see below) in order to narrow down the issue. This revealed that most CPU time was being spent within the V8 Heap Explorer. Specifically, it took around 30 minutes just to walk through the heap to visit each node and collect the name. This didn’t seem to make much sense - why would recording the name of each property take so long?
+
+This is when I was asked to take a look.
```
syg, line 48: Again I'd just stick with "Igalian engineers" or "Bloomberg engineers".

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+We also identified other widely-used JavaScript applications that demonstrated the same slowdown, in particular, running ESLint on TypeScript. The problem was not app-specific.
+
+Furthermore, we found the problem happened on both Windows and Linux. The problem was also not platform-specific.
+
+## Windows Performance Analyzer to the rescue
+
+As the problem was initially reported on a Windows platform, I used [Windows Performance Toolkit](https://learn.microsoft.com/en-us/windows-hardware/test/wpt/), based on [ETW](https://learn.microsoft.com/en-us/windows-hardware/drivers/devtest/event-tracing-for-windows--etw-), for analysis. This is a powerful low-level expert tool to find out exactly what a program is doing on Windows.
+
+To record the session, I followed these steps:
```
syg, line 68: This section seems overly detailed to me. The upshot seems to be that there's a Windows-specific profiler that gave you good leads, which I think suffices. The meat of this is what you found out from those leads and how you optimized the code.

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+One third of the samples was spent in `v8::internal::StringsStorage::GetEntry`:
+
+```javascript
+181 base::HashMap::Entry* StringsStorage::GetEntry(const char* str, int len) {
+182   uint32_t hash = ComputeStringHash(str, len);
+183   return names_.LookupOrInsert(const_cast<char*>(str), hash);
+184 }
+```
+
+Where exactly? I added the source line numbers to the break down and found this:
```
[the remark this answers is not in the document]
syg: +1

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+![](/_img/speeding-up-v8-heap-snapshots/59293b17-6d52-4d9e-8737-5b23d038d50a.png){.no-darkening}
+
+So, that inlined `ComputeStringHash` call is causing over 30% of the delay! But why?
+
+Let’s first talk about `StringsStorage`. Its purpose is to store a unique copy of all the strings that will be used in the heap snapshot. For fast access and handling uniqueness, this class uses a flatmap: a hashmap backed by an array, where collisions are handled by storing elements in the next position of the array.
```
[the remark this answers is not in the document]
syg: Big +1 here, uniqueness is a strange way to put deduplication.

syg, line 101: I don't think that's a well-understood use of "flatmap", especially in a JS context where [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) exists. I'd avoid giving it a name and just describe the representation.

src/blog/speeding-up-v8-heap-snapshots.md
syg: Could you perhaps add code snippets demonstrating the improved hash directly? V8 blog audience should be used to bitwise manipulation code.

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+## What’s next?
+
+First, it would be useful for Node.js to accept the new `--profile-heap-snapshot` flag in `NODE_OPTIONS`. In some use cases, users cannot control the the command line options passed to Node.js directly and have to configure them through the environment variable `NODE_OPTIONS`. Today, Node.js filters V8 command line options set in the environment variable, and only allows a known subset, which could make it harder to test new V8 flags in Node.js, as what happened in our case.
+
+Information accuracy in snapshots can be improved further. Today, each script source code line information is stored in a representation in the V8 heap itself. And that’s a problem because we want to measure the heap precisely without the observability infrastructure affecting the subject we are observing. Ideally, we would store the cache of line information outside the V8 heap in order to make heap snapshot information more accurate.
```
syg, line 155: What is "observability infrastructure"? Do you mean performance measurement?

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+Finally, now that we improved the generation phase, the biggest cost is now the serialization phase. Further analysis may reveal new optimization opportunities in serialization.
+
+## The end
+
+This work was possible thanks to the sponsorship of [Igalia](https://www.igalia.com) and [Bloomberg](https://techatbloomberg.com).
```
syg, line 161: "Sponsorship" is not the right word here, which makes it sound like the V8 team is being sponsored. Please phrase it as who did this work instead.

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+## What’s next?
+
+First, it would be useful for Node.js to accept the new `--profile-heap-snapshot` flag in `NODE_OPTIONS`. In some use cases, users cannot control the the command line options passed to Node.js directly and have to configure them through the environment variable `NODE_OPTIONS`. Today, Node.js filters V8 command line options set in the environment variable, and only allows a known subset, which could make it harder to test new V8 flags in Node.js, as what happened in our case.
+
+Information accuracy in snapshots can be improved further. Today, each script source code line information is stored in a representation in the V8 heap itself. And that’s a problem because we want to measure the heap precisely without the observability infrastructure affecting the subject we are observing. Ideally, we would store the cache of line information outside the V8 heap in order to make heap snapshot information more accurate.
+
+Finally, now that we improved the generation phase, the biggest cost is now the serialization phase. Further analysis may reveal new optimization opportunities in serialization.
+
+## The end
```
syg, line 159:
```suggestion
## Credits
```

src/blog/speeding-up-v8-heap-snapshots.md
syg, line 30: Re-reading this section after reading the entire post, I'm unclear on the point of listing the flags in detail. It seems almost incidental: here are the flags that we were using to capture heap snapshots, and where we observed surprising slowdowns. Could this section be shortened? Perhaps I'm missing the intention though?

src/_img/speeding-up-v8-heap-snapshots/59293b17-6d52-4d9e-8737-5b23d038d50a.png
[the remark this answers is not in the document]
jdapena: Sure, I will crop them to improve readability.

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: Yup, the old friend reference can be dropped because I talk later about the similar issue fixed in ETW. Thanks!

src/blog/speeding-up-v8-heap-snapshots.md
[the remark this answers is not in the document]
jdapena: No. The only node-specific flag is `--heapsnapshot-near-heap-limit`. The other ones are also available in V8 (Chromium and d8).

src/blog/speeding-up-v8-heap-snapshots.md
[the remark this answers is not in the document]
jdapena: Ditto.

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+
+Jason took recordings using Windows Performance Analyzer (see below) in order to narrow down the issue. This revealed that most CPU time was being spent within the V8 Heap Explorer. Specifically, it took around 30 minutes just to walk through the heap to visit each node and collect the name. This didn’t seem to make much sense - why would recording the name of each property take so long?
+
+This is when I was asked to take a look.
+
+## Quantifying the Problem
+
+We knew the snapshots were dramatically increasing execution time, so the first step was adding support in V8 to better understand the time used by heap snapshotting. The capture process itself is split into two phases: generation, then serialization.
```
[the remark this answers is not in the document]
jdapena: I agree. It is kind of the second part of the same reasoning. I will put them together in the same paragraph.

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: Applying after fixing the typo with prodctuion -> production

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,161 @@
...
+ - memory
+ - tools
+---
+*This blog post has been authored by José Dapena Paz (Igalia), with contributions from Jason Williams (Bloomberg), Ashley Claymore (Bloomberg), Rob Palmer (Bloomberg) and Joyee Cheung(Igalia).*
+
+In this post about V8 heap snapshots, I will talk about some performance problems found by Bloomberg engineers, and how we fixed them to make JavaScript memory analysis faster than ever.
+
+## The Problem
+
+Ashley Claymore was working on diagnosing a memory leak in a JavaScript application. It was failing with *Out-Of-Memory* errors. Despite the process having access to plenty of system memory, V8 places a hard limit on the amount of memory dedicated to the garbage-collected heap from which all JavaScript objects are allocated. For the tested application, this limit was configured to be around 1400MB. Normally V8's garbage collector should be able to keep the heap usage under that limit, so the failures indicated that there was a leak.
```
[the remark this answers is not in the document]
jdapena: Applying manually.

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: Yes, I will avoid explicit names mention in the contents of the post body.

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: In this case I wanted to be explicit about the procedure used for the investigation, so other people can benefit from it investigating performance problems in V8 (and not only V8) scope.

Even when Windows Performance Toolkit documentation is not bad, the specific steps to investigate this V8 problem are not completely trivial. The actual solutions to the detected problems were not really complex, but identifying exactly where the overhead was has been more significant. And for that, the step-by-step looks valuable to me.

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: I'll replace "observability infrastructure" with "performance measurement", but add the word "overhead".

It is the performance measurement overhead that affects the subject we are observing.

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: Rewriting as:
"This was possible thanks to the work of [Igalia](https://www.igalia.com) and [Bloomberg](https://techatbloomberg.com) engineers."

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: Removed reference to the specific engineers names.

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: Initially the problem was detected just extracting regular heap snapshots from DevTools.

But using these command line arguments allowed to get a detailed breakdown of what was happening, and also allowed to reproduce the heap snapshot performance problem several times, without requiring to use a remote DevTools connection.

So this is part of the investigation steps (increasing and simplifying reproducibility).

In general, the intent of the whole post is not only explaining the specific fixes, but the whole investigation procedure that led to them.

src/blog/speeding-up-v8-heap-snapshots.md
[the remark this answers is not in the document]
jdapena: Done.

jdapena:
Provided a new version of the explanation of the hash algorithm changes, with value examples, and snippets of the original and the new hash functions.

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Second Optimization: Caching source positions
+
+After fixing the hashing, we re-profiled and found a further optimization opportunity that would reduce a significant part of the overhead.
+
+For each allocation in the V8 heap, the heap snapshot tries to record the call stack responsible for the allocation. Therefore, for each stack frame, it needs to know the function name and its source location (filename, line number, column number). It was the fetching of this information that turned out to be super slow!
+
+What was happening? It was something similar to [what I fixed in the ETW stack walk](https://chromium-review.googlesource.com/c/v8/v8/+/3669698) and that I explained in [this post](https://blogs.igalia.com/dape/2022/12/21/native-call-stack-profiling-3-3-2022-work-in-v8/).
+
+To compute the source code lines of a function, V8 knows the linear position of the function in the script. But, to find the source line, it needs to traverse the whole script to identify where each newline occurs. This is expensive.
```
joyeecheung, line 155: It's a bit difficult for me to understand what this is explaining, even though I already kind of know how the code works, I think mentioning the line and column numbers here would help:

```suggestion
V8 usually only stores where the the function starts and ends in the form of linear offsets in the script. To calculate line and column numbers for allocation sites within a function, it needs to traverse the whole script and record where the line breaks are.
```

And maybe this can be merged into the `For each allocation in the V8 heap..` paragraph, so that we don't have to repeat "this turns out to be slow" and "this is expensive".

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,190 @@
...
+
+So, for each number, we would introduce 3 free positions. Why rotating 2 bits? Empirical testing across several work-sets showed this number was the best choice to minimize collisions.
+
+[This hashing fix](https://chromium-review.googlesource.com/c/v8/v8/+/4428811) has landed in V8.
+
+## Second Optimization: Caching source positions
+
+After fixing the hashing, we re-profiled and found a further optimization opportunity that would reduce a significant part of the overhead.
+
+For each allocation in the V8 heap, the heap snapshot tries to record the call stack responsible for the allocation. Therefore, for each stack frame, it needs to know the function name and its source location (filename, line number, column number). It was the fetching of this information that turned out to be super slow!
```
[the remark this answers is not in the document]
joyeecheung: Actually looking at the snapshot generator code again I don't think the line ends are recorded for this purpose - this is describing allocation profile where the allocation site matters. For generating a full snapshot, V8 only calculates the line/column numbers for the DevTools to display the positions of certain retainers (most importantly closures) with links to the source code: https://bugs.chromium.org/p/chromium/issues/detail?id=854097 and that's why all it needs is just the *start* position of the function. So perhaps this is more accurate (I also merged suggestions from below to make the paragraph look more logically cohesive):

```suggestion
When generating a heaps snapshot, for each function in the heap, V8 tries to record its start position in a pair of line and column numbers. This information can be used by the DevTools to display a link to the source code of the function. During usual compilation, however, V8 only stores the start position of each function in the form of a linear offset from the beginning of the script. To calculate the line and column numbers based on the linear offset, V8 needs to traverse the whole script and record where the line breaks are. This calculation turns out to be very expensive.
```

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,190 @@
...
+require('v8').writeHeapSnapshot();
+```
+
+They wanted to capture several snapshots at different points in the application's life, so that DevTools Memory viewer could be used to show the difference between the heaps at different times. The problem was that capturing a single full-size (500MB) snapshot was taking **over 30 minutes** alone!
+
+It was this slowness in the memory analysis workflow that we needed to solve.
+
+## Narrowing the Problem
+
+Then, Bloomberg engineers started investigating the issue using some V8 parameters. As described in the previous post, V8 has some nice command line parameters that can help with that. These options were used to create the heap snapshots, simplify the reproduction, and improve observability:
```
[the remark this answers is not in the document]
jdapena: The post was originally the second part of another blog post... that is not even mentioned at all. I will replace that with "this post" that would be a link to the initial post.

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: I agree. It is quite redundant. I will merge them.

syg (approved):
Looks pretty good % remaining comments! Thanks for the many rounds of iterations.

jdapena:
Applied changes. Only pending running the workflow and merging.

I'd prefer to keep the Windows Performance Analyzer section, that is now really small, as it still give readers hint about the tool used for the investigation.

syg:
> I'd prefer to keep the Windows Performance Analyzer section, that is now really small, as it still give readers hint about the tool used for the investigation.

I still think something as short as "I used Windows Performance Toolkit to do the profiling" suffices. I feel like that workflow (1) isn't generally applicable and (2) out of character for the V8 blog. V8 blog posts tend to be about highlighting improvements to the infrastructure, or VM guts, or language hacks, not play-by-play accounts of how we achieved the results.

jdapena:
@syg < I finally removed the WPA section, and just made the reference that it was the tool used (that is visible in the screenshots of the reports). I cannot apply the resulting patch myself now it is approved.

BTW, thanks everybody for all the careful review. Post is now far better!

src/blog/speeding-up-v8-heap-snapshots.md
```diff
@@ -0,0 +1,179 @@
...
+
+Normally, after V8 finishes calculating the offsets of line breaks in a script, it caches them in a newly allocated array attached to the script. Unfortunately, the snapshot implementation cannot modify the heap when traversing it, so the newly calculated line information cannot be cached.
+
+The solution? Before generating the heap snapshot, we now iterate over all the scripts in the V8 context to compute and cache the offsets of the line breaks. As this is not done when we traverse the heap for heap snapshot generation, it is still possible to modify the heap and store the source line positions as a cache.
```
joyeecheung, line 148:
```suggestion
The solution? Before generating the heap snapshot, we now iterate over all the scripts in the V8 context to compute and cache the offsets of the line breaks. As this is done when we traverse the heap for heap snapshot generation, it is still possible to modify the heap and store the source line positions as a cache.
```

joyeecheung (approved):

src/blog/speeding-up-v8-heap-snapshots.md
jdapena: The original sentence was OK. When we traverse the heap for heap snapshot generation we cannot add entries to the heap. That's the reason we do that before.

joyeecheung: Ah, yes, I think I misread 👍

syg (approved):
lgtm, thanks for the many rounds

syg:
Any last concerns before I publish? @joyeecheung?

joyeecheung (approved):
LGTM, thanks

PR #721, opened 2023-12-07
Update Node to 20

victorgomes:
Firebase requires a newer Node. See https://github.com/v8/v8.dev/actions/runs/7125569363/job/19401750565.


lutzvahl (approved):
LGTM

PR #724, opened 2023-12-08
s/single static assignment/static single assignment/

zmodem:


LeszekSwirski:
Oops, you're right, I missed this. Thanks for the fix.

LeszekSwirski (approved):

zmodem:
Thanks! I don't think I have permission to merge, so someone else will have to push the button for me.

PR #729, opened 2024-01-30
Add import attributes feature post

syg:


syg:
@rmahdav PTAL. We should publish this ASAP so we have something to point to for the deprecation of the `assert` syntax.

@dtig for lgtm stamp since @rmahdav isn't on the V8 Core GH team yet.

syg:
Also @nicolo-ribaudo PTAL if you have some time.

dtig (approved):

src/features/import-attributes.md
```diff
@@ -0,0 +1,63 @@
...
+date: 2024-01-30
+tags:
+  - ECMAScript
+description: 'Import attributes: the evolution of import assertions'
+tweet: ''
+---
+
+## Previously
+
+V8 shipped the [import assertions](https://chromestatus.com/feature/5765269513306112) feature in v9.1. This feature allowed module import statements to include additional information by using the `assert` keyword. This additional information is currently used import JSON and CSS modules inside JavaScript modules.
```
jridgewell, line 15:
```suggestion
V8 shipped the [import assertions](https://chromestatus.com/feature/5765269513306112) feature in v9.1. This feature allowed module import statements to include additional information by using the `assert` keyword. This additional information is currently used to import JSON and CSS modules inside JavaScript modules.
```

src/features/import-attributes.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Import attributes
+
+Since then, import assertions has evolved into [import attributes](https://github.com/tc39/proposal-import-attributes). The point of the feature remains the same: to allow module import statements to include additional information.
+
+The most important difference is that import assertions had assert-only semantics, while import attributes has more relaxed semantics. Assert-only semantics means that the additional information as part no effect on _how_ a module is loaded, only on _whether_ it is loaded. For example, a JSON module is always loaded as JSON module by virtue of its MIME type, and the `assert { type: 'json' }` clause can cause loading to fail if the requested module's MIME type is not `application/json`.
```
jridgewell, line 21:
```suggestion
The most important difference is that import assertions had assert-only semantics, while import attributes has more relaxed semantics. Assert-only semantics means that the additional information has no effect on _how_ a module is loaded, only on _whether_ it is loaded. For example, a JSON module is always loaded as JSON module by virtue of its MIME type, and the `assert { type: 'json' }` clause can only cause loading to fail if the requested module's MIME type is not `application/json`.
```

jridgewell (approved):

rmahdav:
@syg Thanks for the post. It looks good to me.

src/features/import-attributes.md
nicolo-ribaudo, line 21:
```suggestion
The most important difference is that import assertions had assert-only semantics, while import attributes has more relaxed semantics. Assert-only semantics means that the additional information as part no effect on _how_ a module is loaded, only on _whether_ it can be executed. For example, a JSON module is always loaded as JSON module by virtue of its MIME type, and the `assert { type: 'json' }` clause can cause loading to fail if the requested module's MIME type is not `application/json`.
```

Nit, but with assertions even when the module was loaded before checking the assertion, and the assertion only prevented it from being actually "executed" (if we can call execution the conversion of JSON/CSS files into JS values)

src/features/import-attributes.md
```diff
@@ -0,0 +1,63 @@
...
+
+However, assert-only semantics had a fatal flaw. On the web, the shape of HTTP requests differs depending on the type of resource that is requested. For example, the [`Accept` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) affects the MIME type of the response, and the [`Sec-Fetch-Dest` metadata header](https://web.dev/articles/fetch-metadata) affects whether the web server accepts or rejects the request. Because an import assertion could not affect _how_ to load a module, it was not able to change the shape of the HTTP request. This meant that it could not correctly work with the security model of the web.
```
nicolo-ribaudo, line 23:
```suggestion
However, assert-only semantics had a fatal flaw. On the web, the shape of HTTP requests differs depending on the type of resource that is requested. For example, the [`Accept` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) affects the MIME type of the response, and the [`Sec-Fetch-Dest` metadata header](https://web.dev/articles/fetch-metadata) affects whether the web server accepts or rejects the request. Because an import assertion could not affect _how_ to load a module, it was not able to change the shape of the HTTP request. The type of the resource that is being requested also affects which [Content Security Policies](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) are used: import assertions could not correctly work with the security model of the web.
```

The security aspect is not directly related to which HTTP headers are used — those are more for content negotiation.

src/features/import-attributes.md
```diff
@@ -0,0 +1,63 @@
...
+
+The `assert` keyword is deprecated as of V8 v12.3 and is planned to be removed by v12.6. Please use `with` instead of `assert`! Use of the `assert` clause will print a warning to the console urging use of `with` instead.
+
+## Import attribute support
+
+<feature-support chrome="123 https://chromestatus.com/feature/5205869105250304"
+                 firefox="no"
+                 safari="17.2 https://developer.apple.com/documentation/safari-release-notes/safari-17_2-release-notes"
+                 nodejs="no"
+                 babel="yes https://github.com/babel/babel/pull/15536"></feature-support>
```
nicolo-ribaudo, line 63:
```suggestion
                 babel="yes https://babeljs.io/blog/2023/05/26/7.22.0#import-attributes-15536-15620"></feature-support>
```

The release post is more user-friendly than the PR, since it explains how to actually use the new syntax :)

src/features/import-attributes.md
syg: I'm going to leave this as is. That level of understanding of "loaded" is not what's intended here, and I think using "executed" in the context of a JSON module, which developers would think of as data instead of code, would be more confusing.

PR #751, opened 2024-03-28
Actualize `core-js` data for `Iterator` helpers

zloirock:
Similarly to #567 and #602

@rmahdav please, next time check it too.

Also, nbsp between tag attributes (O_o) are replaced with usual spaces.

### DIFF
--- a/src/features/iterator-helpers.md
+++ b/src/features/iterator-helpers.md
@@ -186,7 +186,7 @@ Iterator helpers are shipped in V8 v12.2.
 ## Import attribute support
 
 <feature-support chrome="122 https://chromestatus.com/feature/5102502917177344"
-                 firefox="no https://bugzilla.mozilla.org/show_bug.cgi?id=1568906"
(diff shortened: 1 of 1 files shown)


rmahdav:
Hi Denis. Thanks for your PR. I checked the babel repository twice (before and after publishing the blog post). I did not find any shipped code related to iterator helpers. That is why I marked shipping in babel as `no` in the post. Based on my understanding, I had to check the babel itself for that part of the post.

zloirock:
@rmahdav for built-ins like `Iterator` helpers, Babel injects `core-js` polyfills - most likely you will not find anything about built-ins in the Babel repo, so for this case is better to check `core-js` docs. Take a look, for example, at [`preset-env`](https://babeljs.io/docs/babel-preset-env) and [`runtime`](https://babeljs.io/docs/babel-plugin-transform-runtime) docs.

rmahdav (approved):
Okay. That makes sense.

PR #755, opened 2024-04-02
Merge legacy.js and main.mjs into main.js

camillobruni:
There is no advantage of using modules over classic scripts if there are no imports.
We currently have the legacy script as a nomodule fallback, we can merge them easily into a single main.js script and simplify the page a bit.


o- (approved):

tomayac:
@camillobruni I didn't look closely, but it seems like this potentially broke dark mode. Could you have a look, please?

camillobruni:
Ups right, I accidentally removed the dark-mode import... planned to replace this with some simpler logic.
Will fix this tomorow.

tomayac:
Filed https://github.com/v8/v8.dev/issues/758 as a reminder. Thanks for working on this!

PR #760, opened 2024-04-08
Disable service workers

camillobruni:
While an offline site is nice to have, it adds more problems than it solves for this kind of page.
In the effort to reduce maintenance burden I'm removing the service workers.
This first PR manually unregisters the service workers and later in a couple of months we can remove the sw.js file.


package.json
```diff
@@ -8,7 +8,7 @@
     "js": "shx mkdir -p dist/_js && terser src/_js/main.js -cm -o dist/_js/main.js",
     "json": "json-minify src/.webmanifest > dist/.webmanifest",
     "xml": "minify-xml --in-place dist/sitemap.xml && minify-xml --in-place dist/blog.atom && minify-xml --in-place dist/features.atom",
-    "sw": "workbox injectManifest workbox-config.js && terser dist/sw.js -cm -o dist/sw.js",
```
ajklein, line 11: Can we also drop the dependency on workbox and remove workbox-config.js?

ajklein (commented):
I think dropping SW support for v8.dev is a great idea! Is there anything blocking this?

o- (approved):
thanks

PR #763, opened 2024-05-03
Update merge handling

lutzvahl:


src/docs/merge-patch.md
```diff
@@ -22,42 +22,42 @@ More information can be found on the [relevant Chromium page](https://chromium.g
...
-tools/release/mergeinfo.py $COMMIT_HASH
-```
 
-If it tells you `Is on Canary: No Canary coverage` you should not merge yet because the fix was not yet deployed on a Canary build. A good rule of the thumb is to wait at least 3 days after the fix has landed until the merge is conducted.
+On top the **Releases** section should show a Canary.
 
 ## How to create the merge CL
 
-### Option 1: Using [gerrit](https://chromium-review.googlesource.com/)
+### Option 1: Using [gerrit](https://chromium-review.googlesource.com/) - Recommended
```
LeszekSwirski, line 51: I actually scripted this process as https://source.chromium.org/chromium/chromium/src/+/main:v8/tools/release/merge_to_branch_gerrit.py, maybe we want to recommend that instead?

lutzvahl: SG, does it make sense to replace Option 2 below with the new script?

LeszekSwirski: sg

lutzvahl (commented):
adjusted to use the new script

LeszekSwirski (approved):

PR #764, opened 2024-05-12
Update stack-trace-api.md

php4fan:
Added that setting `Error.stackTraceLimit = undefined` also disables stack traces like setting it to zero does.

I suspect that is true for any non-number value (and also negative numbers, but that is implicit in the statement that any finite integer defines the maximum), but I can only say for sure it's the case for `undefined`. 

Note that at some point in the past setting it to `undefined` used to have the same effect that `Infinity` has now. I'm not sure when that changed, it may have been many years ago, but at some point it was true, at least in Chrome.

### DIFF
--- a/src/docs/stack-trace-api.md
+++ b/src/docs/stack-trace-api.md
@@ -28,7 +28,7 @@ The stack trace is collected when the error is created and is the same regardles
 Error.stackTraceLimit
 ```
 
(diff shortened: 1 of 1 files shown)


camillobruni:
Could you sign the CLA to land your PR?  https://cla.developers.google.com/. Thanks!

php4fan:
> Could you sign the CLA to land your PR?

I already did

PR #806, opened 2025-05-06
Add a blog post about explicit resource management proposal

rmahdav:


rmahdav:
@syg  please take a look and let me know if the example makes more sense now.

src/features/explicit-resource-management.md
```diff
@@ -0,0 +1,219 @@
...
+    let value;
+  
+    while (!done) {
+        ({ done, value } = await reader.read());
+        if (value) {
+            loaded += value.byteLength;
+            if (contentLength) {
+            const progress = (loaded / total) * 100;
+            console.log(`Downloaded ${loaded} of ${total} bytes (${progress.toFixed(2)}%)`);
+            updateProgressBar(progress); //some UI update
```
syg, line 46: The if body isn't indented.

src/features/explicit-resource-management.md
```diff
@@ -0,0 +1,219 @@
...
+            }
+        }
+    }
+   
+    reader.releaseLock(); // Explicitly release the lock in a finally block
```
syg, line 51: Comment isn't right, since there is no finally block.

src/features/explicit-resource-management.md
```diff
@@ -0,0 +1,219 @@
...
+    
+    ... // Do some other things
+
+    // Now, another part of your application tries to get a reader for the same stream, and save it to a file. 
+    // If the lock has not been released, it will throw an error.  
+    
+    const reader2 = response.body.getReader();
```
syg, line 58: Does this actually work? Since the first loop above already read all of the body, does this reset the stream to the beginning or something?

src/features/explicit-resource-management.md
```diff
@@ -0,0 +1,219 @@
...
+
+```javascript
+let stack = new DisposableStack();
+stack.use(firstDisposableResource);
+stack.dispose();
+```
+
+## Availability
+
+Explicit Resource Management is shipped in V8 v13.4.
```
syg, line 211: I think a strict reading of this sentence is false -- we shipped it in Chrome, but didn't flip the flag in V8 until recently. I think we should list the two versions separately.

src/features/explicit-resource-management.md
```diff
@@ -0,0 +1,219 @@
...
+    ... // The rest of the code
+  }
+  
+  // Example usage:
+  downloadFileWithUsing('https://example.com/largefile.dat');
+```
+
+## `DisposableStack` and `AsyncDisposableStack`
+
+To further facilitate managing multiple disposable resources, the proposal introduces `DisposableStack` and `AsyncDisposableStack`. These stack-based structures allow developers to group and dispose of multiple resources in a coordinated manner. Resources are added to the stack, and when the stack is disposed, either synchronously or asynchronously, the resources are disposed of in the reverse order they were added, ensuring that any dependencies between them are handled correctly. This simplifies the cleanup process when dealing with complex scenarios involving multiple related resources. Both structures provide methods like `use()`, `adopt()`, and `defer()` to add resources or disposal actions, and a `dispose()` or `asyncDispose()` method to trigger the cleanup. They offer a robust way to manage the disposal of multiple resources within a defined scope.
```
syg, line 159: I think it should also be mentioned that these two kinds of objects have `[Symbol.dispose]` and `[Symbol.asyncDispose]`, respectively, so you can use them with `using`.

src/features/explicit-resource-management.md
```diff
@@ -0,0 +1,219 @@
...
+
+    // Do some other things... 
+    
+    { // Introduce a new scope with a block
+        // Get a new reader for the same stream
+        using readerResource = {
+        reader2: response.body.getReader(),
+        [Symbol.dispose]() {
+            this.reader2.releaseLock();
+        },
```
syg, line 130: Indent.

syg (commented):
I'm not sure if the main example actually works. Can you get a second reader after the first reader is done? What actually happens if the API allows this, like does it seek to the beginning of the already downloaded file, does it re-download?

src/features/explicit-resource-management.md
```diff
@@ -0,0 +1,219 @@
...
+
+## `using` and `await using` declarations
+
+The core of the Explicit Resource Management proposal lies in the `using` and `await using` declarations. The `using` declaration is designed for synchronous resources, ensuring that the `[Symbol.dispose]()` method of a disposable resource is called when the scope in which it's declared is exited. For asynchronous resources, the `await using` declaration works similarly, but ensures that the `[Symbol.asyncDispose]()` method is called, allowing for asynchronous cleanup operations. This distinction enables developers to reliably manage both synchronous and asynchronous resources, preventing leaks and improving overall code quality. The `using` and `await using` keywords can be used within a Block, ForStatement, ForInOfStatement, FunctionBody, GeneratorBody, AsyncGeneratorBody, AsyncFunctionBody, or ClassStaticBlockBody.
```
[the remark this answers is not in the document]
rmahdav: Let me know if the change to this part is okay.

src/features/explicit-resource-management.md
rmahdav: No actually. It gets the second reader but the reader does nothing. After our brainstorm, I changed the example to show error handling but it is still not good, IMO.

src/features/explicit-resource-management.md
rmahdav: Good point, fixed now.

syg (commented):
Please also generate a local preview and see if any of the long lines in the examples need to be broken up.

src/features/explicit-resource-management.md
```diff
@@ -0,0 +1,219 @@
...
+avatars:
+  - 'rezvan-mahdavi-hezaveh'
+date: 2025-05-09
+tags:
+  - ECMAScript
+description: 'The Explicit Resource Management proposal empowers developers to explicitly manage the lifecycle of resources.'
+tweet: ''
+---
+
+The *Explicit Resource Management* proposal introduces a more deterministic approach, empowering developers to explicitly manage the lifecycle of resources like file handles, network connections, and more. This proposal brings the following additions to the language: the `using` and `await using` declarations, which automatically calls dispose method when a resource goes out of scope; `[Symbol.dispose]()` and `[Symbol.asyncDispose]()` symbols for cleanup operations; two new global objects `DisposableStack` and `AsyncDisposableStack` as containers to aggregate disposable resources; and `SuppressedError` as a new type of error which would contain both the error that was most recently thrown, as well as the error that was suppressed. These additions enable developers to write more robust, performant, and maintainable code by providing fine-grained control over resource disposal.
```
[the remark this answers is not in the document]
rmahdav: Is it better now?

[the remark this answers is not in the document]
syg: New sentence reads good to me, thanks.

src/features/explicit-resource-management.md
```diff
@@ -0,0 +1,245 @@
...
+    }
+    const response = await responsePromise;
+    if (!response.ok) {
+      throw new Error(`HTTP error! status: ${response.status}`);
+    }
+    const processedData = await processData(response);
+
+    // Do something with processedData
+    ...
+ }
```
syg, line 80: IMO you don't need to keep repeating the `readFile` function in subsequent examples if nothing in it changes.

src/features/explicit-resource-management.md
```diff
@@ -0,0 +1,245 @@
...
+
+ async function processData(response) {
+    const reader = response.body.getReader();
+    let done = false;
+    let value;
+    
+    { // Introduce a new scope with a block
```
syg, line 132: You don't need an extra block scope here, right?

src/features/explicit-resource-management.md
```diff
@@ -0,0 +1,245 @@
...
+}
+// Reader lock released.
+```
+
+`adopt(value, onDispose)` adds a non-disposable resource and a disposal callback to the top of the stack.
+
+```javascript
+{
+    using stack = new DisposableStack();
+    stack.adopt(response.body.getReader(), reader => {reader.releaseLock(); console.log('Reader lock released.');});
```
syg, line 190: This line looks too line. Please check how it looks in a local build.

syg (approved):
lgtm % comments and fixing the build

src/features/explicit-resource-management.md
rmahdav: Good point, that's right.

src/features/explicit-resource-management.md
rmahdav: Done

PR #817, opened 2025-08-21
Update elements kind link in fast-properties.md

SumeetP96:
The updated link will redirect to the current location of the `elements-kind.h` file.

Old location: `/src/elements-kind.h`
New location: `/src/objects/elements-kind.h`


LeszekSwirski:
Thanks!

PR #841, opened 2026-04-15
Add sparkpug and turboshaft logo

camillobruni:
Drive-by-fix:
- add npm run dev
- use http-server instead of supertstatic (which seems to fail on node 24)


victorgomes:
LGTM!

victorgomes (approved):

PR #843, opened 2026-05-04
Publish V8's revised shipping guidelines

ecmziegler:
We harmonized JS & Wasm shipping guidelines and formalized already existing practices like V8 launch reviews and extended them further.


ecmziegler:
I noticed some minor font and presentation issues which I will address.

